PR #16, opened 2020-09-14
HCAP-6 xml parser script

mauriciofreshw:


ashtonmeuser (dismissed):

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,82 @@
...
+
+// run: node parse-xml ./folder-containing-xml-files
+/* eslint-disable no-console */
+(async () => {
+  try {
+    const xmlStrings = getDirContent(process.argv[2]);
+    const parsedJsonObjs = [];
+    xmlStrings.forEach((xml) => {
+      const jsonObj = fromXmlStrings(xml);
+      const parsedJsonObj = {
```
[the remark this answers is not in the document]
mauriciofreshw: I added validation before the submission. In case an XML is invalid the script will stop and print which file is, all XML files must be valid.

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    return filesString;
+  } catch (e) {
+    if (typeof pathString === 'undefined') throw Error('Provide a path to a folder containing XMLs as the first command line argument');
+    throw Error(`Could not read contents of file ${pathString}`);
+  }
+};
+
+const fromXmlStrings = (xmlStrings) => parser.parse(xmlStrings);
```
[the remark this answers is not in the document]
mauriciofreshw: Yep, in the main try/catch

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,82 @@
...
+      const parsedJsonObj = {
+        // number: _.get(jsonObj, 'form.formControl.grid-4.number'),
+        eligibility: _.get(jsonObj, 'form.section-3.grid-7.legallyEligible') === 'Yes',
+        firstName: _.get(jsonObj, 'form.contactInformation.grid-1.firstName'),
+        lastName: _.get(jsonObj, 'form.contactInformation.grid-1.lastName'),
+        phoneNumber: _.get(jsonObj, 'form.contactInformation.grid-1.primaryPhone').toString(),
+        emailAddress: _.get(jsonObj, 'form.formControl.grid-4.recipient'),
+        postalCode: _.get(jsonObj, 'form.contactInformation.grid-1.postalCode'),
+        preferredLocation: [_.get(jsonObj, 'form.locationPreference.grid-2.healthRegion')],
+        consent: _.get(jsonObj, 'form.consent.grid-3.confirmed') === 'Yes',
```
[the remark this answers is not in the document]
mauriciofreshw: I added **orbeonId** as a nullable unique Id, so this can work as a source identifier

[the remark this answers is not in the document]
fwkendall: Sounds good!

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,101 @@
...
+
+const makeTransactionIterator = (endpoint) => (d) => postHcapSubmission(endpoint, d);
+
+// run: node parse-xml ./folder-containing-xml-files
+// run mocks: node parse-xml ../../tests/mock/xml
+/* eslint-disable no-console */
+(async () => {
+  let currentFile;
+  try {
+    const xmlStrings = getDirContent(process.argv[2]);
```
fwkendall, line 69: This line needs to be nullsafe, right now we fail out with a TypeError undefined instead of hitting the error message on line 33 :)

fwkendall:
High level thoughts now that I've run this thing locally:

- I feel like since it hits the API and not the DB directly, it makes more sense to move this from server/db/scripts/parse-xml.js to server/scripts/parse-xml.js (one level up from the /db folder, since it's a server script)
- It would be nice to add a command for this to package.json and a note about it to the README
- The script should exit right away if it finds 0 files to parse, instead of continuing on to prompt for which endpoint you want to use

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,101 @@
...
+    /* eslint-disable no-restricted-syntax */
+    for (const xml of xmlStrings) {
+      currentFile = xml;
+      const jsonObj = fromXmlStrings(xml.content);
+      const parsedJsonObj = {
+        orbeonId: xml.name.match(/^Health Career Access Program - Expression of Interest - ([\w-]{16}).xml$/)[1],
+        eligibility: _.get(jsonObj, 'form.section-3.grid-7.legallyEligible') === 'Yes',
+        firstName: _.get(jsonObj, 'form.contactInformation.grid-1.firstName'),
+        lastName: _.get(jsonObj, 'form.contactInformation.grid-1.lastName'),
+        phoneNumber: _.get(jsonObj, 'form.contactInformation.grid-1.primaryPhone').toString(),
```
fwkendall, line 81: Need to add optional chaining here for calling toString()

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,101 @@
...
+const fromXmlStrings = (xmlStrings) => parser.parse(xmlStrings);
+
+const getUserInput = async () => { // Prompt user for query, output format
+  const { 'Select Endpoint': endpoint } = await inquirer.prompt([{ name: 'Select Endpoint', type: 'list', choices: endpoints }]);
+  const { Proceed: proceed } = await inquirer.prompt([{ name: 'Proceed', type: 'confirm' }]);
+  if (!proceed) throw Error('User cancelled operation');
+  return { endpoint };
+};
+
+const extracPreferredLocations = (preferredLocationString) => {
```
fwkendall, line 47: Typo - extract* :)

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,101 @@
...
+        eligibility: _.get(jsonObj, 'form.section-3.grid-7.legallyEligible') === 'Yes',
+        firstName: _.get(jsonObj, 'form.contactInformation.grid-1.firstName'),
+        lastName: _.get(jsonObj, 'form.contactInformation.grid-1.lastName'),
+        phoneNumber: _.get(jsonObj, 'form.contactInformation.grid-1.primaryPhone').toString(),
+        emailAddress: _.get(jsonObj, 'form.formControl.grid-4.recipient'),
+        postalCode: _.get(jsonObj, 'form.contactInformation.grid-1.postalCode'),
+        preferredLocation: extracPreferredLocations(_.get(jsonObj, 'form.locationPreference.grid-2.healthRegion')),
+        consent: _.get(jsonObj, 'form.consent.grid-3.confirmed') === 'Yes',
+      };
+      await validate(FormSchema, parsedJsonObj);
```
fwkendall, line 87: I'm wondering if we want a try/catch here as well since this throws an exception back when something fails validation - we probably want to find out how many forms are going to fail instead of just stopping execution on the first one (so we know if we have one form with an issue, or if we can't parse *any* form)

fwkendall: The catch would be setting a flag or building a "failed" list of forms or something I guess, then we would handle that once we've finished iterating through all the forms

server/db/scripts/parse-xml.js
```diff
@@ -0,0 +1,101 @@
...
+  try {
+    const xmlStrings = getDirContent(process.argv[2]);
+    const parsedJsonObjs = [];
+    /* eslint-disable no-await-in-loop */
+    /* eslint-disable no-restricted-syntax */
+    for (const xml of xmlStrings) {
+      currentFile = xml;
+      const jsonObj = fromXmlStrings(xml.content);
+      const parsedJsonObj = {
+        orbeonId: xml.name.match(/^Health Career Access Program - Expression of Interest - ([\w-]{16}).xml$/)[1],
```
fwkendall, line 77: Confirmed this fails on files with (1) in the name currently

` File error: Health Career Access Program - Expression of Interest - a3719059eabb811a (1).xml

TypeError: Cannot read property '1' of null
    at /Users/kendallolsen-maier/workspace/coco/hcap-health-career-access-program/hcap/server/db/scripts/parse-xml.js:77:110
    at process.runNextTicks [as _tickCallback] (internal/process/next_tick.js:47:5)
    at Function.Module.runMain (internal/modules/cjs/loader.js:800:11)
    at executeUserCode (internal/bootstrap/node.js:526:15)
    at startMainThreadExecution (internal/bootstrap/node.js:439:3)`

server/db/scripts/parse-xml.js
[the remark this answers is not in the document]
fwkendall: Confirmed this bails out - it looks kinda ugly though...

``` File error: sitemap.xml

TypeError: Cannot read property '1' of null
    at /Users/kendallolsen-maier/workspace/coco/hcap-health-career-access-program/hcap/server/db/scripts/parse-xml.js:77:110
    at process.runNextTicks [as _tickCallback] (internal/process/next_tick.js:47:5)
    at Function.Module.runMain (internal/modules/cjs/loader.js:800:11)
    at executeUserCode (internal/bootstrap/node.js:526:15)
    at startMainThreadExecution (internal/bootstrap/node.js:439:3)```

File: server/db/scripts/parse-xml.js [hunk 9]
```diff
@@ -0,0 +1,101 @@
...
+        eligibility: _.get(jsonObj, 'form.section-3.grid-7.legallyEligible') === 'Yes',
+        firstName: _.get(jsonObj, 'form.contactInformation.grid-1.firstName'),
+        lastName: _.get(jsonObj, 'form.contactInformation.grid-1.lastName'),
+        phoneNumber: _.get(jsonObj, 'form.contactInformation.grid-1.primaryPhone').toString(),
+        emailAddress: _.get(jsonObj, 'form.formControl.grid-4.recipient'),
```
fwkendall, line 82: After a bunch of messing around I finally figured out that this location needs to change - this field in the XML refers to the address the form is actually being submitted to. The correct path is now:

```
form.contactInformation.grid-1.emailAddress

This is the main difference I see in the latest XML copy vs. the earlier test files.

File: server/db/scripts/parse-xml.js (hunk 5 above)
Comment (fwkendall) [reply server/db/scripts/parse-xml.js:81]: In addition to the above issue, I confirmed this field name actually changed from the earlier XML as well. The new field name is "phoneNumber" instead of "primaryPhone"

File: server/tests/mock/xml/Health Career Access Program - Expression of Interest - 9353ec90e6ea8727 (1).xml [hunk 10]
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?><form>
```
fwkendall, line 1: I'm assuming that we were provided a test file with no valid data, but I would still feel more comfortable if we swap out all of the values in the file for ones we know *for sure* are test data e.g. A1A 1A1 for the postal code, obviously non-existent emails and names, etc.

number/owner look fine but we should make sure to swap out the approver + recipient values as well

mauriciofreshw: Totally! Nice catch.

server/scripts/parse-xml.js
```diff
@@ -0,0 +1,125 @@
...
+        return 'Vancouver Coastal';
+      case 'Island':
+        return 'Vancouver Island';
+      default:
+        return item;
+    }
+  });
+};
+
+const extractOrbeonId = (name) => name.match(/^Health Career Access Program - Expression of Interest - ([\w-]{16})( \([0-9]+\))?.xml$/)[1];
```
fwkendall, line 61: Discovered this falls over pretty hard if you have issues in your XML file names, e.g. if you're missing the space between the - and the ID.

```
 File error: Health Career Access Program - Expression of Interest -hcc9tlb0srg8bned.xml
TypeError: Cannot read property '1' of null
```

I'm wondering if it makes the most sense just to include the expected file name pattern in getDirContent? Maybe it would make sense for the script to exit there if it finds any files that are not XML, or any files that don't match the expected XML filename format?

mauriciofreshw: Nice catch. I added the file name validation to the error array, this way it will be part of all validations.

fwkendall (dismissed):

ashtonmeuser (approved):

fwkendall (approved):

PR #20, opened 2020-09-15
Added test pipeline

mauriciofreshw:


.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+          mongodb-version: ${{ matrix.mongodb-version }}
+
+      - name: NPM Test
+        working-directory: ./server
+        env:
+          DB_SERVER: localhost
+          DB_USER: development
+          DB_PASSWORD: development
+          DB_NAME: development
+          DB_PORT: 27017
```
fwkendall, line 38: Part of the current issue might be related to these... This is a DB instance that only lives for the duration of the tests run, right? The tests are expecting that hardcoded test DB name I was asking about earlier 😂 

Best path forward is probably to parameterize the test DB info and get it into secrets?

mauriciofreshw: No need anymore. As you said these parameters don’t make sense since is a short-lived DB (neither the plugin has support for this), I just change for a non-auth and now the tests are working properly :)

fwkendall: Awesome! That's way better 🙌

fwkendall (approved):
Looks good! :)

fwkendall (approved):

ashtonmeuser (approved):

ashtonmeuser:
Just so we're all clear, this is not running tests in the same way we do locally. May cause problems at some point.

fwkendall:
> Just so we're all clear, this is not running tests in the same way we do locally. May cause problems at some point.

Discussed, we are OK deferring this risk to later and working through any issues that might come up.

PR #37, opened 2020-10-08
Hcap 66 update api to postgres

mauriciofreshw:


.config/.env.example
```diff
@@ -1,5 +1,6 @@
 # Database Config
-MONGODB_USER=development
-MONGODB_PASSWORD=
-MONGODB_DATABASE=development
-MONGODB_URI=mongodb://${MONGODB_USER}:${MONGODB_PASSWORD}@mongodb:27017/${MONGODB_DATABASE}
+POSTGRES_HOST=
+POSTGRES_PORT=
+POSTGRES_DB=db_dev
+POSTGRES_USER=admin
+POSTGRES_PASSWORD=development
```
fwkendall, line 6: Obviously this is the example and not used anywhere, but I still prefer to leave password fields blank :) I usually provide an example dummy value for everything but passwords

.github/workflows/test.yml
```diff
@@ -9,30 +9,31 @@ jobs:
...
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Use Node.js ${{ matrix.node-version }}
```
fwkendall, line 21: curious why we don't need this now?

server/db/scripts/seed.js
```diff
@@ -2,50 +2,32 @@
 const { dbClient } = require('../db.js');
 const { schema } = require('../schema.js');
 
+//Postgres duplicated error code
```
fwkendall, line 5: Couple of tiny thoughts for this file, I'm cool if we don't want to address these:
- there's a missing space after the // on this line
- since we've had some on-and-off issues with local dev, I think it might be nice to put some logging after each successful operation in this script (e.g. table created, index created)

.github/workflows/test.yml
```diff
@@ -9,30 +9,31 @@ jobs:
...
-        with:
-          mongodb-version: ${{ matrix.mongodb-version }}
-
       - name: NPM Test
         working-directory: ./server
         env:
-          MONGODB_URI: mongodb://localhost:27017/test
+          POSTGRES_DB: db_test
+          POSTGRES_USER: admin
+          POSTGRES_PASSWORD: admin
```
fwkendall, line 36: I can't shake the dirty feeling of having this in source, even if it's intentionally a dummy value 😓 

I can't think of a solution more elegant than adding the test user/password to our .env though. Thoughts?

.docker/postgres/sql-init/setup.sql
```diff
@@ -0,0 +1,3 @@
+CREATE DATABASE db_test;
+
+GRANT ALL PRIVILEGES ON DATABASE db_test TO admin;
```
fwkendall, line 3: my comment about whether we want to make the test user an env var would affect this file too 😄

fwkendall:
So much clean up in this PR! The pg lib being used looks really slick.

docker-compose.dev.yml
```diff
@@ -8,14 +8,14 @@ services:
       context: ./client
       dockerfile: Dockerfile.dev
     ports:
-    - 4000:4000
+      - 4000:4000
```
ashtonmeuser, line 11: Nitpick: We should decide on a format for YAML and stick with it. If we like our arrays indented, we may want to align the OpenShift templates with this format.

.config/.env.example
mauriciofreshw: Good point, changing it.

.github/workflows/test.yml
mauriciofreshw: Turns out that there’s a better way to do that without the need for an Action

server/package.json
```diff
@@ -12,20 +12,20 @@
     "db:migrate-create": "cd db && migrate create",
     "watch": "NODE_ENV=development nodemon main.js",
     "start": "NODE_ENV=production node main.js",
-    "test": "NODE_ENV=test jest --runInBand"
+    "test": "NODE_ENV=test jest --runInBand --detectOpenHandles"
```
ashtonmeuser, line 15: Not sure we should be using `--detectOpenHandles` in all of our tests. Looks like it is primarily used for debugging (see [docs](https://jestjs.io/docs/en/cli)). Also, this arg implies `--runInBand`, so if we're confident that we need `--detectOpenHandles`, we should remove `--runInBand`.

server/server.js
```diff
@@ -60,33 +60,29 @@ app.post(`${apiBaseUrl}/employer-form`,
...
-      if (await formsCollection.countDocuments({ id: randomHexId }, { limit: 1 }) > 0) {
+      if (await dbClient.db[collections.FORMS].countDoc({ body: { formId: randomHexId } }) > 0) {
         return generateRandomHexId();
       }
       return randomHexId;
     }
 
     // Form ID
-    const id = await generateRandomHexId();
+    const formId = await generateRandomHexId();
```
ashtonmeuser, line 76: As we discussed, we may not need this. Removing this logic would also avoid the inherent race condition of a colliding ID being created between querying for the generated ID and writing the item to the DB. We could use something like `id UUID PRIMARY KEY DEFAULT gen_random_uuid()` or even use the default ID generated by Massive instead.

server/tests/util/db.js
```diff
@@ -1,46 +1,34 @@
...
-  schema.forEach((schemaItem) => {
-    const schemaItemCollection = dbClient.db.collection(schemaItem.collection);
-    schemaItem.indexes.forEach((index) => {
-      results.push(schemaItemCollection.createIndex({ [index.key]: 1 }, index.options));
-    });
-  });
-
-  await Promise.all(results); // Wait for all synchronous operations to pass/fail
+  /* eslint-disable no-await-in-loop */
+  /* eslint-disable no-restricted-syntax */
```
ashtonmeuser, line 5: We should consider avoiding these warnings. At the very least, we should re-enable the linting rules after this little loop.

server/tests/util/db.js
```diff
@@ -1,46 +1,34 @@
...
-  const collectionsToDelete = await dbClient.db.collections();
-
-  const results = [];
-  collectionsToDelete.forEach((collection) => {
-    results.push(collection.deleteOne());
-  });
-
-  return Promise.all(results); // Wait for all synchronous operations to pass/fail
+  /* eslint-disable no-await-in-loop */
+  /* eslint-disable no-restricted-syntax */
```
ashtonmeuser, line 16: Same as above. We can structure this differently to avoid the warning entirely.

server/db/scripts/seed.js
mauriciofreshw: No problem. Just did :)

fwkendall: Looks great! :)

docker-compose.dev.yml
fwkendall: I personally like indentation, just feels easier to read to me. But I'm good to go with whatever format, consistency is more important to me than having it one way or the other

.github/workflows/test.yml
mauriciofreshw: But for tests on GH actions the dummy password will be exposed anyway, even using in an .env file. Should we use it as a secret?

.docker/postgres/sql-init/setup.sql
mauriciofreshw: For this one is a little bit complicated, we would need to think in a way to inject those variables in an .sql file.

docker-compose.dev.yml
mauriciofreshw: I just realized this now, this is what my VSCode do when I format yaml files.

server/package.json
mauriciofreshw: I was debugging this and I forgot to remove it because with this command the tests were able to work without warning about open async operations (I guess it’s because massive.js has no close() connection function, and jest takes a while with this command, giving time to massive.js detect an idle connection and close it). So the only way I found to solve this problem is using --forceExit

server/server.js
mauriciofreshw: Yes, I also think this is not necessary since the id generated by massive.js is also simple to read. (a simple increment integer). Are we fine removing this feature in this branch?

server/tests/util/db.js
mauriciofreshw: True. There are some implications using **async** in a **forEach** instead of a **for** in this code, but since it's just a seed script I think we are fine just enabling the lint rules for now.

server/server.js
fwkendall: Totally good with removing this! Let's streamline

.github/workflows/test.yml
fwkendall: Discussed offline - since this is ephemeral and can be seen via the GH Actions, we aren't worried about exposing the default credentials here since they're a known entity for a PostgreSQL image anyway.

docker-compose.dev.yml
fwkendall: Decided as a team we'll start getting these indented going forward. Requires no updates to this PR.

fwkendall (approved):

ashtonmeuser (approved):

PR #41, opened 2020-10-09
Added operator info

mauriciofreshw:


fwkendall (dismissed):

fwkendall:
I think we're actually missing a field in this section... it's hard to tell from the spec, but it looks like there are four fields?

- Operator name
- Operator contact for EOI process
- Operator contact email
- Operator contact phone number

I guess operator name is the person who runs/manages the location, but there might be a separate person they should actually contact for followup?

fwkendall:
I'm also wondering if we should just go ahead and follow our usual first name vs. last name field split, rather than having a single field for the name?

client/src/components/employer-form/OperatorInfo.js
```diff
@@ -0,0 +1,54 @@
...
+            Operator Information
+          </Typography>
+          <Divider />
+        </Grid>
+
+        <Grid item xs={12}>
+          <FastField
+            name="operatorName"
+            component={RenderTextField}
+            label="* Operator Name:"
```
fwkendall, line 26: This is a style difference I think we need to make a decision about as well - all of our forms use sentence case for labels, not title case

mauriciofreshw:
> I'm also wondering if we should just go ahead and follow our usual first name vs. last name field split, rather than having a single field for the name?

I think it was mistyped, the 2nd field is actually the title of the following two fields. At least was that what I thought.

fwkendall (approved):

fwkendall (approved):

ashtonmeuser (approved):

ashtonmeuser (approved):

fwkendall (approved):

PR #45, opened 2020-10-14
Hcap 81 openshift patroni

ashtonmeuser:
Patroni templates
Build in `-tools` namespace

### DIFF
--- a/openshift/patroni.dc.yml
+++ b/openshift/patroni.dc.yml
@@ -30,13 +30,13 @@ parameters:
     description: Registry from which the Patroni and Postgres images will be pulled
     value: image-registry.openshift-image-registry.svc:5000
     required: true
-  - name: IMAGE_STREAM_NAMESPACE
+  - name: IMAGE_NAMESPACE
     description: The OpenShift Namespace where the Patroni and Postgres image streams reside
     value: bcgov
     required: true
-  - name: IMAGE_STREAM_TAG
-    description: Patroni image tag
-    value: patroni:v11-stable
+  - name: IMAGE_TAG
+    description: Patroni image tag to pull from image stream
+    value: v11-stable
     required: true
   - name: PVC_SIZE
     description: The size of the persistent volume to create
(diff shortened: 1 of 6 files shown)


openshift/patroni.dc.yml
```diff
@@ -0,0 +1,201 @@
...
+                - containerPort: 5432
+                  protocol: TCP
+              resources:
+                requests:
+                  cpu: ${CPU_REQUEST}
+                  memory: ${MEMORY_REQUEST}
+                limits:
+                  cpu: ${CPU_LIMIT}
+                  memory: ${MEMORY_LIMIT}
+              terminationMessagePath: /dev/termination-log
```
fwkendall, line 170: so we do already have some logging going to file :) should we follow this path paradigm for other logging for the time being?

openshift/patroni.dc.yml
```diff
@@ -0,0 +1,201 @@
...
+                valueFrom:
+                  secretKeyRef:
+                    key: replication-username
+                    name: ${APP_NAME}-patroni
+              - name: PATRONI_REPLICATION_PASSWORD
+                valueFrom:
+                  secretKeyRef:
+                    key: replication-password
+                    name: ${APP_NAME}-patroni
+              - name: APP_USER
```
fwkendall, line 123: it seems a little weird that these are called APP_USER and APP_PASSWORD here, but seem to be the same things defined in patroni.prep where they're referred to as DB_USERNAME and DB_PASSWORD?

mauriciofreshw (dismissed):

fwkendall (dismissed):
Looks good, just a couple of questions from me :)

mauriciofreshw (dismissed):

fwkendall (approved):

mauriciofreshw (approved):

openshift/patroni.dc.yml
ashtonmeuser: Have not even looked into this, but it appears as if it's got some basic logging baked in. As it stands, `/dev/termination_log` is empty.

openshift/patroni.dc.yml
ashtonmeuser: That does seem quite weird. However, as far as I can tell, this is to appease Patroni. So we can call it whatever we like everywhere _except_ the names of these env vars used by Patroni. I called it `DB_NAME`, `DB_USERNAME`, etc. elsewhere to align with how we've done things elsewhere.

PR #49, opened 2020-10-15
HCAP-84 & HCAP-85: Added phone contact info to top of form and consent checkbox at bottom

fwkendall:


mauriciofreshw (approved):

mauriciofreshw (approved):

ashtonmeuser (approved):

ashtonmeuser (approved):

server/validation.js
```diff
@@ -126,6 +126,9 @@ const EmployerFormSchema = yup.object().noUnknown('Unknown field for form').shap
     vacancyPartTime: yup.number().integer('Number must be an integer').moreThan(-1, 'Number must be positive'),
     vacancyCasual: yup.number().integer('Number must be an integer').moreThan(-1, 'Number must be positive'),
   })),
+
+  // Consent
+  consent: yup.boolean().typeError(errorMessage).required(errorMessage).test('is-true', errorMessage, (v) => v === true),
```
mauriciofreshw, line 131: Does it make sense to have on the server-side? The user can only hit the endpoint if the consent is true, what could we benefit in having a request with a consent = false being handled in the server?

fwkendall: I generally think that server-side validation should be even more aggressive than client-side. Client side is preventing user error, but server side is preventing bad data from actually getting to the DB. It's super unlikely anything hits our endpoint that passes all validation except this field, but we really don't want any submissions to come into the DB without the consent being true so I think it makes sense to apply validation.

We actually did get attempted empty submissions on our other forms when they went live, for what it's worth - initially on ETS the consent checkbox was the only server-side validation to prevent bots from submitting the form / hitting the API without content.

mauriciofreshw (approved):

PR #60, opened 2020-10-16
Hcap 100 baseline update

mauriciofreshw:


fwkendall (approved):

ashtonmeuser (approved):

fresh-rachit (approved):
I'm suggesting we go ahead with validation the way it is right now due to time constraints. The user lands on the Staff Workforce Baseline section and sees a number of validation errors. Not the best user experience, but does not impact functionality and gives the program area what they want in terms of layout. At worst, we can pre-populate all the values with "0" as a workaround if this is release blocking for tomorrow.

PR #68, opened 2020-10-16
Hcap 111 validation optional

fwkendall:
Officially the worst PR

Expected behaviour is that users can leave every field blank, but if they enter values in a field the validation rules apply

### DIFF
--- a/client/src/constants/validation.js
+++ b/client/src/constants/validation.js
@@ -7,6 +7,7 @@ const healthRegions = [
   'Vancouver Coastal',
   'Vancouver Island',
   'Northern',
+  '',
 ];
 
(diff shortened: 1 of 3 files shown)


fresh-rachit (approved):

client/src/constants/validation.js
```diff
@@ -7,13 +7,15 @@ const healthRegions = [
   'Vancouver Coastal',
   'Vancouver Island',
   'Northern',
+  '',
```
fwkendall, line 10: this is an extreme hack to work around wanting to use nullable + oneOf when the value is filled in

ashtonmeuser (approved):

PR #83, opened 2020-10-22
Added stats script

mauriciofreshw:


server/scripts/stats.js
```diff
@@ -0,0 +1,76 @@
...
+        default: dayjs().format('YYYY-MM-DD')
+      }]));
+
+      ({ 'End Date (not inclusive)': endDate } = await inquirer.prompt([{
+        name: 'End Date (not inclusive)',
+        type: 'input',
+        default: dayjs().add(1, 'day').format('YYYY-MM-DD')
+      }]));
+
+      const todayResults = await dbClient.db[collections.EMPLOYER_FORMS]
```
fwkendall, line 44: Should this be called timeSpanResults or something like that?

server/scripts/stats.js
```diff
@@ -0,0 +1,76 @@
...
+
+      console.log('------------------');
+      console.log(`${startDate} to ${endDate} submissions:`)
+      console.table(healthAuthorityCount);
+      console.log(`Total submissions: ${healthAuthorityCount.reduce((total, item) => (item.submissions + total), 0)}`);
+      console.log('------------------');
+
+      return process.exit();
+    } catch (error) {
+      console.error(`Failed to retrive stats, ${error}`);
```
fwkendall, line 72: minor typo, retrieve* :)

server/scripts/stats.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable */
+const { dbClient } = require('../db/db.js');
+const { collections } = require('../db/schema.js');
+const inquirer = require('inquirer');
+const dayjs = require('dayjs');
+const utc = require('dayjs/plugin/utc');
+require('dotenv').config({ path: '../.env' });
```
fwkendall, line 7: Is environment selection toggled via the .env? (e.g. dev/test/prod)

server/scripts/stats.js
```diff
@@ -0,0 +1,76 @@
...
+      let endDate;
+
+      ({ 'Start Date': startDate } = await inquirer.prompt([{
+        name: 'Start Date',
+        type: 'input',
+        default: dayjs().format('YYYY-MM-DD')
+      }]));
+
+      ({ 'End Date (not inclusive)': endDate } = await inquirer.prompt([{
+        name: 'End Date (not inclusive)',
```
fwkendall, line 39: Just making sure I understand this correctly - if I wanted to search records submitted today (2020-10-22) I would enter 2020-10-22 as my start date and 2020-10-23 as my end date?

server/scripts/stats.js
```diff
@@ -0,0 +1,76 @@
...
+  return dayjs(date).utc().format('YYYY-MM-DD HH:mm:ssZ');
+}
+
+/* eslint-disable no-console */
+(async () => {
+  if (require.main === module) {
+    try {
+      await dbClient.connect();
+
+      const healthAuthorities = [
```
fwkendall, line 21: Just a note since I ran the report on prod data last night, we already have some submissions that came in without this field being populated (since we relaxed the validation to allow that to happen). So we might want to add a blank string here as well, just to make it easy to report on those numbers rather than needing to math it out :)

server/scripts/stats.js
mauriciofreshw: The env selection is defined when you create the tunnel only.

server/scripts/stats.js
mauriciofreshw: True. Nice catch.

server/scripts/stats.js
mauriciofreshw: That's right. The end date already has the next day as default, so you just need to hit enter in this step.

fwkendall (approved):

PR #86, opened 2020-10-23
Allow skipping steps forward in form

ashtonmeuser:


client/src/components/employer-form/index.js
```diff
@@ -182,6 +182,22 @@ export const Form = ({ initialValues, isDisabled }) => {
     scrollUp();
   };
 
+  const handleStepClicked = async (index, setTouched, values) => {
+    if (index <= activeStep) { // Can go backwards despite errors
```
fwkendall, line 186: I'm now debating whether this is a good idea... Are they capable of skipping forward, like going from step 1 directly to step 3? Because then we wouldn't check the current fields and could let them miss a validation error

fwkendall (approved):

mauriciofreshw (approved):

fwkendall (approved):

PR #93, opened 2020-11-05
Hcap 150 auth keycloak

mauriciofreshw:
This branch contains a basic example of auth against keycloak and then hitting a protected endpoint `/admin`. The keyclaok admin is located [here](https://dev.oidc.gov.bc.ca/auth/realms/4qjrpzzl/protocol/openid-connect/auth?client_id=security-admin-console&redirect_uri=https%3A%2F%2Fdev.oidc.gov.bc.ca%2Fauth%2Fadmin%2F4qjrpzzl%2Fconsole%2F%23%2Frealms%2F4qjrpzzl%2Fclients&state=b4a8dc1c-5389-4d18-888b-a610e61463ce&response_mode=fragment&response_type=code&scope=openid&nonce=27bf35c8-a22b-4219-81af-1bc8f17dc96a&code_challenge=LBACSX-RbGBx4Y8NuucYDopPwnSe3T9VSI9AV0CmSEQ&code_challenge_method=S256) and all the redirects have the root URL to localhost. In other words, dev is pointing to localhost right now.


.config/.env.example
```diff
@@ -10,3 +10,11 @@ MONGO_PORT=
 MONGO_DB=logs
 MONGO_USER=admin
 MONGO_PASSWORD=
+
+# KEYCLOAK_USER=
```
ashtonmeuser, line 14: Not sure I like commented out vars in the example .env file. It's not obvious when they may be needed. Can you explain the purpose of this @mauriciofreshw?

.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules
 client/.pnp
 client/.pnp.js
 client/build
+!client/.env
```
ashtonmeuser, line 8: Only one of these changes is required to keep `client/.env` tracked.

client/.env
```diff
@@ -0,0 +1,3 @@
+REACT_APP_KEYCLOAK_REALM=${KEYCLOAK_REALM}
```
ashtonmeuser, line 1: If the env vars on the right (`KEYCLOAK_REALM`, `KEYCLOAK_AUTH_URL`, `KEYCLOAK_FE_CLIENTID`) are already required, why don't we just use those in `keycloak.js`? As far as I can tell, this file is simply mapping the names of env vars, which could be completely avoided.

client/src/keycloak.js
```diff
@@ -0,0 +1,9 @@
+import Keycloak from 'keycloak-js';
+
+const keycloak = new Keycloak({
+  realm: process.env.REACT_APP_KEYCLOAK_REALM,
```
ashtonmeuser, line 4: See previous comment. We could use the original names (`KEYCLOAK_REALM`, `KEYCLOAK_AUTH_URL`, `KEYCLOAK_FE_CLIENTID`) here.

.config/.env.example
mauriciofreshw: This env var is being used in the keycloak container that’s also commented out. So if someone intends to use keyclaok fully locally those env var will be necessary.

client/.env
mauriciofreshw: Env vars on the client side must start with `REACT_APP_` according to the [docs](https://create-react-app.dev/docs/adding-custom-environment-variables/) unless you eject the app, so if we want to share the same var between `server` and `client` we have to map to that prefix first.

.gitignore
mauriciofreshw: Not sure if I understood your comment, I tried `.env` for the root one, and didn’t work. I suppose if you don't specify `/` before the filename it will ignore all occurrences in the project.

.docker/keycloak/local.json
```diff
@@ -0,0 +1,2522 @@
...
+      "clientId": "hcap-api",
+      "name": "Health Career Access Program - API",
+      "rootUrl": "http://localhost:8080",
+      "adminUrl": "http://localhost:8080",
+      "baseUrl": "http://localhost:8080",
+      "surrogateAuthRequired": false,
+      "enabled": true,
+      "alwaysDisplayInConsole": false,
+      "clientAuthenticatorType": "client-secret",
+      "secret": "**********",
```
fwkendall, line 735: This is weird, does this get replaced somehow at build/runtime?

.config/.env.example
fwkendall: The .env is really only used locally right? Seems like we could uncomment them

client/src/constants/routes.js
```diff
@@ -1,6 +1,8 @@
 export default Object.freeze({
   // Public routes
   Login: '/login',
+  Admin: '/admin',
```
fwkendall, line 4: Tiny nitpick but would love to move this to its own block commented for private routes vs. being in with the public ones here :)

client/src/pages/private/Admin.js
```diff
@@ -0,0 +1,40 @@
...
+        'Accept': 'application/json',
+        'Content-type': 'application/json',
+        'Authorization': `Bearer ${store.get('TOKEN')}`,
+      },
+      method: 'GET',
+    });
+
+    if (response.ok) {
+      const { result } = await response.json();
+      console.log(result);
```
fwkendall, line 22: Should remove this now or make a note somewhere to make sure we come back to this before EOW :)

client/src/pages/private/Admin.js
```diff
@@ -0,0 +1,40 @@
...
+    const response = await fetch('/api/v1/admin', {
+      headers: {
+        'Accept': 'application/json',
+        'Content-type': 'application/json',
+        'Authorization': `Bearer ${store.get('TOKEN')}`,
+      },
+      method: 'GET',
+    });
+
+    if (response.ok) {
```
fwkendall, line 20: How should we handle the else for this?

client/src/pages/public/Keycloak.js
```diff
@@ -0,0 +1,27 @@
...
+  }, [keycloak.authenticated, keycloak.token]);
+
+
+  return keycloak.authenticated
+    ?
+    <Redirect to={store.get('REDIRECT')} />
+    :
+    keycloak.loginRequired
+      ?
+      <Redirect to={Routes.Login} />
```
fwkendall, line 24: I guess this actually takes care of the 401 page :)

fwkendall: Although maybe not, I guess once we add user roles there will be a little more nuance where some screens might be inaccessible even to logged-in users, depending on their role

docker-compose.dev.yml
```diff
@@ -45,6 +53,30 @@ services:
     networks:
       - network
 
+  # keycloak:
```
fwkendall, line 56: this docker-compose is only for local dev so should we uncomment this now?

server/server.js
```diff
@@ -30,9 +31,18 @@ app.use(helmet({
...
 app.use(express.static(path.join(__dirname, '../client/build')));
 
+app.get(`${apiBaseUrl}/admin`,
+  keycloak.protect('admin'),
+  asyncMiddleware(async (req, res) => {
+    const result = 'hello admin';
+    return res.json({ result });
+  }));
+
 // Login endpoint
```
fwkendall, line 46: I guess we can remove this now?

client/src/pages/private/Admin.js
mauriciofreshw: This component no longer exists in my next PR that is a superset of this one

client/src/pages/public/Keycloak.js
mauriciofreshw: The login screen will be centralized anyway, we can although pass parameters to the Login component in the future

docker-compose.dev.yml
mauriciofreshw: only local but we can’t test BCeID and IDIR

server/server.js
mauriciofreshw: Removed in my current PR :)

ashtonmeuser (approved):

client/src/pages/private/Admin.js
fwkendall: Awesome!

client/src/pages/private/Admin.js
fwkendall: Addressed on subsequent PR

ashtonmeuser (approved):

fwkendall (approved):

PR #104, opened 2020-11-14
Hcap 175 view eoi

mauriciofreshw:


server/server.js
```diff
@@ -93,19 +106,13 @@ app.post(`${apiBaseUrl}/employees`,
...
-        roles.push(item);
-      }
-    });
-    req.roles = roles;
-    return !token.isExpired();
-  }),
+  keycloak.protect(),
   asyncMiddleware(async (req, res) => {
+    const { roles } = req.kauth
+      .grant.access_token.content
```
mauriciofreshw, line 112: @fwkendall `req.kauth.grant.access_token.content` contains a lot of info including the user first and last name. I just didn’t figure out yet how to get values from the Attributes, I created a token mapper but seems not working.

PR #108, opened 2020-11-17
Added additional roles and role-based permissions to FE

fwkendall:


server/server.js
```diff
@@ -40,6 +40,7 @@ app.use(bodyParser.json());
 app.use(express.static(path.join(__dirname, '../client/build')));
 
 const allowRoles = (...roles) => (token) => {
+  roles.push('superuser');
```
ashtonmeuser, line 43: Nitpick: Haven't checked, but the linter may complain about overwriting a param. We could use `if (![...roles, 'superuser'].some((role) => token.hasRole(role))) return false;` if this is the case.

ashtonmeuser (dismissed):

ashtonmeuser:
Looks like we're wrapping *all* private components in the `CheckPermissions` component. Is this correct? Should all private pages contain the wrapper from now on?

server/server.js
fwkendall: I could write this in a more clear way too, which would be fair:
The idea being that superusers can do everything - so check if the token hasRole superuser and return true if so, if not then continue on to checking the params

fwkendall:
> Looks like we're wrapping _all_ private components in the `CheckPermissions` component. Is this correct? Should all private pages contain the wrapper from now on?

Yep, I think so. The route-level protection just checks that a user is authenticated. For role-based permissions I think we're going to need this to wrap everything private because of how the IDP works - someone can login with their BCeID or IDIR right away, just without a role, so we need to control what authenticated users without a role have access to as they may be going to a direct link shared with them prior to role assignment.

client/src/components/generic/CheckPermissions.js
```diff
@@ -0,0 +1,27 @@
...
+import { checkPermissions } from '../../utils';
+
+export const CheckPermissions = ({ roles, permittedRoles, children, renderMessage = false }) => {
+  // If user has a permitted role assigned, grant access to components
+  if (checkPermissions(roles, permittedRoles)) {
+    return children;
+  }
+
+  // Optional: display message if no access granted by role(s)
+  if (renderMessage) {
```
mauriciofreshw, line 13: Maybe `renderErrorMessage` would be more appropriate here?

client/src/pages/private/Admin.js
```diff
@@ -51,33 +53,26 @@ export default () => {
...
   return (
     <Page >
-      <Grid container alignContent="center" justify="center" alignItems="center" direction="column">
-        <Box pb={4} pl={4} pr={4} pt={2}>
-          <Grid container direction="column">
-            {
-              roles.length === 0 ?
-                <Typography variant="subtitle1" gutterBottom>
-                  You don't have enough permissions.
+      <CheckPermissions roles={roles} permittedRoles={['maximus', 'employer', 'health_authority', 'ministry_of_health']} renderMessage={true}>
```
mauriciofreshw, line 56: I think in the next PRs we should centralize those role names constants, I'm starting to feel the pain of changing string occurrences around the code xD

client/src/pages/private/Admin.js
```diff
@@ -51,33 +53,26 @@ export default () => {
...
-                      default:
-                        return null
-                    }
-                  })
-            }
-          </Grid>
-        </Box>
-      </Grid>
+              <CheckPermissions roles={roles} permittedRoles={['maximus']}>
+                { renderAdminButton(0, Routes.ApplicantUpload, 'Upload Applicants') }
```
mauriciofreshw, line 64: I'm not 100% sure, but I think this first parameter is no longer necessary since these components have their own parent now.

client/src/pages/private/Admin.js
fwkendall: Yeah I was trying to figure out a good way to manage constants in general, since we have so many that need to be accessed by both the FE and BE... It seems very unwieldy to pass them in via env/make/docker since they aren't secrets or config.

mauriciofreshw (approved):

client/src/components/generic/CheckPermissions.js
fwkendall: Coming next commit! :)

client/src/pages/private/Admin.js
fwkendall: Looks correct from my testing :) thanks!

mauriciofreshw (approved):

ashtonmeuser (approved):

PR #109, opened 2020-11-17
Add applicant upload screen, change upload response format

ashtonmeuser:


client/src/pages/private/ApplicantUploadResults.js
```diff
@@ -0,0 +1,86 @@
...
+        <Grid container alignContent="center" justify="center" alignItems="center" direction="column">
+          <Box pt={4} pb={4} pl={2} pr={2}>
+            <Typography variant="subtitle1" gutterBottom>
+              Applicant Upload Results
+            </Typography>
+          </Box>
+          <Box pb={4} pl={2} pr={2}>
+            <Button
+                onClick={() => history.push(Routes.ApplicantUpload)}
+                text="Return to Application Upload"
```
fwkendall, line 70: Should this say applicant instead of application? I guess this also makes sense so not sure if it was deliberate :)

fwkendall (approved):

mauriciofreshw (approved):

PR #110, opened 2020-11-17
Added migrations

mauriciofreshw:


server/main.js
```diff
@@ -32,10 +34,25 @@ process.on('SIGTERM', () => {
...
+  try {
+    const { stdout, stderr } = await exec(`DATABASE_URL=${dbUrl} npm run migrate up`);
+
+    if (stderr) {
+      throw Error(`Migration error: ${stderr}`);
+    } else {
+      logger.info(`Migration success: ${stdout}`);
+    }
+  } catch (err) {
+    throw Error(`Migration error: ${err}`);
```
fwkendall, line 47: So I'm thinking how can we test this? Obviously we don't expect migration errors, but I'm wondering how we validate detection/recovery during a deployment

mauriciofreshw: If the migration fails the server won't start and you'll be able to see the error direct on the application console or over the log database.

ashtonmeuser: The logging to Slack functionality from ETS is _exremely_ simple. We could port that over to HCAP. I'm sure there are other uses beyond this case. Alternatively, we just make sure to check the Mongo logging DB regularly.

fwkendall: Presumably we would only be running migrations during a deployment, when we would be monitoring things anyway... We just need to be sure we're checking the logs (or add alerting) in case we do end up in a rollback scenario on a failed migration

fwkendall (approved):

ashtonmeuser (approved):

ashtonmeuser:
Tested this after clearing the Postgres volume and confirming the index on `maximusId` is created. Great job, @mauriciofreshw 👏 this was a huge QoL improvement for the app.

PR #115, opened 2020-11-19
Filter HA regions

mauriciofreshw:


server/server.js
```diff
@@ -73,7 +105,13 @@ app.get(`${apiBaseUrl}/employer-form`,
   keycloak.protect(allowRoles('employer', 'health_authority', 'ministry_of_health')),
   asyncMiddleware(async (req, res) => {
     try {
-      const result = await dbClient.db[collections.EMPLOYER_FORMS].findDoc({});
+      const roles = getUserRoles(req);
+      const isHA = roles.includes('health_authority');
+      const isSuperUser = roles.includes('superuser');
+      const criteria = isHA && !isSuperUser ? getUserRegionsCriteria(req, 'healthAuthority') : {};
```
fwkendall, line 111: I think this logic can be a little simpler, and also needs to include MOH

`const criteria = isSuperUser || isMOH ?  {} : getUserRegionsCriteria(req, 'healthAuthority');`

Basically:
- if superuser OR ministry_of_health, don't filter by region
- For all other cases (currently just HAs but will soon include employers), filter by region
:)

fwkendall (approved):

server/server.js
```diff
@@ -73,7 +105,13 @@ app.get(`${apiBaseUrl}/employer-form`,
   keycloak.protect(allowRoles('employer', 'health_authority', 'ministry_of_health')),
   asyncMiddleware(async (req, res) => {
     try {
-      const result = await dbClient.db[collections.EMPLOYER_FORMS].findDoc({});
+      const roles = getUserRoles(req);
+      const isMOH = roles.includes('ministry_of_health');
+      const isSuperUser = roles.includes('superuser');
+      const criteria = isSuperUser || isMOH ? {} : getUserRegionsCriteria(req, 'healthAuthority');
```
ashtonmeuser, line 111: Not sure what `criteria` is supposed to look like and I think we're getting into some dangerous territory using it in a condition/ternary/whatever. As far as I can tell, `criteria` can be either an empty object (`{}`), `null`, or an object with a single property `or` (which is an array). These evaluate to `true`, `false`, and `true`, respectively. Not sure if this is expected, but it certainly isn't obvious at first glance e.g. `{}` used for its truthiness.

ashtonmeuser: Upon further inspection, looks like `criteria` is basically a query. This kinda makes sense now. `{}` and `{or: [...]}` are valid queries and should results in a DB hit, `null` should not and therefore sets results to `[]`.

ashtonmeuser (approved):

PR #122, opened 2020-11-24
Table pagination

ashtonmeuser:


mauriciofreshw (approved):

client/src/components/generic/Table.js
```diff
@@ -46,44 +47,66 @@ const StyledTableRow = withStyles((theme) => ({
   },
 }))(TableRow);
 
-export const Table = ({ order, orderBy, onRequestSort, columns, rows, isLoading }) => {
+export const Table = ({ order, orderBy, onRequestSort, columns, rows, isLoading, rowsPerPage = 10 }) => {
```
fwkendall, line 50: I think we're expecting people to use this on desktop - how does 20 look for the default? 10 seems a little low to m but not sure how it looks in practice

fwkendall (approved):

client/src/components/generic/Table.js
fwkendall: Resolved offline - 10 looks good

mauriciofreshw (approved):

fwkendall (approved):

PR #128, opened 2020-11-27
Wrap keycloak-connect

ashtonmeuser:
Not to be merged just yet. This is merely a preview of the wrapper around the [Keycloak connect library](https://www.npmjs.com/package/keycloak-connect). Simplifies getting user info (name, roles, region, etc.) using the connect middleware `getUserInfo()` method. Also saves realm and client info so we don't have to pull from `process.env` all the time.


server/keycloak.js
```diff
@@ -0,0 +1,81 @@
...
+      if (roles.length === 0) return true; // Allows any role
+      if (token.hasRole('superuser')) return true;
+      if (!roles.some((role) => token.hasRole(role))) return false;
+      if (token.isExpired()) return false;
+      return true;
+    };
+    return this.keycloakConnect.protect(allowRoles);
+  }
+
+  getUserInfo() { // Connect middleware for adding HCAP user info to request object
```
mauriciofreshw, line 61: We should call this `getUserInfoMiddleware()` since it's returning a middleware and not the user info per se.

ashtonmeuser: Will do!

openshift/keycloak.prep.yml
```diff
@@ -34,6 +42,9 @@ objects:
     stringData:
       realm: ${KEYCLOAK_REALM}
       auth-url: ${KEYCLOAK_AUTH_URL}
-      fe-client-id: ${APP_NAME}${KEYCLOAK_FE_CLIENTID_SUFFIX}
+      fe-client-name: ${APP_NAME}${KEYCLOAK_FE_CLIENTID_SUFFIX}
```
fwkendall, line 45: doesn't this name need to be changed in server.dc.yml as well?

server/keycloak.js
```diff
@@ -0,0 +1,122 @@
...
+    };
+  }
+
+  expressMiddleware(...args) { // Default connect middleware for Keycloak connect library
+    return this.keycloakConnect.middleware(...args);
+  }
+
+  allowRolesMiddleware(...roles) { // Connect middleware for limiting roles
+    const allowRoles = () => (token) => {
+      if (roles.length === 0) return true; // Allows any role
```
fwkendall, line 53: is this temporary?

mauriciofreshw (approved):

server/keycloak.js
ashtonmeuser: This was to match the way that we were using `keycloak.protect()`. If no roles are provided, it just makes sure that the user is logged in. We could change this and all endpoints that depend on it. Looks like it's only the `/user` API endpoint that uses it.

openshift/keycloak.prep.yml
ashtonmeuser: Good catch. Fixed.

mauriciofreshw (approved):

server/server.js
```diff
@@ -112,29 +87,26 @@ app.post(`${apiBaseUrl}/employees`,
...
-    }
-
+    if (req.fileError) return res.json({ status: 'Error', message: req.fileError });
     const response = await parseAndSaveParticipants(req.file);
     return res.json(response);
   }));
 
+// Get pending users from Keycloak
+app.get(`${apiBaseUrl}/pending-users`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
```
fwkendall, line 97: should be just MOH here actually, HA approval was descoped

ashtonmeuser: Fixed.

fwkendall (dismissed):

mauriciofreshw (approved):

server/keycloak.js
```diff
@@ -0,0 +1,121 @@
...
+    this.access_token = response.data.access_token;
+  }
+
+  async authenticateIfNeeded() {
+    // Race condition if token expires between this call and the desired authenticated call
+    const config = { headers: { Authorization: `Bearer ${this.access_token}` } };
+    try {
+      await axios.get(`${this.authUrl}/realms/${this.realm}/protocol/openid-connect/userinfo`, config);
+    } catch (error) {
+      await this.authenticateServiceAccount();
```
fwpushan, line 96: As a suggestion, I recommend, do a little bit digging on error before re-authenticate, like check is it timeout / network issue / 401.

server/keycloak.js
```diff
@@ -0,0 +1,121 @@
...
+
+const defaults = {
+  'ssl-required': 'external',
+  'confidential-port': 0,
+  'verify-token-audience': true,
+  'use-resource-role-mappings': true,
+  'policy-enforcer': {},
+};
+
+class Keycloak { // Wrapper class around keycloak-connect
```
fwpushan, line 21: This is nice to have Keycloak manager like this. It helps us to Mock this class for testing.

fwkendall (approved):

mauriciofreshw (approved):

fwpushan (approved):

PR #135, opened 2020-11-30
[HCAP-224] Refactor and transform participants data. 

fwpushan:
Update participates validation schema, processing and parsing.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -116,8 +116,12 @@ app.post(`${apiBaseUrl}/employees`,
       return res.json({ status: 'Error', message: req.fileError });
     }
 
-    const response = await parseAndSaveParticipants(req.file);
-    return res.json(response);
+    try {
+      const response = await parseAndSaveParticipants(req.file);
+      return res.json(response);
+    } catch (excp) {
+      return res.status(400).send(`${excp}`);
+    }
   }));
 
 // Get user info from token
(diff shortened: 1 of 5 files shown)


mauriciofreshw:
> Update participates validation schema, processing and parsing.

@fwpushan I removed the old lib and added a new one more optimized. The upload should work now.

mauriciofreshw (approved):

server/server.js
```diff
@@ -116,8 +116,12 @@ app.post(`${apiBaseUrl}/employees`,
       return res.json({ status: 'Error', message: req.fileError });
     }
 
-    const response = await parseAndSaveParticipants(req.file);
-    return res.json(response);
+    try {
+      const response = await parseAndSaveParticipants(req.file);
+      return res.json(response);
+    } catch (excp) {
+      return res.status(400).send(`${excp}`);
```
ashtonmeuser, line 123: Just so we're all aware, we can throw inside the `asyncMiddleware`. The error will be presented as a generic 500 error in `error-handler.js`. If we see a whole bunch of these try/catch blocks, we might want to consider reworking the error handler.

ashtonmeuser (approved):

server/server.js
fwpushan: OK, but I tried to throw error but it is not showing the appropriate error message.

fwpushan: I got it I will fix it different PR.

PR #142, opened 2020-12-02
removed nonHCAP and interested sanitization

mauriciofreshw:


fwkendall (approved):

ashtonmeuser (approved):

fwkendall:
We confirmed that Massive will store a blank value here as simply not creating a key on the jsonb object. If we need this to be present later (for querying, reporting, or for editing purposes) then we can write a migration to add this key with a null value for any records that don't have it.

PR #144, opened 2020-12-03
Participant service tests

mauriciofreshw:


fwpushan (approved):
- [X] Tests passed locally 
- [X] Tests passed in GithubAction
- [X]  Code changes look good.

server/services/participants.js
```diff
@@ -6,8 +6,7 @@ const { dbClient, collections } = require('../db');
 const { userRegionQuery } = require('./user.js');
 
 const getParticipants = async (user) => {
-
-  const criteria = user.isSuperUser || user.isMOH ? {} : userRegionQuery(user.regions, 'preferredLocation');
+  const criteria = user.isSuperUser || user.isMoH ? {} : userRegionQuery(user.regions, 'preferredLocation');
```
fwkendall, line 9: Checking against master I think this is right, but also looks like this PR needs to be rebased :)

server/db/db.js
```diff
@@ -23,7 +23,7 @@ class DBClient {
     this.db = null;
   }
 
-  async runMigration() {
+  async runMigration(useTestDb) {
```
ashtonmeuser, line 26: I think we should try to avoid knowledge of tests in implementation. Ideally, this kind of thing would be set by the testing framework or test preflight setup and the actual `runMigrations` method is oblivious.

server/tests/participant.form.test.js
```diff
@@ -0,0 +1,242 @@
...
+describe('Participants Service', () => {
+  beforeAll(async () => {
+    await startDB();
+  });
+
+  afterAll(async () => {
+    await closeDB();
+  });
+
+  const allParticipants = [
```
ashtonmeuser, line 17: Might make sense to keep this in a JSON file in the same location as our test .xlsx fikes.

ashtonmeuser (dismissed):

fwkendall:
Looks like a great foundation we can build on for more tests, thanks Mauricio! :)

server/db/db.js
fwkendall: If possible, I agree

mauriciofreshw: Yeah, makes sense totally. In the beginning, we were not using a makefile to run the tests in the container, and therefore we couldn’t have access to the NODE_ENV.

fwkendall (approved):

server/tests/participant.form.test.js
mauriciofreshw: I feel that we do too many tweaks over the test data currently to adapt to the tests result, we can add that json for upcoming PRs if the sheet won’t change too much

ashtonmeuser (approved):

PR #148, opened 2020-12-04
Hcap 187 approve access request

ashtonmeuser:
BE route for pending access request approval. Takes user ID (set by Keycloak), region, role (only MoH for now).

Keycloak class takes user ID, array of regions, and a role.

FE has a dialog form that allows choosing a single region and single role for a pending user and approving said user. I'm sure the Dialog component can be improved. Specifically, I think that either the Dialog component or a new component should take some of the responsibility from the UserView component which is getting pretty heavy.

The Table component was updated to improve the way it lays out its cells. Previously, it iterated through the keys of a row item, adding each as a contiguous cell. Now, it iterated through the ID of each header cell and accesses properties of each row item using this key. This allows for row items with an arbitrary order and properties that are not displayed in the table. Additionally, columns can no longer be used to order the table if they do not include a label (see Employers table Details column in previous builds). Finally, Table now handles missing cell values and places `''` (empty string) when cell value is missing.

Validation updated. Now `ParticipantFormSchema.preferredLocation` does not accept `''` (empty string).

### DIFF
--- a/client/src/components/generic/Table.js
+++ b/client/src/components/generic/Table.js
@@ -73,12 +73,12 @@ export const Table = ({ order, orderBy, onRequestSort, columns, rows, isLoading,
           <TableRow>
             {columns.map((column, index) => (
               <StyledHeaderTableCell key={index}>
-                <TableSortLabel
+                {column.name && <TableSortLabel // Disable sorting if column has no header
                   active={orderBy === column.id}
                   direction={orderBy === column.id ? order : 'asc'}
                   onClick={createSortHandler(column.id)}>
                   {column.name}
-                </TableSortLabel>
+                </TableSortLabel>}
               </StyledHeaderTableCell>
             ))}
           </TableRow>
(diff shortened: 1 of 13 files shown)


client/src/components/generic/Table.js
```diff
@@ -92,8 +92,8 @@ export const Table = ({ order, orderBy, onRequestSort, columns, rows, isLoading,
             </StyledTableRow>
           )) : pageRows.map((row, index) => (
             <StyledTableRow hover key={index}>
-              {Object.keys(row).map((key) => (
-                <StyledTableCell key={key}>{row[key]}</StyledTableCell>
+              {columns.map((column) => (
+                <StyledTableCell key={column.id}>{row[column.id] || ''}</StyledTableCell>
```
ashtonmeuser, line 96: @mauriciofreshw I'm not sure where you added the logic for empty cells, but this should handle it now. So we can send an array of any object to a table and the props should align with columns as well as empty cells being handled correctly.

mauriciofreshw: The value when null was being set as '', but the way you did here is better.

mauriciofreshw (dismissed):

client/src/components/generic/Table.js
ashtonmeuser: Yeah, I figured we'd be doing the same thing everywhere that Table is used. This will save us from repeating ourselves.

client/src/pages/private/UserView.js
```diff
@@ -83,19 +104,75 @@ export default () => {
...
+                  { value: 'Vancouver Island', label: 'Vancouver Island Health' },
+                  { value: 'Northern', label: 'Northern Health' },
+                ]}
+              />
+              <FastField
+                name="role"
+                component={RenderSelectField}
+                label="* User Role"
+                options={[
+                  { value: 'ministry_of_health', label: 'Ministry of Health' },
```
fwkendall, line 150: This should be HA right?

fwkendall (approved):

mauriciofreshw (approved):

PR #153, opened 2020-12-08
Hcap 173 employer site entity

mauriciofreshw:
Refactored @fwpushan branch with the new requirements


Makefile
```diff
@@ -45,13 +45,24 @@ local-close:
 	@echo "Stopping local app container"
 	@docker-compose -f docker-compose.dev.yml down
 
+local-clean:
```
fwkendall, line 48: Is this in here twice now...?

server/tests/employer.form.test.js
```diff
@@ -17,6 +19,25 @@ describe('Server V1 Form Endpoints', () => {
 
   const formEndpoint = '/api/v1/employer-form';
 
+  const site = {
+    siteId: 67,
+    siteName: 'Test site',
+    earlyAdaptorAllocation: 1,
```
fwkendall, line 25: minor typo - this should be adopter, with an o :)

server/tests/employer.form.test.js
```diff
@@ -150,4 +171,35 @@ describe('Server V1 Form Endpoints', () => {
...
+
+  it('Create new site, receive success', async () => {
+    const res = await saveSites(site);
+    const expectedRes = [
+      { siteId: 67, status: 'Success' },
+    ];
+    expect(res).toEqual(expectedRes);
+  });
+
+  it('Get sites, receive all successfully', async () => {
```
fwkendall, line 183: I'm not super familiar with this testing framework, do the tests always run in order?

server/validation.js
```diff
@@ -154,31 +155,56 @@ const EmployerFormSchema = yup.object().noUnknown('Unknown field in form').shape
   doesCertify: yup.boolean().typeError(errorMessage).required(errorMessage).test('is-true', errorMessage, (v) => v === true),
 });
 
+const EmployerSiteBatchSchema = yup.array().of(
+  yup.lazy((item, options) => {
+    const index = options.parent.indexOf(item);
+    return yup.object().noUnknown(`Unknown field in site data (index ${index})`).shape({
```
fwkendall, line 161: Just a note on some fields that I think are required and not nullable:

- siteId
- siteName
- healthAuthority
- postalCode

I think too many other parts of the application hinge on these fields to allow them to be nullable. We need the site name in order to show it in the access approval dropdown, and we need the HA + postal code for region access controls and sorting/searching later

fwkendall:
Nice work, especially appreciate the test coverage! Just a few comments from me :)

server/tests/employer.form.test.js
mauriciofreshw: Yes, we use `--runInBand` when running the test command

Makefile
mauriciofreshw: Good catch!

fwkendall (approved):

Zenen-freshworks (approved):
Looks good to me!

PR #155, opened 2020-12-08
Add employer role to access request approval

ashtonmeuser:


client/src/pages/private/UserView.js
```diff
@@ -149,6 +149,7 @@ export default () => {
                 label="* User Role"
                 options={[
                   { value: 'health_authority', label: 'Health Authority' },
+                  { value: 'employer', label: 'Employer' },
```
fwkendall, line 152: I wonder if we should use the label "Private Employer" here?

ashtonmeuser: Updated

mauriciofreshw (approved):

Zenen-freshworks (approved):

fwkendall (approved):

PR #156, opened 2020-12-09
Hcap 183 select sites request

mauriciofreshw:
This branch also contains a script to parse and feed the DB with the employer sites


server/scripts/feed-sites.js
```diff
@@ -0,0 +1,50 @@
...
+      const columnMap = {
+        'HCAP Site ID': 'siteId',
+        'Site Name': 'siteName',
+        'Early Adopter Allocation': 'earlyAdopterAllocation',
+        'Street Address': 'address',
+        'Health Authority': 'healthAuthority',
+        City: 'city',
+        'Post Code': 'postalCode',
+        'Registered Business Name': 'registeredBusinessName',
+        'Operator Contact First Name': 'operatorContactFirstName',
```
fwkendall, line 21: There's a missing field here, Operator name. This is actually a separate field to first/last since it might be a company name and not a person :)

fwkendall:
Code looks good 👍 I just noticed that we're missing one of the fields for sites - I hope I didn't miss this on the earlier PR also lol

fwkendall (approved):

client/src/constants/validation.js
```diff
@@ -88,6 +88,7 @@ export const LoginSchema = yup.object().noUnknown().shape({
 });
 
 export const ApproveAccessRequestSchema = yup.object().noUnknown().shape({
+  sites: yup.array().required('Employer Sites is required').min(1, 'At least 1 employer site is required'),
```
ashtonmeuser, line 91: Should this only be required for Private Employer? Should only `region` be required for Health Authority role?

server/validation.js
```diff
@@ -211,6 +212,7 @@ const ParticipantBatchSchema = yup.array().of(
 
 const AccessRequestApproval = yup.object().noUnknown('Unknown field in form').shape({
   userId: yup.string().required('User ID is required'),
+  sites: yup.array().required('Employer Sites is required').min(1, 'At least 1 employer site is required'),
```
ashtonmeuser, line 215: Same as above.

client/src/constants/validation.js
fwkendall: Nope, required for both

ashtonmeuser (approved):

fwkendall (approved):

PR #158, opened 2020-12-10
Hcap 249 user entity

ashtonmeuser:
Dependant on #156 
Upgrades to Node 14 (LTS) to support some convenience methods like optional chaining
Fetches the user object in `getUserInfoMiddleware` and adds the `sites` property to the `req.hcapUserInfo` object


server/migrations/1607381259543_update-schema.js
```diff
@@ -1,25 +1,12 @@
...
+    await dbClient.createDocumentTableIfNotExists(schemaItem.collection);
     for (const index of schemaItem.indexes) {
-      try {
-        await dbClient.db.query(index);
-      } catch (err) {
-        if (err.code !== ERROR_DUPLICATED) {
-          throw err;
-        }
-      }
+      dbClient.createIndexIfNotExists(schemaItem.collection, index);
```
mauriciofreshw, line 9: We can actually get rid of all update-schema.js migrations if we just run those functions on every server startup (on the migration function, before db.reload();) what do you think?

ashtonmeuser: Hmmm this is a good question. My vote would be to avoid that under normal circumstances but it made me think about how our migrations are being performed. We're not doing things quite by the book as migrations should be able to walk through each state of the database over time. Instead, we go straight to the most up-to-date schema. An intermediary migration that relies on a different database state would actually fail.

For example, we've got the rename applicants table migration. Starting the app on fresh infra (all migrations should run) will basically skip this migration as the initial schema migration will update things to after there being a `applicants` table. This isn't a big deal in this example but if we were modifying the table structure i.e. formatting phone numbers, this would be skipped.

Zenen-freshworks (approved):

mauriciofreshw (approved):

fwkendall (approved):

PR #161, opened 2020-12-11
Hcap 247 participant status

mauriciofreshw:
This branch has the foundation of how we gonna handle relational data joins. I created two service functions as a POC that actually can be used if we all agree with this approach.


server/services/participants.js
```diff
@@ -6,12 +6,81 @@ const { dbClient, collections } = require('../db');
 const { createRows, verifyHeaders } = require('../utils');
 const { userRegionQuery } = require('./user.js');
 
-const getParticipants = async (user) => {
+const setParticipantStatus = async (employerId, participantId, status) => {
+  await dbClient.db[collections.PARTICIPANTS_STATUS].update({
```
ashtonmeuser, line 10: I've got some concerns about this not being in a transaction. If the first write succeeds and second write fails, we'll end up with no active statuses. I think we could either
1. Wrap these actions in a transaction
2. Use timestamps to mark which status is active (most recent)

server/services/participants.js
```diff
@@ -6,12 +6,81 @@ const { dbClient, collections } = require('../db');
...
+
+    participantsStatusMap.set(
+      item.participants_status__participant_id,
+      participantStatus ? [...participantStatus, item] : [item],
+    );
+  });
+
+  const participants = [];
+
+  prticipantsMap.forEach((item, participantId) => {
```
ashtonmeuser, line 43: Is there no way to do a join? The current implementation performs a query for each participant, right?

ashtonmeuser: Ignore this. I was mistaken.

mauriciofreshw: The join is in line 67. This function is necessary because we can have repeated participants in the output due to the combination with the `participants_status` table. For each `participantId` we have a join with the participant table, and therefore an employer can engage multiple participants.

server/services/participants.js
mauriciofreshw: Good catch, we actually can’t use the 2. approach here due to the limitation of massive.js and also because we would need to perform a scan over the dates to get the most recent one, which could lead to performance issues when doing the join (pretty sure that there's an optimized way to do that using raw query but that requires too much time googling and testing).

server/services/participants.js
```diff
@@ -6,12 +6,86 @@ const { dbClient, collections } = require('../db');
...
+  return tx[collections.PARTICIPANTS_STATUS].save({
+    employer_id: employerId,
+    participant_id: participantId,
+    status,
+    current: true,
+  });
+});
+
+const flatJoinedParticipantStatus = (raw) => {
+  const prticipantsMap = new Map();
```
fwkendall, line 29: small typo here :)

ashtonmeuser (approved):

fwkendall (approved):

PR #163, opened 2020-12-11
Hcap 237

Zenen-freshworks:
Running the test suite is as simple as "make local-cypress-test".

In order to work properly, Cypress needs the values in cypress.env.json to be filled out and sitting in the project's root directory (same place as the .env). I can share the file I use as needed.


cypress/support/commands.js
```diff
@@ -0,0 +1,90 @@
...
+    Cypress.log({ name: "Logout" });
+    let authBaseUrl = Cypress.env("auth_base_url");
+    let realm = Cypress.env("auth_realm");
+    return cy.request({
+        url: authBaseUrl + "/realms/" + realm + "/protocol/openid-connect/logout"
+    });
+});
+
+Cypress.Commands.add("kcNavAs", function (user, visitUrl) {
+  if (visitUrl === void 0) { visitUrl = ""; }
```
ashtonmeuser, line 41: What's this checking for? If just wanting to make sure `visitUrl` is defined, we could probably use something simpler i.e. `visitUrl = visitUrl || ""`. This may raise linting errors if we apply linting, which I would recommend, as it is modifying a function param.

cypress/support/commands.js
```diff
@@ -0,0 +1,90 @@
...
+    session_state: utils_1.createUUID(),
+    scope: "openid"
+  };
+
+  let localStorageObj = {
+    state: state,
+    nonce: nonce,
+    expires: Date() + 3600
+  };
+  let localStorageKey = "kc-callback-" + state;
```
ashtonmeuser, line 65: I'd suggest we apply linting in the Cypress dir. Will avoid this kind of thing. We would typically use constants and template literals for this kind of op i.e. `const localStorageKey = `kc-callback-${state}`;

cypress/support/commands.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    if (userType === "employer") return {
+      name: "Test Employer",
+      roles: [ "employer" ]
+    }
+  }
+
+  window.localStorage.setItem(localStorageKey, JSON.stringify(localStorageObj));
+  cy.intercept("post", authBaseUrl + "/realms/" + realm + "/protocol/openid-connect/token", token);
```
ashtonmeuser, line 82: Again, template literals preferred.

cypress/support/commands.js
```diff
@@ -0,0 +1,90 @@
...
+  cy.intercept("get", authBaseUrl + "/realms/" + realm + "/protocol/openid-connect/auth", userInfo(user));
+
+  // in case visitUrl is an url with a hash, a second hash should not be added to the url
+  let joiningCharacter = visitUrl.indexOf("#") === -1 ? "#" : "&";
```
ashtonmeuser, line 86: The internal Node [URL package](https://nodejs.org/api/url.html) could help us avoid this logic.

cypress/support/utils.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+function createUUID() {
```
ashtonmeuser, line 3: I'm actually wondering why we need to generate a semi-legit UUID v4 for the app state if we're just faking things anyway. We're already intercepting many calls to Keycloak (see `cy.intercept()` returning a token and user info) so the state token doesn't really have to change or even be legit.

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "devDependencies": {
+    "cypress": "^6.1.0"
+  },
+  "dependencies": {
+    "puppeteer": "^5.5.0"
+  }
+}
```
ashtonmeuser, line 8: If this is deps for Cypress, I'd vote for creating the Node package in `/cypress` rather than root project dir.

.config/cypress.env.json.example
```diff
@@ -0,0 +1,7 @@
+{
+  "kc_sa_pass": "",
+  "kc_api_secret": "",
+  "access_token": "",
+  "refresh_token": "",
+  "id_token": ""
+}
```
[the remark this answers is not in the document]
fwkendall: Just adding to this, under /config there's a .env.example you can update to show what needs to be added for Cypress :) Just leave any sensitive values blank, just like you did here

cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
...
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
+
+// Alternatively you can use CommonJS syntax:
```
fwkendall, line 19: Could probably remove this (as well as the comment on line 16) if we're using the other option :)

cypress/support/utils.js
fwkendall: Fair point. And if we do want a UUID why not just use uuidv4 :) there's an npm package

cypress/support/utils.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    s[14] = "4";
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    s[19] = hexDigits.substr(s[19] | 0x8, 1);
+    s[8] = s[13] = s[18] = s[23] = "-";
+    var uuid = s.join("");
+    return uuid;
+}
+exports.createUUID = createUUID;
+function getAuthCodeFromLocation(location) {
```
fwkendall, line 17: This might also be something a URL lib could help simplify

fwkendall:
Overall this looks great, I'm super impressed with how easy it will be to write new tests with this framework in place 👍 Just some best practices to put into place to clean it up and we should be good to go!

cypress/support/utils.js
Zenen-freshworks: good point! swapping to UUIDv4

cypress/support/utils.js
Zenen-freshworks: this function was obselete so I removed it!

Zenen-freshworks:
I managed to get everything wrangled into the .env file with cypress-dotenv and I updated the example file accordingly.

.config/.env.example
```diff
@@ -1,22 +1,25 @@
 # Database Config
 POSTGRES_HOST=
 POSTGRES_PORT=
-POSTGRES_DB=db_dev
-POSTGRES_USER=admin
+POSTGRES_DB=
+POSTGRES_USER=
 POSTGRES_PASSWORD=
 
-# Logging database config
```
fwkendall, line 8: I'm going to be opening a PR soon with some README fixes, so I'll add these comments back in then :)

cypress/support/commands.js
```diff
@@ -0,0 +1,90 @@
+// These commands were sourced from the cypress-keycloak-login package
+const utils_1 = require('./utils');
```
ashtonmeuser, line 2: Any reason we're calling this `utils_1`? Could use destructing here to get only the `decodeToken` function by using `const {.decodeToken } = require('./utils');`.

fwkendall (approved):

mauriciofreshw (approved):

ashtonmeuser (approved):

PR #165, opened 2020-12-15
Hcap 222 engage with participants

mauriciofreshw:
A few adjustments have been done to avoid some react weirdness:

1 - Columns with buttons which the onClick apply changes to the current page should be rendered on the Table component to avoid a chain of dependencies with global methods like `openToast`.
~~2 - If your function relies on the value of a given state outside a render, that value should be accessed through a reference (useRef) and not the state directly, unless you are 100% sure about your state value.~~ (not necessary anymore due to the last push)

### DIFF
--- a/client/src/components/generic/Table.js
+++ b/client/src/components/generic/Table.js
@@ -47,15 +47,29 @@ const StyledTableRow = withStyles((theme) => ({
   },
 }))(TableRow);
 
-export const Table = ({ order, orderBy, onRequestSort, columns, rows, isLoading, rowsPerPage = 10 }) => {
+export const Table = ({ order, orderBy, renderObjectCell, onRequestSort, columns, rows, isLoading, rowsPerPage = 10 }) => {
 
   const [pageRows, setPageRows] = useState([]);
   const [page, setPage] = useState(0);
 
   useEffect(() => {
     const offset = page * rowsPerPage;
-    setPageRows(rows.slice(offset, offset + rowsPerPage))
-  }, [rows, page, rowsPerPage]);
+
+    const paginatedRows = rows.slice(offset, offset + rowsPerPage);
+
+    if (renderObjectCell) {
+      paginatedRows.forEach(row => {
+        Object.keys(row).forEach((key) => {
+          if (!React.isValidElement(row[key]) &&
+            typeof row[key] === 'object' && row[key] !== null) {
+            row[key] = renderObjectCell(row[key]);
+          }
+        });
+      });
+    }
+
+    setPageRows(paginatedRows);
+  }, [rows, page, rowsPerPage, renderObjectCell]);
 
   const createSortHandler = (property) => (event) => {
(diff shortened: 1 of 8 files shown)


server/server.js
```diff
@@ -71,6 +71,20 @@ app.get(`${apiBaseUrl}/participants`,
     return res.json({ data: result });
   }));
 
+// Engage participant
+app.post(`${apiBaseUrl}/engage-participant`,
```
fwkendall, line 75: I'm wondering if we actually want a generic endpoint here - like POST to /employer-actions since we're already handling the status in the request body? Then we would just need to expand this endpoint to handle more options vs. creating bespoke endpoints for each status

Thoughts?

server/server.js
```diff
@@ -71,6 +71,20 @@ app.get(`${apiBaseUrl}/participants`,
...
+  keycloak.allowRolesMiddleware('health_authority', 'employer'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    const user = req.hcapUserInfo;
+    const result = await setParticipantStatus(
+      user.id,
+      req.body.participantId,
+      req.body.disengage ? 'open' : 'prospecting',
```
fwkendall, line 83: we might want a different status here - like "archived" or something - so that in future we still have an easy way to separate out these participants into a separate tab

fwkendall (dismissed):

server/server.js
fwkendall: Looking at the designs again I think we want "rejected here" - it's okay if there's no real filtering happening on the FE for now for that status, if we have it tracked in the DB it will be useful in the next release anyway :)

client/src/components/generic/Table.js
```diff
@@ -47,15 +47,29 @@ const StyledTableRow = withStyles((theme) => ({
...
   const [page, setPage] = useState(0);
 
   useEffect(() => {
     const offset = page * rowsPerPage;
-    setPageRows(rows.slice(offset, offset + rowsPerPage))
-  }, [rows, page, rowsPerPage]);
+
+    const paginatedRows = rows.slice(offset, offset + rowsPerPage);
+
+    if (renderObjectCell) {
```
ashtonmeuser, line 60: This feels pretty weird. But we've got a deadline so if it works, so be it.

ashtonmeuser (approved):

server/server.js
mauriciofreshw: Totally makes sense, I just turn it into a generic actions endpoint for now

client/src/components/generic/Table.js
mauriciofreshw: Indeed, I found a better and more readable way to do that.

Zenen-freshworks (approved):
Looks good from my end!

fwkendall (approved):

ashtonmeuser (approved):

PR #167, opened 2020-12-15
Hcap 24X logging

Zenen-freshworks:
Logging finished for GET and POST from the participants API

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -69,16 +69,14 @@ app.get(`${apiBaseUrl}/participants`,
   asyncMiddleware(async (req, res) => {
     const user = req.hcapUserInfo;
     const result = await getParticipants(user);
-
-    // HCAP-242 - Log View Access Of Participant Data
     const { content } = req.kauth.grant.access_token;
     const logData = {
       "action": "participant_get",
-      "accessed_by": {
+      "performed_by": {
         "username": content.preferred_username,
         "id": content.sub
       },
-      //slicing so that things don't get out of hand with the IDs
+      // Slicing to one page of results
       "ids_viewed": result.map(person => person.id).slice(0,10),
       "timestamp": new Date()
     };
@@ -123,16 +121,14 @@ app.post(`${apiBaseUrl}/participants`,
 
     try {
       const response = await parseAndSaveParticipants(req.file.buffer);
-
-      // HCAP-244 - Log Participant Upload
       const { content } = req.kauth.grant.access_token;
       const logData = {
         "action": "participant_post",
-        "uploaded_by": {
+        "performed_by": {
           "username": content.preferred_username,
           "id": content.sub
         },
-        //slicing so that things don't get out of hand with the IDs
+        // Slicing to one page of results
         "ids_posted": response.map(entry => entry.id).slice(0,10),
         "timestamp": new Date()
       };
--- a/server/server.js
+++ b/server/server.js
@@ -70,7 +70,7 @@ app.get(`${apiBaseUrl}/participants`,
     const user = req.hcapUserInfo;
     const result = await getParticipants(user);
     const { content } = req.kauth.grant.access_token;
-    const logData = {
+    logger.info({
       "action": "participant_get",
       "performed_by": {
         "username": content.preferred_username,
@@ -78,9 +78,7 @@ app.get(`${apiBaseUrl}/participants`,
       },
       // Slicing to one page of results
       "ids_viewed": result.map(person => person.id).slice(0,10),
-      "timestamp": new Date()
-    };
-    logger.info(logData)
+    });
     return res.json({ data: result });
   }));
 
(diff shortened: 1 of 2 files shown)


server/server.js
```diff
@@ -68,6 +69,20 @@ app.get(`${apiBaseUrl}/participants`,
...
+
+    // HCAP-242 - Log View Access Of Participant Data
+    const { content } = req.kauth.grant.access_token;
+    const logData = {
+      "action": "participant_get",
+      "accessed_by": {
+        "username": content.preferred_username,
+        "id": content.sub
+      },
+      //slicing so that things don't get out of hand with the IDs
```
fwkendall, line 81: I would probably tweak this comment - the important thing to note here is that we picked 10 to log because those are the records that come back on the first page of results :)

server/server.js
```diff
@@ -68,6 +69,20 @@ app.get(`${apiBaseUrl}/participants`,
   asyncMiddleware(async (req, res) => {
     const user = req.hcapUserInfo;
     const result = await getParticipants(user);
+
+    // HCAP-242 - Log View Access Of Participant Data
```
fwkendall, line 73: These comments can be removed, once a feature is implemented we don't need to track the ticket in code :)

server/server.js
```diff
@@ -94,6 +109,21 @@ app.post(`${apiBaseUrl}/participants`,
 
     try {
       const response = await parseAndSaveParticipants(req.file.buffer);
+
+      // HCAP-244 - Log Participant Upload
+      const { content } = req.kauth.grant.access_token;
+      const logData = {
+        "action": "participant_post",
+        "uploaded_by": {
```
fwkendall, line 117: I think I would try to harmonize the format of these logs a little bit - like I think that instead of "uploaded_by" and "accessed_by" we can change both to "performed_by" since we have a separate key for the action already. Similar for "ids_viewed" and "ids_posted" - we could probably change both of these to "participant_ids"

fwkendall (dismissed):

server/server.js
```diff
@@ -68,6 +69,18 @@ app.get(`${apiBaseUrl}/participants`,
...
+    const { content } = req.kauth.grant.access_token;
+    const logData = {
+      "action": "participant_get",
+      "performed_by": {
+        "username": content.preferred_username,
+        "id": content.sub
+      },
+      // Slicing to one page of results
+      "ids_viewed": result.map(person => person.id).slice(0,10),
+      "timestamp": new Date()
```
mauriciofreshw, line 81: timestamp is not necessary since Winston already create it for each log message

server/server.js
```diff
@@ -94,6 +107,19 @@ app.post(`${apiBaseUrl}/participants`,
...
+        "action": "participant_post",
+        "performed_by": {
+          "username": content.preferred_username,
+          "id": content.sub
+        },
+        // Slicing to one page of results
+        "ids_posted": response.map(entry => entry.id).slice(0,10),
+        "timestamp": new Date()
+      };
+      logger.info(logData);
```
mauriciofreshw, line 121: Consider wrapping this into a function to avoid repeated code.

Zenen-freshworks: done, good call

fwkendall (approved):

server/server.js
```diff
@@ -68,6 +69,16 @@ app.get(`${apiBaseUrl}/participants`,
   asyncMiddleware(async (req, res) => {
     const user = req.hcapUserInfo;
     const result = await getParticipants(user);
+    const { content } = req.kauth.grant.access_token;
+    logger.info({
+      "action": "participant_get",
+      "performed_by": {
+        "username": content.preferred_username,
```
ashtonmeuser, line 76: This may be considered PII and should be kept out of logs. We don't even really need to use `content` because we are using the `keycloak.getUserInfoMiddleware` middleware with the results assigned to the `user` var on line 70. This could be reduced to `performed_by: user.id`.

server/server.js
```diff
@@ -108,6 +119,17 @@ app.post(`${apiBaseUrl}/participants`,
 
     try {
       const response = await parseAndSaveParticipants(req.file.buffer);
+      const { content } = req.kauth.grant.access_token;
+      logger.info({
+        "action": "participant_post",
+        "performed_by": {
+          "username": content.preferred_username,
```
ashtonmeuser, line 126: Same as above.

server/server.js
```diff
@@ -6,6 +6,7 @@ const dayjs = require('dayjs');
 const multer = require('multer');
 const { getParticipants, parseAndSaveParticipants, setParticipantStatus } = require('./services/participants.js');
 const { getEmployers, saveSites, getSites } = require('./services/employers.js');
+const { getUser } = require('./services/user');
```
ashtonmeuser, line 9: Can't see where we're using this. I'd remove the import.

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #168, opened 2020-12-15
Hcap 253 multiple regions fix

mauriciofreshw:
Depends on #165

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -153,14 +153,13 @@ app.post(`${apiBaseUrl}/approve-user`,
   keycloak.allowRolesMiddleware('ministry_of_health'),
   asyncMiddleware(async (req, res) => {
     await validate(AccessRequestApproval, req.body);
-    const promises = req.body.sites.map((siteId) => dbClient.db[collections.EMPLOYER_SITES].findDoc({ siteId }));
-    const results = await Promise.all(promises);
-    const siteRegions = results.map(item => item[0].healthAuthority);
-    const userRegions = siteRegions.includes(req.body.region) ?
-      siteRegions
-      :
-      [...siteRegions, req.body.region];
-    await keycloak.approvePendingRequest(req.body.userId, req.body.role, userRegions);
+    const results = await dbClient.db[collections.EMPLOYER_SITES].findDoc({
+      or: req.body.sites.map((siteId) => ({ 'siteId ilike': `%${siteId}%` })),
+    });
+    const siteRegions = results.map((item) => item.healthAuthority);
+    const joinedRegions = [...siteRegions, req.body.region];
+    const uniqueRegions = [...new Set(joinedRegions)];
+    await keycloak.approvePendingRequest(req.body.userId, req.body.role, uniqueRegions);
     await dbClient.db.saveDoc(collections.USERS, {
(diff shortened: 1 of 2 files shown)


Zenen-freshworks (approved):

server/server.js
```diff
@@ -153,7 +153,14 @@ app.post(`${apiBaseUrl}/approve-user`,
   keycloak.allowRolesMiddleware('ministry_of_health'),
   asyncMiddleware(async (req, res) => {
     await validate(AccessRequestApproval, req.body);
-    await keycloak.approvePendingRequest(req.body.userId, req.body.role, [req.body.region]);
+    const promises = req.body.sites.map((siteId) => dbClient.db[collections.EMPLOYER_SITES].findDoc({ siteId }));
```
ashtonmeuser, line 156: Is the only way to do this via multiple `findDoc` calls? There's no way to query by array?

fwkendall (dismissed):

server/server.js
```diff
@@ -153,7 +153,14 @@ app.post(`${apiBaseUrl}/approve-user`,
...
+    const results = await Promise.all(promises);
+    const siteRegions = results.map(item => item[0].healthAuthority);
```
ashtonmeuser, line 158: Could avoid using the misleading array index `[0]` by using the `single` option of massive `findDoc(query, { single: true }`.

server/server.js
```diff
@@ -153,7 +153,14 @@ app.post(`${apiBaseUrl}/approve-user`,
...
+    const userRegions = siteRegions.includes(req.body.region) ?
+      siteRegions
+      :
+      [...siteRegions, req.body.region];
```
ashtonmeuser, line 162: Okay I see what we're doing here. This is essentially a dedupe for `siteRegions` + `req.body.region`. Alternative here:
```
const joinedRegions = [...siteRegions, req.body.region];
const uniqueRegions = [...new Set(joinedRegions)];
```

ashtonmeuser (approved):

server/server.js
fwkendall: Yeah this is what I had to state at for a little bit. Adding everything then casting to set is how I would usually approach this as well, if we both find that more readable maybe it's worth swapping to it

fwkendall (approved):

server/server.js
```diff
@@ -153,7 +153,13 @@ app.post(`${apiBaseUrl}/approve-user`,
   keycloak.allowRolesMiddleware('ministry_of_health'),
   asyncMiddleware(async (req, res) => {
     await validate(AccessRequestApproval, req.body);
-    await keycloak.approvePendingRequest(req.body.userId, req.body.role, [req.body.region]);
+    const results = await dbClient.db[collections.EMPLOYER_SITES].findDoc({
+      or: req.body.sites.map((siteId) => ({ 'siteId ilike': `%${siteId}%` })),
```
ashtonmeuser, line 157: We're using `ilike` quite a bit throughout the app. Shouldn't this be a strict match on something like IDs.

ashtonmeuser (dismissed):

fwkendall (approved):

ashtonmeuser (approved):

PR #177, opened 2020-12-18
Logic for resetting the page has been implemented

Zenen-freshworks:
I haven't finished my ticket yet but I'm wondering about the logic on the useEffect at line 145 of Table.js. Does this constitute component coupling, and is this the appropriate way to go about solving the problem of resetting the page?


Zenen-freshworks:
This PR is no longer relevant

PR #209, opened 2021-01-06
Hcap 256 cypress in pipeline

Zenen-freshworks:
Cypress test suite can be activated by tagging cypress or on PR into master branch!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,6 @@ jobs:
           cat > .env << EOF
           ${{ secrets.ENVFILE }}
           EOF
-          cat .env
 
       - name: Build Locally
         run: make local-build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,44 @@
 name: Test application (e2e)
 
 on:
+  push:
+    tags:
+      - cypress
+
   pull_request:
     branches:
       - master
 
+
 jobs:
+  cypress-test:
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Checkout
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -1,11 +1,44 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Make env
+        run: |
+          cat > .env << EOF
+          ${{ secrets.ENVFILE }}
+          EOF
+          cat .env
```
ashtonmeuser, line 25: Seems like a red flag to `cat` this file.

ashtonmeuser (changes requested):

.github/workflows/test.yml
```diff
@@ -1,11 +1,43 @@
...
+
+      - name: Build Locally
+        run: make local-build
+
+      - name: Run locally
+        run: make local-run &
+
+      - name: Waiting for server to spin up
+        run: |
+          while ! nc -z localhost 4000; do
```
ashtonmeuser, line 34: I hope this doesn't put GH Actions in a state where it runs forever and eats into our compute quota. I guess `make local-run` would have to fail for that to be the case which would kill the Action.

ashtonmeuser: On second thought, Docker Compose doesn't fail if a service fails, so I think this is a possibility.

ashtonmeuser (approved):

.github/workflows/test.yml
fwkendall: Logging another ticket to look into this :)

fwkendall (approved):

PR #212, opened 2021-01-09
Added json data field validation

mauriciofreshw:


server/validation.js
```diff
@@ -250,7 +256,40 @@ const AccessRequestApproval = yup.object().noUnknown('Unknown field in form').sh
 
 const ParticipantStatusChange = yup.object().noUnknown('Unknown field in form').shape({
   participantId: yup.number().required('Participant ID is required'),
-  data: yup.object(),
+  data: yup.object().when(['status'], (status, schema) => {
+    if (status === 'interviewing') {
+      return schema.noUnknown('Unknown field in form').shape({
+        contacted_at: yup.string().required('Date of contact is required').test('is-date', 'Not a valid date', validateDateString),
```
mauriciofreshw, line 262: Looks like for **interviewing** and **final_status** we are not using camelCase for the data object, and also for this line the field name is different from the frontend (contactedDate). Not sure what we can do about that or just leave it as it is since is a minor thing.

Zenen-freshworks (approved):

server/validation.js
ashtonmeuser: Yikes. I suppose this was bound to happen again. Agreed that we have no easy way of changing this as we've released to prod with the current keys.

ashtonmeuser (approved):

Zenen-freshworks (approved):

PR #225, opened 2021-01-18
Hcap 240 moh request access

Zenen-freshworks:


client/src/pages/private/UserView.js
```diff
@@ -220,6 +223,39 @@ export default () => {
                   }}
                 />
               </Box>}
+            {(values.role === 'employer') && <Box mt={3}>
```
fwkendall, line 226: Isn't this condition already handled on line 204? I'm confused why we check for employer twice

client/src/pages/private/ParticipantView.js
```diff
@@ -335,7 +335,8 @@ export default () => {
         setLoadingUser(false);
         setSites(sites);
         setRoles(roles);
-        
+        // a Boolean filter removes all undefined values
+        setLocations(roles.map((loc) => locationRoles[loc]).filter(Boolean));
```
fwkendall, line 339: Curious why this was added? I thought the tests might need it but that doesn't seem to be the case :)

fwkendall:
Looks good! (And tests look great) - just a couple of questions from me :)

client/src/pages/private/UserView.js
Zenen-freshworks: Good catch! This was some redundant code that I was going to use as a template and then forgot to remove it.

client/src/pages/private/ParticipantView.js
Zenen-freshworks: More redundant code! I implemented a better version of this setter a few lines down.

fwkendall (approved):

mauriciofreshw (approved):

PR #229, opened 2021-01-20
added missing edge case

mauriciofreshw:


fwkendall (approved):

ashtonmeuser (approved):

ashtonmeuser:
We're really getting into some dense biz logic here. Not sure how to avoid that, but seems like it will reach a fever pitch at some point.

PR #246, opened 2021-01-29
Hcap 323 update participant

Zenen-freshworks:
Big PR... aside from adding the EditParticipant form I also slimmed down some of the tests and accidentally unindented a couple of HTML blocks - gonna have to find some way to improve that moving forwards.

### DIFF
--- a/client/src/components/modal-forms/EditParticipantForm.js
+++ b/client/src/components/modal-forms/EditParticipantForm.js
@@ -1,0 +1,133 @@
+import React, { useEffect, useState } from 'react';
+import Grid from '@material-ui/core/Grid';
+import { Button } from '../generic';
+import { Box } from '@material-ui/core';
+import { RenderDateField, RenderCheckbox, RenderTextField, RenderSelectField } from '../fields';
+import { Field, Formik, Form as FormikForm } from 'formik';
+import store from 'store';
+
+export const EditParticipantForm = ({ initialValues, validationSchema, onSubmit, onClose, sites }) => {
+
+  const [sitesDetail, setSitesDetail] = useState([]);
+
+  useEffect(() => {
+    const fetchSites = async () => {
+      const response = await fetch('/api/v1/employer-sites', {
+        headers: { 'Authorization': `Bearer ${store.get('TOKEN')}` },
+        method: 'GET',
+      });
+
+      if (response.ok) {
+        const { data } = await response.json();
+        setSitesDetail(data.filter((site) => sites.includes(site.siteId)));
+      }
+    };
+
+    fetchSites();
+  }, [sites]);
+
+  const getTodayDate = () => {
+    const today = new Date();
+    const dd = String(today.getDate()).padStart(2, '0');
+    const mm = String(today.getMonth() + 1).padStart(2, '0');
+    const yyyy = today.getFullYear();
+    return `${yyyy}-${mm}-${dd}`;
+  }
+
+  return <Formik
+    initialValues={initialValues}
+    validationSchema={validationSchema}
+    onSubmit={onSubmit}
+  >
+    {({ submitForm, values }) => (
+      <FormikForm>
+        <Box>
+      <Field
+      name="firstName"
+      component={RenderTextField}
+      label="First Name"
+      />
+      <Field
+      name="lastName"
+      component={RenderTextField}
+      label="Last Name"
+      />
+      <Field
+      name="phoneNumber"
+      component={RenderTextField}
+      label="* Phone Number"
+      type="tel"
+      />
+      <Field
+      name="emailAddress"
+      component={RenderTextField}
+      label="* Email Address"
+      type="email"
+      />
+          <Field
+            name="nonHcapOpportunity"
+            component={RenderCheckbox}
+            label="Non-HCAP Opportunity"
+          />
+          {
+            values.nonHcapOpportunity && (<>
+              <Field
+                name="positionType"
+                component={RenderSelectField}
+                label="* Position Type"
+                options={[
+                  { value: 'Full-Time', label: 'Full-Time' },
+                  { value: 'Part-Time', label: 'Part-Time' },
+                  { value: 'Casual', label: 'Casual' },
+                ]}
+              />
+            </>)
+          }
+          <Field
+            name="hiredDate"
+            component={RenderDateField}
+            maxDate={getTodayDate()}
+            label="* Date Hired"
+          />
+          <Field
+            name="startDate"
+            component={RenderDateField}
+            label="* Start Date"
+          />
+          <Field
+            name="site"
+            component={RenderSelectField}
+            label="* Site"
+            options={sitesDetail.map((siteDetail) => ({
+              value: siteDetail.siteId, label: siteDetail.siteName,
+            }))}
+          />
+          <Field
+            name="acknowledge"
+            component={RenderCheckbox}
+            label="I acknowledge that the participant has accepted the offer in writing."
+          />
+        </Box>
+        <Box mt={3}>
+          <Grid container spacing={2} justify="flex-end">
+            <Grid item>
+              <Button
+                onClick={onClose}
+                color="default"
+                text="Cancel"
+              />
+            </Grid>
+            <Grid item>
+              <Button
+                onClick={submitForm}
+                variant="contained"
+                color="primary"
+                text="Submit"
+              />
+            </Grid>
+          </Grid>
+        </Box>
+      </FormikForm>
+    )}
+  </Formik>;
+};
--- a/client/src/components/modal-forms/EditParticipantForm.js
+++ b/client/src/components/modal-forms/EditParticipantForm.js
@@ -1,20 +1,11 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import Grid from '@material-ui/core/Grid';
 import { Button } from '../generic';
 import { Box } from '@material-ui/core';
-import { RenderDateField, RenderCheckbox, RenderTextField, RenderSelectField } from '../fields';
+import { RenderTextField, RenderSelectField } from '../fields';
 import { Field, Formik, Form as FormikForm } from 'formik';
-import store from 'store';
 
 export const EditParticipantForm = ({ initialValues, validationSchema, onSubmit, onClose }) => {
-  const getTodayDate = () => {
-    const today = new Date();
-    const dd = String(today.getDate()).padStart(2, '0');
-    const mm = String(today.getMonth() + 1).padStart(2, '0');
-    const yyyy = today.getFullYear();
-    return `${yyyy}-${mm}-${dd}`;
-  }
-
   return <Formik
     initialValues={initialValues}
     validationSchema={validationSchema}
@@ -69,7 +60,7 @@ export const EditParticipantForm = ({ initialValues, validationSchema, onSubmit,
                 onClick={submitForm}
                 variant="contained"
                 color="primary"
-                text="Submit"
+                text="Save"
               />
             </Grid>
           </Grid>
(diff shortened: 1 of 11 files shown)


client/src/constants/validation.js
```diff
@@ -213,6 +213,14 @@ export const HiredParticipantSchema = yup.object().noUnknown('Unknown field in f
   site: yup.number().required('Site is required'),
 });
 
+export const EditParticipantFormSchema = yup.object().shape({
+  firstName: yup.string().required(errorMessage),
+  lastName: yup.string().required(errorMessage),
+  phoneNumber: yup.string().required(errorMessage).matches(/^\d{10}$/, 'Phone number must be provided as 10 digits'),
+  emailAddress: yup.string().required(errorMessage).matches(/^(.+@.+\..+)?$/, 'Invalid email address'),
+  //postalCode: yup.string().required(errorMessage).matches(/^[A-Z]\d[A-Z]\s?\d[A-Z]\d$/, 'Format as A1A 1A1'),
```
fwkendall, line 221: Commented code can be removed :)

Do we want validation on the interest dropdown pick as well?

client/src/pages/private/ParticipantView.js
```diff
@@ -745,7 +818,7 @@ export default () => {
             />
           </Box>
         </Grid>
-        <Menu
+    {(!roles.includes('ministry_of_health')) && <Menu
```
fwkendall, line 821: Does this prevent superusers from accessing this feature, or are their permissions being handled earlier?

cypress/support/commands.js
```diff
@@ -176,30 +90,18 @@ Cypress.Commands.add("kcNavAs", function (user, visitUrl) {
...
-      "operatorContactFirstName":"Heather",
-      "id":2,
-      "created_at":"2021-01-26T17:22:43.804Z",
-      "updated_at":null
-    }
   ];
 
+  const participant = [{
+    "id":1,
+    "emailAddress":"graeme.clarke@gov.bc.ca",
```
fwkendall, line 97: @Zenen-freshworks make sure we're using fake user data in our tests :)

server/server.js
```diff
@@ -114,6 +119,45 @@ app.get(`${apiBaseUrl}/participants`,
...
+    });
+    return res.json(result);
+  }));
+
+// Update participant data
+app.post(`${apiBaseUrl}/participant`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health', 'employer'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    // await validate(ParticipantQuerySchema, req.body);
```
fwkendall, line 148: should be cleaned up :)

client/src/constants/validation.js
Zenen-freshworks: I'm not sure. For the users that don't have an interest value, do we want selecting an option to be mandatory?

server/server.js
Zenen-freshworks: Oh! Good reminded that I never implemented validation. I'll get on that.

client/src/pages/private/ParticipantView.js
Zenen-freshworks: This was implemented to avoid a bug that occurred with anchor elements - I'll extend it to superusers.

client/src/constants/validation.js
fwkendall: Good point, I don't think we want it mandatory in that case. Can we make it nullable and ensure it's an expected value if entered though?

server/server.js
```diff
@@ -114,6 +120,45 @@ app.get(`${apiBaseUrl}/participants`,
...
+      },
+      on: {
+        id,
+      },
+    });
+    return res.json(result);
+  }));
+
+// Update participant data
+app.post(`${apiBaseUrl}/participant`,
```
fwkendall, line 145: IMO this should be a PUT or PATCH and not a POST operation

server/server.js
```diff
@@ -114,6 +120,45 @@ app.get(`${apiBaseUrl}/participants`,
...
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health', 'employer'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    await validate(ParticipantEditSchema, req.body);
+    const user = req.hcapUserInfo;
+    const result = await updateParticipant(req.body);
+    logger.info({
+      action: 'participant_get',
```
fwkendall, line 153: Should be participant_put or participant_patch

server/server.js
```diff
@@ -114,6 +120,45 @@ app.get(`${apiBaseUrl}/participants`,
...
+      performed_by: {
+        username: user.username,
+        id: user.id,
+      },
```
fwkendall, line 157: need to add participant_id so we know which record was edited (keep consistent key with other participant endpoints)

client/src/components/modal-forms/EditParticipantForm.js
```diff
@@ -0,0 +1,71 @@
...
+export const EditParticipantForm = ({ initialValues, validationSchema, onSubmit, onClose }) => {
+  return <Formik
+    initialValues={initialValues}
+    validationSchema={validationSchema}
+    onSubmit={onSubmit}
+  >
+    {({ submitForm, values }) => (
+      <FormikForm>
+        <Box>
+      <Field
```
ashtonmeuser, line 17: Indenting seems a little weird from here on.

client/src/pages/private/ParticipantView.js
```diff
@@ -524,95 +534,134 @@ export default () => {
...
             });
-          }}
-          onClose={defaultOnClose}
-        />}
-      </Dialog>
+          }
+        });
+        values.history = (actionMenuParticipant.history)? [history, ...actionMenuParticipant.history] : [history];
+        const response = await fetch('/api/v1/participant', {
+          method: 'POST',
```
fwkendall, line 618: This will also need to change if we swap to put/patch - just flagging to make it easy to find

client/src/pages/private/ParticipantView.js
```diff
@@ -524,95 +534,134 @@ export default () => {
...
-        />}
-
-        {activeModalForm === 'interviewing' && <InterviewingForm
-          initialValues={{ contactedDate: '' }}
-          validationSchema={InterviewingFormSchema}
-          onSubmit={(values) => {
-            handleEngage(actionMenuParticipant.id, 'interviewing', { contacted_at: values.contactedDate });
-          }}
-          onClose={defaultOnClose}
+    <Dialog
```
ashtonmeuser, line 537: Also strange indenting here.

server/server.js
```diff
@@ -114,6 +120,45 @@ app.get(`${apiBaseUrl}/participants`,
...
+      on: {
+        id,
+      },
+    });
+    return res.json(result);
+  }));
+
+// Update participant data
+app.post(`${apiBaseUrl}/participant`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health', 'employer'),
```
fwkendall, line 146: should be MoH only

client/src/pages/private/ParticipantView.js
```diff
@@ -584,6 +594,45 @@ export default () => {
...
+            Object.keys(values).forEach(key => {
+              if (values[key] !== actionMenuParticipant[key]) {
+                history.changes.push({
+                  field: key,
+                  from: actionMenuParticipant[key],
+                  to: values[key],
+                });
+              }
+            });
+            values.history = (actionMenuParticipant.history)? [history, ...actionMenuParticipant.history] : [history];
```
ashtonmeuser, line 616: Should these be in chronological order?
```suggestion
            values.history = (actionMenuParticipant.history)? [...actionMenuParticipant.history, history] : [history];
```

cypress/support/commands.js
```diff
@@ -213,11 +115,12 @@ Cypress.Commands.add("kcNavAs", function (user, visitUrl) {
...
-          "id":"17b2254c-269e-4c21-b19d-0c005365a5e9",
-          "emailAddress":"graeme.clarke@gov.bc.ca",
-          "username":"gclarke@idir",
-          "firstName":"Graeme",
-          "lastName":"Clarke",
+          "id":1,
+          "emailAddress":"tasty@snac.ks",
+          "username":"graham",
+          "firstName":"Graham",
+          "lastName":"Crackers",
```
ashtonmeuser, line 122: I love this change 😄

server/server.js
```diff
@@ -114,6 +120,47 @@ app.get(`${apiBaseUrl}/participants`,
...
+    return res.json(result);
+  }));
+
+// Update participant data
+app.patch(`${apiBaseUrl}/participant`,
+  keycloak.allowRolesMiddleware('ministry_of_health'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    console.log('req.body');
+    console.log(req.body);
```
ashtonmeuser: Looks like some leftover debugging.
```suggestion
```

server/server.js
```diff
@@ -114,6 +120,47 @@ app.get(`${apiBaseUrl}/participants`,
...
+    await validate(ParticipantEditSchema, req.body);
+    const user = req.hcapUserInfo;
+    const result = await updateParticipant(req.body);
+    logger.info({
+      action: 'participant_get',
```
ashtonmeuser, line 153: Think this should reflect the correct action being taken.
```suggestion
      action: 'participant_updated',
```

server/services/participants.js
```diff
@@ -71,6 +71,32 @@ const setParticipantStatus = async (
...
 });
 
+const getParticipantByID = async (participantInfo) => dbClient.db.withTransaction(async (tx) => {
+  const participant = await tx[collections.PARTICIPANTS].findDoc({
+    id: participantInfo.id,
+  });
+  return participant;
+});
+
+const updateParticipant = async (participantInfo) => dbClient.db.withTransaction(async (tx) => {
```
ashtonmeuser, line 81: This totally seems like it will accomplish what we want but I'm concerned about the safety of taking a user-submitted change object and modifying the record. I wonder if we should filter out props that we definitely don't want changed (`id`, `maximusId`, etc.)?

ashtonmeuser: Just saw that this is actually dependent on the validation which only lets certain properties through. Therefore, at the time of writing, only first name, last name, phone number, and email address are editable.

server/validation.js
```diff
@@ -337,6 +337,13 @@ const ParticipantQuerySchema = yup.object().shape({
   ),
 });
 
+const ParticipantEditSchema = yup.object().shape({
```
ashtonmeuser, line 340: Without specifying no additional fields, a malicious actor could slip in another field to be edited. See my comments above. Adding `noUnknown` should fix this.
```suggestion
const ParticipantEditSchema = yup.object().noUnknown('Unknown field in form').shape({
```

server/services/participants.js
```diff
@@ -71,6 +71,32 @@ const setParticipantStatus = async (
   return { status };
 });
 
+const getParticipantByID = async (participantInfo) => dbClient.db.withTransaction(async (tx) => {
+  const participant = await tx[collections.PARTICIPANTS].findDoc({
+    id: participantInfo.id,
+  });
+  return participant;
+});
```
mauriciofreshw, line 79: I think there’s no reason to use a transaction for a simple find operation only, you can just jump to the findDoc directly.

ashtonmeuser (approved):

mauriciofreshw (dismissed):

ashtonmeuser (approved):

mauriciofreshw (approved):

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #247, opened 2021-01-30
Show participants that do not have CRC clear, show callback status

ashtonmeuser:
Some context: Tried to get a Jest `expect` extension going for comparing raw participant data (seed data in tests) to data returned from the API to employers. Could not get it to work and development is taking far too long as Docker is not updating container FS for me at the moment.

### DIFF
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -141,6 +141,16 @@ const getParticipants = async (user, pagination, sortField,
     };
   }
 
+  const callbackStatus = (participant) => {
+    const d = new Date(participant.updated_at);
+    if (d instanceof Date && !Number.isNaN(d.getTime())) {
+      if (d < new Date('2021-01-01T00:00:00-08:00')) {
+        return 'Primed';
+      }
+    }
+    return 'Available';
+  };
+
   // Returned participants for employers
   return {
     data: participants.map((item) => {
@@ -151,6 +161,7 @@ const getParticipants = async (user, pagination, sortField,
         postalCodeFsa: item.postalCodeFsa,
         preferredLocation: item.preferredLocation,
         nonHCAP: item.nonHCAP,
+        callbackStatus: callbackStatus(item),
       };
 
       const hiredBySomeoneElseStatus = item.statusInfos?.find((statusInfo) => statusInfo.status === 'hired'
--- a/server/tests/participant.form.test.js
+++ b/server/tests/participant.form.test.js
@@ -12,6 +12,7 @@ const {
 } = require('../services/participants.js');
 const { getReport } = require('../services/reporting.js');
 const { evaluateBooleanAnswer } = require('../validation');
+// require('./util/db');
 
 describe('Participants Service', () => {
   beforeAll(async () => {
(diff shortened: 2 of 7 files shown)


server/services/participants.js
```diff
@@ -141,6 +141,16 @@ const getParticipants = async (user, pagination, sortField,
     };
   }
 
+  const callbackStatus = (participant) => {
+    const d = new Date(participant.updated_at);
+    if (d instanceof Date && !Number.isNaN(d.getTime())) {
```
mauriciofreshw, line 146: Doesn’t will `d` be always an instance of Date?

ashtonmeuser: It should be, yes.

server/tests/participant.form.test.js
```diff
@@ -12,6 +12,7 @@ const {
 } = require('../services/participants.js');
 const { getReport } = require('../services/reporting.js');
 const { evaluateBooleanAnswer } = require('../validation');
+// require('./util/db');
```
fwkendall, line 15: sneaky :)

server/tests/participant.form.test.js
```diff
@@ -264,21 +265,33 @@ describe('Participants Service', () => {
...
+      ];
+      return a.map((i) => (
+        Object.keys(i)
+          .filter((k) => employerColumns.includes(k))
+          .reduce((o, k) => ({ ...o, [k]: i[k] }), { nonHCAP: undefined })
+      ));
+    };
+    const trimIds = (a) => a.map((i) => (
+      Object.keys(i)
+        .filter((k) => !['id', 'callbackStatus'].includes(k)) // TODO: Should not ignore callback status
```
fwkendall, line 286: Is this TODO still relevant?

server/tests/participant.form.test.js
```diff
@@ -264,21 +265,33 @@ describe('Participants Service', () => {
...
+        .reduce((o, k) => ({ ...o, [k]: i[k] }), { nonHCAP: undefined })
+    ));
+
+    const expected = mapRawToEmployerColumns(allParticipants
+      .filter((i) => (evaluateBooleanAnswer(i.interested))));
+    expect(trimIds(res.data)).toEqual(expect.arrayContaining(expected));
+    // const expected = allParticipants.filter((i) => (evaluateBooleanAnswer(i.interested)));
```
fwkendall, line 293: Some more draft leftovers here :)

server/tests/participant.form.test.js
ashtonmeuser: Still a TODO. Not sure how we should solve this besides replicating the exact algo used on server. Currently this is:
```
const callbackStatus = (participant) => {
  const d = new Date(participant.updated_at);
  if (d instanceof Date && !Number.isNaN(d.getTime())) {
    if (d < new Date('2021-01-01T00:00:00-08:00')) {
      return 'Primed';
    }
  }
  return 'Available';
};
```
Think we're good to keep ignoring this for the time being. The properties that are being compared in this test are way stricter than those in other tests which are just looking at keys. This is what I was looking to address with the `expect` extension, but did not have the time to finish.

fwkendall: Could you log a ticket in the backlog for this? A spike or a task, doesn't really matter - just don't want to lose line of sight on this

mauriciofreshw (approved):

fwkendall (approved):

mauriciofreshw (approved):

PR #259, opened 2021-02-03
Hcap 340 freshness

Zenen-freshworks:
Sorting doesn't work properly yet.

### DIFF
--- a/client/src/pages/private/ParticipantView.js
+++ b/client/src/pages/private/ParticipantView.js
@@ -30,7 +30,7 @@ const defaultColumns = [
   { id: 'postalCodeFsa', name: 'FSA' },
   { id: 'preferredLocation', name: 'Preferred Region(s)' },
   { id: 'nonHCAP', name: 'Non-HCAP' },
-  { id: 'freshness', name: 'Last Updated' },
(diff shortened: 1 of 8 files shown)


client/src/pages/private/ParticipantView.js
```diff
@@ -131,6 +134,8 @@ export default () => {
   const [locations, setLocations] = useState([]);
 
   const handleRequestSort = (event, property) => {
+    if (property === 'freshness') property = 'updated_at';
```
mauriciofreshw, line 137: Try to avoid modifying function parameters. I know this is a primitive type and doesn’t mutate the original string but in case of an object this could lead to a code smell.

mauriciofreshw (approved):

client/src/pages/private/ParticipantView.js
Zenen-freshworks: yea, this bit was kind of slapped on as a band-aid - I'll fix it up.

mauriciofreshw (approved):

fwkendall (approved):

ashtonmeuser (approved):

PR #261, opened 2021-02-03
Update role/sites feature

mauriciofreshw:


client/src/constants/label.js
```diff
@@ -0,0 +1,8 @@
+export const regionLabelsMap = {
+  region_interior: 'Interior',
+  region_fraser: 'Fraser',
+  region_vancouver_coastal: 'Vancouver Coastal',
+  region_vancouver_island: 'Vancouver Island',
+  region_northern: 'Northern',
+  none: 'None'
+};
```
Zenen-freshworks, line 8: factoring stuff out of participantsview, nice!

cypress/integration/participantView.spec.js
```diff
@@ -37,15 +37,15 @@ describe("Participant View", () => {
...
+    cy.get('ul.MuiMenu-list[role=listbox] > li').should('have.length', 7);
   });
 
   it("Visits Participant View as a MoH user", () => {
     cy.kcNavAs('ministry_of_health', 'participant-view');
     cy.get('div.MuiSelect-selectMenu').should('not.have.class', 'Mui-disabled');
     cy.get('div.MuiSelect-selectMenu').click();
     cy.get('ul.MuiMenu-list[role=listbox]').should('be.visible');
-    cy.get('ul.MuiMenu-list[role=listbox] > li').should('have.length', 6);
+    cy.get('ul.MuiMenu-list[role=listbox] > li').should('have.length', 7);
```
Zenen-freshworks, line 48: Question: if we have None as an option, should "Preferred Location" still be available as a catch-all? Do they serve different purposes?

Zenen-freshworks (changes requested):
I'm getting an undefined error on filtering by None in ParticipantsView.

Zenen-freshworks (approved):

Zenen-freshworks (dismissed):

Zenen-freshworks (approved):

client/src/pages/private/EOIView.js
```diff
@@ -56,19 +56,10 @@ export default () => {
...
-      region_vancouver_island: 'Vancouver Island',
-      region_northern: 'Northern',
-      none: 'None'
-    };
-
     setHealthAuthorities(
       (roles.includes("superuser") || roles.includes("ministry_of_health"))
-      ? Object.values(locationRoles) 
-      : roles.map((loc) => locationRoles[loc]).filter(Boolean)
+        ? Object.values(regionLabelsMap).filter(value => value !== 'None')
```
fwkendall, line 61: I think this is incorrect, if I'm reading it right - MoH and superuser should have access to None for EEOIs

fwkendall: Note: does not apply to participants, None is only needed for EEOIs

server/server.js
```diff
@@ -355,12 +361,43 @@ app.get(`${apiBaseUrl}/employer-sites-detail`,
     return res.json({ data: '' });
   }));
 
+app.post(`${apiBaseUrl}/update-user`,
```
fwkendall, line 364: should be patch :)

server/server.js
```diff
@@ -355,12 +361,43 @@ app.get(`${apiBaseUrl}/employer-sites-detail`,
...
+        keycloakId: req.body.userId,
+      },
+      {
+        sites: req.body.sites,
+      },
+    );
+
+    const user = req.hcapUserInfo;
+    logger.info({
+      action: 'update-user_post',
```
fwkendall, line 381: just commenting so we remember to update this according to the endpoint update (user_patch)

server/server.js
```diff
@@ -355,12 +361,43 @@ app.get(`${apiBaseUrl}/employer-sites-detail`,
...
+      performed_by: {
+        username: user.username,
+        id: user.id,
+      },
+      role_assigned: req.body.role,
+      granted_access_to: req.body.userId,
```
fwkendall, line 387: this key should probably have a different name for this type of action - user_id would be fine IMO (honestly I think it should have been like this for the other endpoint too, but that ship has sailed)

fwkendall: I guess we might want to leave it as-is, even though I think it's a bad key, so that it's consistent

ashtonmeuser (approved):

Zenen-freshworks (approved):

PR #263, opened 2021-02-05
Views and tables have been decoupled

Zenen-freshworks:
Prominent Changes:
- Most of the logic in participantsView has been moved to participantsTable, leaving the view as a wrapper for the table
- siteView is a wrapper for the new siteTable, accessible via the admin page
- Some smaller functions have been factored out of participantsTable into utils

Current Issues:
- ~~overall bugginess with how the tables render~~
- ~~sitesView shows all sites by default, regardless of relevance to given health authorities (sometimes?)~~
- ~~"My sites" tab does not exclusively show sites assigned to the health authority~~

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -1,0 +1,799 @@
+import React, { useEffect, useState } from 'react';
+import Grid from '@material-ui/core/Grid';
+import { withStyles } from '@material-ui/core/styles';
+import Tabs from '@material-ui/core/Tabs';
+import Tab from '@material-ui/core/Tab';
+import { Box, Typography, TextField, Menu, MenuItem } from '@material-ui/core';
+import store from 'store';
+import {
+  ToastStatus,
+  InterviewingFormSchema,
+  RejectedFormSchema,
+  HireFormSchema,
+  HiredParticipantSchema,
+  EditParticipantFormSchema,
+  regionLabelsMap,
+} from '../../constants';
+import { Table, CheckPermissions, Button, Dialog } from '../../components/generic';
+import { ProspectingForm, InterviewingForm, RejectedForm, HireForm, NewParticipantForm, EditParticipantForm } from '../../components/modal-forms';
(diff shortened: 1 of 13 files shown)


client/src/pages/private/ParticipantTable.js
```diff
@@ -0,0 +1,799 @@
...
+          {actionMenuParticipant?.status === 'interviewing' && <MenuItem onClick={() => handleEngage(actionMenuParticipant.id, 'offer_made')}>Offer Made</MenuItem>}
+          {actionMenuParticipant?.status === 'offer_made' && <MenuItem onClick={() => setActiveModalForm('hired')}>Hire</MenuItem>}
+          {['prospecting', 'interviewing', 'offer_made'].includes(actionMenuParticipant?.status) && <MenuItem onClick={() => setActiveModalForm('rejected')}>Archive</MenuItem>}
+          {actionMenuParticipant?.status === 'rejected' && <MenuItem onClick={() => handleEngage(actionMenuParticipant.id, 'prospecting')}>Re-engage</MenuItem>}
+        </Menu>
+        }
+      </CheckPermissions>
+    </>
+  );
+};
```
ashtonmeuser, line 799: So the reasoning here is to relocate the majority of the participant view logic to this file? We've gone from one 850LOC file to 800LOC and 50LOC files. Probably a very small gain, but a start I suppose. This may be further addressed by HCAP-351.

ashtonmeuser (approved):

client/src/pages/private/ParticipantTable.js
ashtonmeuser: I think I see what you were doing. This is following a pattern of keeping tables separate from the views they're included in. Also done with sites view.

Zenen-freshworks: Yes, since the sites table needs to be accessed in more than one way, I figured it best to encapsulate it. I did the same with the participants table for the sake of consistency.

mauriciofreshw (approved):

PR #264, opened 2021-02-06
Hcap 338 flag callback

mauriciofreshw:
As per the ticket description, once provided with a list of IDs for these participants, we’ll write a script/migration to flag them as having gone through this process. In the meanwhile, I refactor the current code to use a jsonb field as the callbackStatus. This way we are able to sort by callbackStatus.


server/migrations/1612489825996_create-callback-field.js
```diff
@@ -0,0 +1,36 @@
...
+
+// const ids = [];
+
+// xlsx[0].data.slice(1, xlsx[0].data.length - 1).forEach((row) => {
+//   if (row[6] === 'Yes') {
+//     ids.push(row[0]);
+//   }
+// });
+
+// console.log(ids.join(', '));
```
Zenen-freshworks, line 24: Can this code be removed?

Zenen-freshworks (approved):

server/migrations/1612489825996_create-callback-field.js
mauriciofreshw: Could be but I decided to let it in case we need to extract new Ids from a new spreadsheet again

Zenen-freshworks (approved):

server/migrations/1612489825996_create-callback-field.js
ashtonmeuser: Just a nitpick: I'm inclined to get rid of this. Would be easy enough to filter in Excel or something and extract the IDs.

ashtonmeuser (approved):

ashtonmeuser (approved):

Zenen-freshworks (dismissed):

ashtonmeuser (approved):

Zenen-freshworks (approved):

PR #265, opened 2021-02-09
Hcap 165 sites view

Zenen-freshworks:
There's a strange bug regarding tab underlining in the ParticipantsView, but it seems to be pretty minor and I think I need to put some work into editing details before the sprint is over


mauriciofreshw (approved):

ashtonmeuser:
It's captured in the ticket, but it's a little odd having two ways of accessing sites (and two UIs) is a little odd. This only applies to HAs. Is it intended for sites to eventually be editable via My Sites but not via View Sites?

ashtonmeuser (approved):

PR #270, opened 2021-02-10
Hcap 176 site details

Zenen-freshworks:
WIP
Accomplishments:
- new page/route for site details
- API is now more RESTful

Issues:
~~- site details displays straight-up JSON~~
~~- EOI details don't fill out according to the passed object~~
~~- the site's ID is used for indexing rather than its siteID~~


client/src/pages/private/EOIViewDetails.js
```diff
@@ -1,19 +1,17 @@
...
 import Grid from '@material-ui/core/Grid';
-import { Routes } from '../../constants';
-import { useLocation, Redirect } from 'react-router-dom';
 import { Page, CheckPermissions } from '../../components/generic';
 import { Form } from '../../components/employer-form';
 import { scrollUp } from '../../utils';
 import store from 'store';
 
-export default () => {
+export default ({ match }) => {
```
mauriciofreshw, line 8: Since you are using only `match.params.id` would be better to just pass the id itself as a prop

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+import Grid from '@material-ui/core/Grid';
+import { Page, CheckPermissions } from '../../components/generic';
+import { Box, Typography } from '@material-ui/core';
+import { scrollUp } from '../../utils';
+import store from 'store';
+import { Link } from 'react-router-dom';
+import routes from '../../constants/routes';
+
+
+export default (props) => {
```
mauriciofreshw, line 11: Same here, only `props` seems too broad.

client/src/pages/private/EOIViewDetails.js
Zenen-freshworks: This is the syntax that's provided through Routes. I don't know if/how I can pass id individually.

mauriciofreshw: Ah, I see. You are not responsible for rendering the component. Nvm!

mauriciofreshw (dismissed):

client/src/pages/private/SiteViewDetails.js
Zenen-freshworks: I fixed this one up.

mauriciofreshw (approved):

server/server.js
```diff
@@ -75,6 +77,14 @@ app.get(`${apiBaseUrl}/employer-form`,
     return res.json({ data: result });
   }));
 
+app.get(`${apiBaseUrl}/employer-form/:id`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
+  keycloak.getUserInfoMiddleware(),
```
fwkendall, line 82: IMO we should still check the region here in case of HA - otherwise it could be possible for a user to navigate to a URL directly and see something they shouldn't have access to

server/server.js
```diff
@@ -75,6 +77,14 @@ app.get(`${apiBaseUrl}/employer-form`,
...
+  asyncMiddleware(async (req, res) => {
+    const [result] = await getEmployerByID(req.params.id);
+    return res.json(result);
```
fwkendall, line 85: I think we're missing logging here?

server/server.js
```diff
@@ -370,6 +380,24 @@ app.get(`${apiBaseUrl}/employer-sites`,
...
+  asyncMiddleware(async (req, res) => {
+    const user = req.hcapUserInfo;
+    const [result] = await getSiteByID(req.params.id);
+    logger.info({
+      action: 'employer-sites-detail_get',
+      performed_by: {
+        username: user.username,
+        id: user.id,
+      },
+      eeoi_id: req.query.id,
```
fwkendall, line 395: Not sure what this field is tracking - I think we're only accessing a site from this endpoint and not an EEOI right?

server/server.js
```diff
@@ -370,6 +380,24 @@ app.get(`${apiBaseUrl}/employer-sites`,
...
+      siteID: result.siteID,
```
fwkendall, line 396: should this be site_id to match the style of other keys?

server/server.js
```diff
@@ -370,6 +380,24 @@ app.get(`${apiBaseUrl}/employer-sites`,
...
+    });
+    return res.json(result);
+  }));
+
 // In development
```
fwkendall, line 401: Tiny nitpick but if we're in this file anyway can we update this comment? :)

fwkendall: Also is this endpoint deprecated now?

server/server.js
```diff
@@ -370,6 +380,24 @@ app.get(`${apiBaseUrl}/employer-sites`,
     return res.json({ data: result });
   }));
 
+app.get(`${apiBaseUrl}/employer-sites/:id`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
+  keycloak.getUserInfoMiddleware(),
```
fwkendall, line 385: we should also check the user's region to make sure they're allowed to see the site :)

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+      'First Name': 'siteContactFirstName',
+      'Last Name': 'siteContactLastName',
+    },
+    'Site Info': {
+      'Site Name': 'siteName',
+      'Business Name': 'registeredBusinessName',
+      'Address': 'address',
+      'Postal Code': 'postalCode',
+      'City': 'city',
+      'Phase 1 Allocation': 'earlyAdopterAllocation',
```
fwkendall, line 65: Oh this should actually reference the new column name, I think - I'm pretty sure the migration to rename this column has been applied to dev already so we're going to have issues deploying this as-is

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+      'Region': 'healthAuthority',
+    },
+    'Operator Info': {
+      'First Name': 'operatorContactFirstName',
+      'Last Name': 'operatorContactLastName',
+      'Phone Number': 'operatorPhone',
+      'Email address': 'operatorEmail',
```
fwkendall, line 72: should be uppercase Address to match the rest of the field names :)

client/src/pages/private/SiteTable.js
```diff
@@ -15,7 +15,9 @@ const columns = [
   { id: 'operatorName', name: 'Operator Name' },
   { id: 'healthAuthority', name: 'Health Authority' },
   { id: 'postalCode', name: 'Postal Code' },
+  { id: 'details', name: 'Details' },
```
fwkendall, line 18: I think for our other tables we didn't name this column - is this going to show up as blank? IMO we should match the style we're using elsewhere, if this doesn't already :)

server/server.js
fwkendall: Okay yes I think this is deprecated.

My vote - leave it in the codebase for now and document this in the comment here. I know Java has an @deprecated tag, not sure about JS. We can state here which endpoint replaces it and mention that legacy logs will need to be search according to the logging shown below

Thoughts?

server/server.js
Zenen-freshworks: I'm unsure about how to resolve this. Would it require rewriting of the getEmployerByID function?

server/server.js
Zenen-freshworks: Quick google says that JSDoc also has a deprecated flag.

server/server.js
fwkendall: I can think of two options offhand:

1. Get the site by ID, then compare here the site region vs. the user's regions and decide what to return accordingly
2. Add the user's regions as a param to this method, and then include it as part of the WHERE clause in the query - so a site only comes back from the DB if the HA is in the list provided *and* the ID matches

I feel like #2 is bulletproof but I'm not sure if it confuses things by moving some business logic into the DB access layer instead of the API endpoints

server/server.js
fwkendall: Awesome 🙌

client/src/pages/private/SiteTable.js
ashtonmeuser: Yep, we made it so you can omit the `name` property and the column title will be blank.

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+import Grid from '@material-ui/core/Grid';
+import { Page, CheckPermissions } from '../../components/generic';
+import { Box, Typography } from '@material-ui/core';
+import { scrollUp } from '../../utils';
+import store from 'store';
+import { Link } from 'react-router-dom';
+import routes from '../../constants/routes';
+
+
+export default ({ match }) => {
```
ashtonmeuser, line 11: Nitpick: `match` is pretty ambiguous. I know we're using `id` everywhere so maybe something like `siteId` would work.

server/server.js
ashtonmeuser: Another option would be to 
1. Add user regions to `req.userInto` object based on sites if HA, etc.
2. Check if `req.userInfo.regions` contains the region we're fetching (would fit well where we tend to do validation)

client/src/pages/private/SiteViewDetails.js
Zenen-freshworks: match is the parameter provided by React, I don't have (easy) control over how it is passed.

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+      }
+    }
+
+    fetchDetails();
+  }, [siteID]);
+
+  const fieldsLabelMap = {
+    'Site Contact': {
+      'First Name': 'siteContactFirstName',
+      'Last Name': 'siteContactLastName',
```
fwkendall, line 57: Can we get the phone + email added here too? The site contact has their own, separate to the operator :)

client/src/pages/private/SiteViewDetails.js
```diff
@@ -0,0 +1,128 @@
...
+    'Site Info': {
+      'Site Name': 'siteName',
+      'Business Name': 'registeredBusinessName',
+      'Address': 'address',
+      'Postal Code': 'postalCode',
+      'City': 'city',
+      'Phase 1 Allocation': 'earlyAdopterAllocation',
+      'Region': 'healthAuthority',
+    },
+    'Operator Info': {
```
fwkendall, line 68: I think we should rename this to Operator Contact Info

Then in Site Info, we should add the operatorName field as well
We can label it "Operator" I think - this is often a business name rather than a person, which is why it's a separate field

client/src/pages/private/SiteViewDetails.js
fwkendall: siteContactPhoneNumber
siteContactEmailAddress

ashtonmeuser (approved):

ashtonmeuser:
I'm good with this as is. This PR still has not addressed the rudimentary UI (unstyled link back to sites, lack of white card encapsulating text), but I think it's good to go. Created a ticket to address UI [here](https://freshworks.atlassian.net/browse/HCAP-382?atlOrigin=eyJpIjoiNjQyMGUzODVjOGU2NDY4ZTgwM2ZmMWMzZDFjYzM4ZGIiLCJwIjoiaiJ9).

mauriciofreshw (approved):

PR #273, opened 2021-02-11
Migration for early participants

ashtonmeuser:


server/migrations/1612489825996_create-callback-field.js
```diff
@@ -29,7 +29,7 @@ exports.up = async () => {
   await dbClient.db.withTransaction(async (tx) => {
     await tx[collections.PARTICIPANTS].updateDoc({}, { callbackStatus: false });
 
-    await tx[collections.PARTICIPANTS].updateDoc({ 'created_at <': '2021-01-01' }, { callbackStatus: true });
+    await tx[collections.PARTICIPANTS].update({ 'created_at <': '2021-01-01' }, { 'body.callbackStatus': true });
```
mauriciofreshw, line 32: Looks like we can’t query the default `created_at` through massivejs functions. I tested this solution locally and worked:

```suggestion
    await tx.query(`UPDATE ${collections.PARTICIPANTS} SET body = jsonb_set(body, '{callbackStatus}', 'true') WHERE created_at < '2021-01-01'`);
```

ashtonmeuser: Hey saw this after poking around and finding another solution. Solution I implemented was to force Massive to look at the table as normal relational (not document table) by providing an `options.body` setting. See [docs](https://massivejs.org/docs/working-with-documents#updatedoc).

Zenen-freshworks (approved):

mauriciofreshw (approved):

PR #277, opened 2021-02-12
Added cleanup migration and refactored code accordingly

mauriciofreshw:


Zenen-freshworks (approved):

server/scripts/feed-sites.js
```diff
@@ -12,7 +12,7 @@ const { createRows, verifyHeaders } = require('../utils');
       const columnMap = {
         'HCAP Site ID': 'siteId',
         'Site Name': 'siteName',
-        'Early Adopter Allocation': 'earlyAdopterAllocation',
+        'Early Adopter Allocation': 'phaseOneAllocation',
```
fwkendall, line 15: This is a bit funny but it does make sense to leave it IMO - I know about this discrepancy so I just need to be careful if any site creation operational requests come in before we add a feature to create new sites in the UI (at which point the feed sites script should be hopefully deprecated other than its utility in testing)

server/migrations/1613157291778_post-import-cleanup.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable max-len */
+/* eslint-disable quotes */
+const { dbClient } = require('../db/db.js');
+const { collections } = require('../db/schema.js');
+
+exports.up = async () => {
+  await dbClient.db.withTransaction(async (tx) => {
+    await tx.query(`UPDATE ${collections.EMPLOYER_SITES} SET body = body - 'earlyAdopterAllocation' || jsonb_build_object('phaseOneAllocation', body->'earlyAdopterAllocation')`);
```
ashtonmeuser, line 8: Did a little reading. This looks good! Relevant docs [here](https://www.postgresql.org/docs/9.5/functions-json.html).

ashtonmeuser (approved):

PR #280, opened 2021-02-17
Hcap 358 rejected report

Zenen-freshworks:


server/package.json
```diff
@@ -11,6 +11,7 @@
     "stats": "node scripts/stats.js",
     "participant-stats-hired": "node scripts/participant-stats-hired.js",
     "participant-stats-in-progress": "node scripts/participant-stats-in-progress.js",
+    "participant-stats-rejected": "node scripts/participant-stats-rejected.js",
```
fwkendall, line 14: Don't forget to update the README for this guy too 🙏

server/scripts/participant-stats-rejected.js
```diff
@@ -0,0 +1,45 @@
...
+const { getRejectedParticipantsReport } = require('../services/reporting');
+
+(async () => {
+  if (require.main === module) {
+    try {
+      await dbClient.connect();
+
+      const results = await getRejectedParticipantsReport();
+
+      console.log(results);
```
fwkendall, line 16: probably want to remove this, there should be no PII in the report but it will have HUGE output on prod

server/scripts/participant-stats-rejected.js
```diff
@@ -0,0 +1,45 @@
...
+
+      const csvStream = csv.format({ headers: true });
+
+      results.forEach((result) => {
+        csvStream.write({
+          'Participant Name': `${result.participantInfo.firstName} ${result.participantInfo.lastName}`,
```
fwkendall, line 22: Should remove this PII - all the other reports just key off IDs the program can cross-reference :)

server/scripts/participant-stats-rejected.js
```diff
@@ -0,0 +1,45 @@
...
+          'Participant ID': result.participantId,
+          'Employer ID': result.employerId,
+          'Reason for Rejection': result.rejection.final_status,
```
fwkendall, line 25: I realize the story didn't say which fields to capture, sorry - it was definitely written assuming we would expand on the in-progress report

I think the fields we should have are:

- Participant ID
- Employer ID
- Employer email
- Employer health region(s)
- Reason for Rejection
- Date*

*date of the rejection action being taken

I think these are all in the other report(s) so hopefully easy to add :)

server/tests/participant.form.test.js
```diff
@@ -582,4 +582,33 @@ describe('Participants Service', () => {
     expect(query2[0].firstName).toEqual('Eddy');
     expect(query2[0].history.length).toEqual(1);
   });
+
+  it('confirms that reporting works properly', async () => {
```
fwkendall, line 586: Love this 😍 it even answered a question I was going to ask about how participants hired elsewhere were being handled!

server/services/reporting.js
```diff
@@ -142,4 +142,29 @@ const getHiredParticipantsReport = async () => {
...
+      type: 'LEFT OUTER',
+      relation: collections.PARTICIPANTS,
+      on: {
+        id: 'participant_id',
+      },
+    },
+  }).find({
+    current: true,
+    status: 'rejected',
+    // 'employerSiteJoin.body.siteId::int >': 0, // Ensures that at least one site is found
```
fwkendall, line 157: We can definitely remove this line, this was only relevant for the hired report since that's the point at which the employer enters their site :)

server/tests/participant.form.test.js
fwkendall: Actually looking at the query, I'm wondering if we do need to exclude this particular rejection status from the report.

What I'm thinking is that a participant could be engaged and rejected, and THEN engaged and hired by another employer after the fact. So they would show up in this report with a normal rejection reason (like "withdrew") but they would actually be hired, and therefore NOT a participant that we want to prioritize in the callback process.

The simplest way to remove these participants from the report is just to exclude all participants with a hired status. That will have the knock-on effect of removing the statuses for "hired by other" unless we intentionally include that back into the report, which I don't think we need to do.

So... I think my vote is filter out participants who are hired, which will consequently filter out "hired by other", which is a totally OK outcome for this report (but does mean we need to update the expected output of this test, plus add another one for a normal rejection reason).

server/scripts/participant-stats-rejected.js
ashtonmeuser, line 16:
```suggestion
```

mauriciofreshw (approved):

ashtonmeuser:
Reviewing HCAP-358, everything works as advertised. I can run npm run participant-stats-rejects and get a report with columns Participant ID, Employer ID, Employer Email, Employer Health Regions, Reason for Rejection, and Date. Few things:
1. Judging by the [ticket](https://freshworks.atlassian.net/browse/HCAP-358?atlOrigin=eyJpIjoiYWU1YTJmYjM2MjRmNGI1MmJmZTVhZmE1ZTZlMzZiMzciLCJwIjoiaiJ9), it sounds like the intent was to have this add to the existing report rather than be a standalone report. Currently, there are three reports that exists: `participant-stats-hired`, `participant-stats-in-progress`, and `stats`. The latter is stats of the whole system e.g. # submissions by HA, total # submissions.
2. The Date column should maybe be renamed to something more descriptive i.e. Date Rejected.

mauriciofreshw (approved):

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #281, opened 2021-02-18
Withdraw duplicate participants migration

ashtonmeuser:


ashtonmeuser:
Relevant convo [here](https://freshworks.atlassian.net/browse/HCAP-374?focusedCommentId=45187).

Zenen-freshworks (dismissed):

server/migrations/1613615463607_reassign-participant-regions.js
```diff
@@ -0,0 +1,51 @@
...
+      .map((v, i) => (textualBoolean(v) ? columnValues[i] : null)) // Map from textual boolean to HA string value
+      .filter((v) => v != null) // Remove null values where column was falsy
+      .join(';'), // For some reason, we are using a semicolon-separated string instead of a JSON array here
+  }));
+
+  await dbClient.db.withTransaction(async (tx) => {
+    for (const change of changes) {
+      await tx[collections.PARTICIPANTS].updateDoc(
+        { maximusId: change.id },
+        { preferredLocation: change.preferredLocation },
```
fwkendall, line 47: we should be recording the _history_ here as well - see patch on /participant

fwkendall:
I'm not sure why the xlsx is empty or why old migrations are being edited/deleted? Can we fix those issues, along with recording the history on the participant?

ashtonmeuser:
> I'm not sure why the xlsx is empty or why old migrations are being edited/deleted? Can we fix those issues, along with recording the history on the participant?

The XLSX thing is weird. The file is certainly not empty. Viewing file directly [here](https://github.com/bcgov/hcap/blob/HCAP-374-correct-participant-regions/server/migrations/assets/reassign-participant-regions.xlsx) shows that it is 51B.

The linter was complaining about at least one of the migrations so I made some tweaks. Nothing functional but removing some linter disable lines, swap double for single quotes, etc. Can remove these edits if we'd prefer.

fwkendall:
It's probably fine to leave it, since someone else is likely just going to get caught by the linter again if we revert it

GitHub also tricked me a little as it was hiding one of the file diffs (even though it's just two lines, and not only whitespace) so I mistakenly thought a migration had been deleted 👀 but we're good

server/migrations/1613615463607_reassign-participant-regions.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const currentParticipants = await dbClient.db[collections.PARTICIPANTS].findDoc(
+    { 'maximusId::int': changes.map((i) => i.id) }, // Find all relevant participants to build history
+  );
+
+  for (const change of changes) {
+    const currentParticipant = currentParticipants.find((i) => i.maximusId === change.id);
+    // We could throw here if currentParticipant is undefined
+    // Currently, this is allowed and will set participant history item from value to null
+    // if (!currentParticipant) throw Error(`Could not find existing participant with Maximus ID ${change.id}`);
```
ashtonmeuser, line 51: On the fence re: enabling this safety check. With it, the migration will fail on dev and test. However, including it will catch the situation where a participant to be updated is not found in the current state of the DB. I think this is superfluous as I'd imagine the transaction used later would fail if this were the case.

Thanks for being my rubber duck. Fine to remove this check and comment if the team so decides.

fwkendall: I don't think we need it. If one participant among many is missing, the outcome should just be that the missing record isn't updated - I'm not sure that should be a failure case, the migration succeeds as long as it updates any records that it actually does find :)

server/migrations/1613615463607_reassign-participant-regions.js
```diff
@@ -0,0 +1,71 @@
...
+  const [headers, ...data] = sheet; // First line is headers
+
+  if (!headers.every((e, i) => columnNames[i] === e)) throw Error('Unexpected headers');
+
+  const changes = data.map((row) => ({
+    id: row?.[0],
+    preferredLocation: row
+      .slice(1) // Ignore Maximus ID column
+      .map((v, i) => (textualBoolean(v) ? columnValues[i] : null)) // Map from textual boolean to HA string value
+      .filter((v) => v != null) // Remove null values where column was falsy
```
fwkendall, line 39: We are still setting interest to false for false entries in the sheet, right?

server/migrations/1613615463607_reassign-participant-regions.js
fwkendall: Okay looking at your comments above - IMO we should just **continue** if currentParticipant is undefined, so keep this check but change the behaviour if hit

server/migrations/1613615463607_reassign-participant-regions.js
fwkendall: I see it - this is being stored as a list and we perform a full replacement of that list, so this is correct 👍

server/migrations/1613615463607_reassign-participant-regions.js
ashtonmeuser: `continue` would mean that `change.history` would be undefined. That would result in the participant still being updated but their `history` prop being set to `undefined`.

ashtonmeuser: The way it's currently written, the participant will still be updated (same as with `continue`) but will have a history array that correctly represents the change we're making. The missing `currentParticipant` is handled here using optional chaining. So if no current participant is found, `history` will look like `[{changes:[{field: 'preferredLocation', to: 'Fraser;Interior', from: null}], timestamp: '1234-56-78'}]`

server/migrations/1613615463607_reassign-participant-regions.js
```diff
@@ -0,0 +1,71 @@
...
+  const sheet = readXlsxFile.parse(file, { raw: true })?.[0]?.data; // Only need the first sheet
+  const [headers, ...data] = sheet; // First line is headers
+
+  if (!headers.every((e, i) => columnNames[i] === e)) throw Error('Unexpected headers');
+
+  const changes = data.map((row) => ({
+    id: row?.[0],
+    preferredLocation: row
+      .slice(1) // Ignore Maximus ID column
+      .map((v, i) => (textualBoolean(v) ? columnValues[i] : null)) // Map from textual boolean to HA string value
```
fwkendall, line 38: So this is not something we should change, given our timeline on this fix and the fact that it's a migration (meaning we are not depending on this code elsewhere and don't need to update/maintain it)

But I do think this is a good teaching/learning thing to point out - this is a really brittle design (again caveat, I know this was written in a hurry).

We should just be storing these values in a map to begin with, since that's fit for purpose, instead of relying on the order happening to be correct and the lengths of the arrays happening to match. This is the kind of thing that becomes very prone to breaking in code that gets used elsewhere in the app and maintained over time. Once you have things in a map, it's easy enough to just pull the keyset out if you only want that or vice versa, and now the mapping is a lot more robust and resistant to missed assumptions.

fwkendall: lol as soon as I write this I see we do check the order on line 32

I still really don't like assuming no one is going to touch your arrays though! Totally doesn't matter here but 👻

server/migrations/1613615463607_reassign-participant-regions.js
ashtonmeuser, line 51:
```suggestion
```

fwkendall: Discussed offline - we're good to remove this check 👍 the failing vs. ignoring missed records is handled on line 65, with the clause on line 66 taking care of it

fwkendall (approved):

mauriciofreshw (approved):

PR #282, opened 2021-02-19
Avoid casting Maximus ID to int which throws error on dev

ashtonmeuser:


fwkendall (approved):

mauriciofreshw (approved):

Zenen-freshworks (approved):
looks like it does exactly what it says it does

PR #284, opened 2021-02-23
Hcap 188 edit site fields

mauriciofreshw:


server/server.js
```diff
@@ -371,6 +373,28 @@ app.post(`${apiBaseUrl}/employer-sites`,
...
+    const user = req.hcapUserInfo;
+    logger.info({
+      action: 'employer-sites_patch',
+      performed_by: {
+        username: user.username,
+        id: user.id,
+      },
+      siteID: req.params.id,
+    });
+    try {
```
ashtonmeuser, line 390: Should we update the site before logging? We seem to be doing this on other PATCH methods.

server/services/employers.js
```diff
@@ -34,13 +34,17 @@ const saveSites = async (sitesArg) => {
   return response;
 };
 
+const updateSite = async (id, site) => dbClient.db[collections.EMPLOYER_SITES]
```
ashtonmeuser, line 37: Think this will require a little more work. As pointed out in the ticket, we'll need to add to the history object similar to what we're doing on other PATCH endpoints.

ashtonmeuser (changes requested):

server/server.js
```diff
@@ -371,6 +373,28 @@ app.post(`${apiBaseUrl}/employer-sites`,
     }
   }));
 
+app.patch(`${apiBaseUrl}/employer-sites/:id`,
+  keycloak.allowRolesMiddleware('ministry_of_health'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    await validate(EditSiteSchema, req.body);
```
ashtonmeuser, line 380: A little meta, but it seems `history` is included in the validation. I don't think that should be allowed to be edited.

Zenen-freshworks (approved):

ashtonmeuser (approved):

ashtonmeuser (approved):

PR #293, opened 2021-02-26
Hcap 372 last updated bug

Zenen-freshworks:
The issue was the format of datestring that was being passed to the Date() constructor, Chrome was able to interpret properly but other browsers were not.

### DIFF
--- a/client/src/utils/fuzzyDateOffset.js
+++ b/client/src/utils/fuzzyDateOffset.js
@@ -1,13 +1,16 @@
+import { stringToDateTime } from './date';
+
 export const fuzzyDateOffset = (datestring) => {
     if (!datestring) return "Never Updated";
 
     // Some browsers (firefox, safari) cannot parse datestrings of the following form:
     // 2021-02-11 16:02:12.877406+00
-    if (datestring.includes(" ")) {
-        datestring = datestring.replace(' ','T').replace('+00', 'Z')
+    let offsetTime = null;
+    if (datestring.includes(' ')) {
+        offsetTime = stringToDateTime(datestring).format('x');
+    } else {
+        offsetTime = new Date(datestring).getTime();
     }
-
-    const offsetTime = new Date(datestring).getTime();
     const currentTime = new Date().getTime();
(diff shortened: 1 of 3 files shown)


client/src/utils/fuzzyDateOffset.js
```diff
@@ -1,4 +1,12 @@
 export const fuzzyDateOffset = (datestring) => {
+    if (!datestring) return "Never Updated";
```
fwkendall, line 2: Is this possible? I thought userUpdatedAt defaulted to the current time on record creation... are there records that are actually null?

client/src/utils/fuzzyDateOffset.js
```diff
@@ -1,4 +1,12 @@
...
+
+    // Some browsers (firefox, safari) cannot parse datestrings of the following form:
+    // 2021-02-11 16:02:12.877406+00
+    if (datestring.includes(" ")) {
+        datestring = datestring.replace(' ','T').replace('+00', 'Z')
```
fwkendall, line 7: This kind of date manipulation raises red flags for me right off the bat.

You can **change** your browser date format, as I understand it this is based on your language in Chrome and Firefox and a separate setting in Safari.

There must be a lib that can help us out here. I'm convinced this is brittle. Open to being persuaded otherwise but dates always set off alarm bells for me.

client/src/utils/fuzzyDateOffset.js
Zenen-freshworks: There were records in my local database, which is why it made its way into the PR. I can remove it but I don't see it causing an issue - if we see it in a deployment then that might indicate another issue but I don't think we will.

client/src/utils/fuzzyDateOffset.js
Zenen-freshworks: Okay, I can see where you're coming from on that front. I'll do a bit more digging into a better solution than string manipulation.

fwkendall (approved):

Zenen-freshworks:
Using moment makes this way easier

PR #296, opened 2021-03-02
HCAP and Non-HCAP hires are displayed in SiteDetailView

Zenen-freshworks:
Still working on the tests, but the actual implementation is up for review

### DIFF
--- a/client/src/pages/private/SiteViewDetails.js
+++ b/client/src/pages/private/SiteViewDetails.js
@@ -103,6 +103,10 @@ export default ({ match }) => {
       'Phone Number': 'operatorPhone',
       'Email Address': 'operatorEmail',
     },
+    'Positions Overview': {
+      'HCAP Hires': 'hcapHires',
+      'Non-HCAP': 'nonHcapHires',
+    },
   };
 
   const defaultOnClose = () => {
@@ -172,8 +176,9 @@ export default ({ match }) => {
       </Dialog>
       <Page>
         <CheckPermissions isLoading={isLoadingUser} roles={roles} permittedRoles={['health_authority', 'ministry_of_health']} renderErrorMessage={true}>
+          <Box p={1} />
           <Card>
-            <Box pt={4} pb={2} pl={4} pr={4}>
+            <Box pt={0} pb={2} pl={4} pr={4}>
               <Box pb={4}>
                 <Box pb={2}>
                   <Typography variant="body1">
@@ -199,7 +204,7 @@ export default ({ match }) => {
                   </CheckPermissions>
                 </Grid>
               </Box>
-              <Grid container>
+              <Grid container direction="row">
                 {
                   Object.keys(fieldsLabelMap).map(title =>
                     <Grid key={title} item>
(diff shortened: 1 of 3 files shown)


client/src/pages/private/SiteViewDetails.js
```diff
@@ -103,6 +103,10 @@ export default ({ match }) => {
       'Phone Number': 'operatorPhone',
       'Email Address': 'operatorEmail',
     },
+    'Positions Overview': {
+      'HCAP Hires': 'hcapHires',
+      'Non-HCAP': 'nonHcapHires',
```
fwkendall, line 108: This is nitpicky but I think we should be super explicit with our naming here - designs have us potentially reporting on more of the funnel later and I think it could confuse people to have a discrepancy. So I'm thinking:
- HCAP Hires
- Non-HCAP Hires

It's repetitive but I think it will help users so they aren't wondering if it's showing them something slightly different because the naming is slightly different.

fwkendall (approved):

fwkendall:
Looks great including the test! Just one thought about the UI labelling 👍

fwkendall (approved):

server/services/employers.js
```diff
@@ -45,7 +45,14 @@ const updateSite = async (id, site) => {
...
 
 const getSites = async () => dbClient.db[collections.EMPLOYER_SITES].findDoc({});
-const getSiteByID = async (id) => dbClient.db[collections.EMPLOYER_SITES].findDoc({ id });
+const getSiteByID = async (id) => {
+  const site = await dbClient.db[collections.EMPLOYER_SITES].findDoc({ id });
+  const hcapHires = await dbClient.db[collections.PARTICIPANTS_STATUS].count({ 'data.site': site[0].siteId, 'data.nonHcapOpportunity': 'false' });
+  const nonHcapHires = await dbClient.db[collections.PARTICIPANTS_STATUS].count({ 'data.site': site[0].siteId, 'data.nonHcapOpportunity': 'true' });
+  site[0].hcapHires = hcapHires;
+  site[0].nonHcapHires = nonHcapHires;
+  return site;
```
ashtonmeuser, line 54: It's kinda weird behavior to return an array where we expect a single site. Everywhere `getSiteByID` is used, the first value is pulled from the returned array.

ashtonmeuser (approved):

server/services/employers.js
fwkendall: Yeah if we're returning a single object elsewhere, makes sense to me for consistency - I can tackle this change while Zenen is heads-down on the other ticket

fwkendall: Actually I'm not sure what's going on here. The query is the exact same for sites as it is for other single entities like a single employer, and the caller (server.js) doesn't handle it like it's an array at all.

I'm pretty sure this is exactly the point I was confused by the last time I was working in this area of the code as well - the site ID wasn't being logged properly, I concluded from running tests locally that the site was coming back as an array and accessed the ID via the first entry, but then when I pushed that code it turned out not to be correct 🤷‍♀️ 

Check where getSiteByID is called in server.js (/employer-sites/:id)

PR #297, opened 2021-03-03
Participant form endpoint

ashtonmeuser:


server/server.js
```diff
@@ -308,6 +308,37 @@ app.post(`${apiBaseUrl}/participants`,
...
+      maximusId: null,
+      postalCodeFsa: req.body.postalCode.substr(0, 3),
+      preferredLocation: req.body.preferredLocation.join(';'),
+      interested: true,
+      nonHCAP: null,
+      crcClear: null,
+      callbackStatus: false,
+      userUpdatedAt: new Date().toJSON(),
+    };
+    delete participant.consent; // Not persisted to DB
```
fwkendall, line 326: I think we actually have an obligation to save this - not losing any user-submitted data is important for FOI requests, even if the user by definition can't submit without this being true

server/validation.js
```diff
@@ -254,6 +257,26 @@ const ParticipantBatchSchema = yup.array().of(
...
 );
 
+const ParticipantSchema = yup.object().shape({
+  // Eligibility
+  eligibility: yup.boolean().typeError(errorMessage).required(errorMessage).test('is-true', errorMessage, (v) => v === true),
+
+  // Contact info
+  firstName: yup.string().required(errorMessage),
+  lastName: yup.string().required(errorMessage),
+  phoneNumber: yup.mixed().required(errorMessage),
```
fwkendall, line 267: Can we make sure we get this format validated? I think this is HCAP-347 sneaking in

fwkendall (approved):

Zenen-freshworks (approved):

PR #298, opened 2021-03-03
Hcap 392 extract report

Zenen-freshworks:
output is on the console, not yet into CSV


ashtonmeuser (dismissed):

fwkendall (dismissed):

fwkendall:
Due to the urgency of this request and the likelihood the report will be generated a single time, we have chosen not to optimize for performance but rather implementation speed. If we need to continue running this report, we'll revisit how much logic is captured outside the query.

fwkendall (approved):

ashtonmeuser (approved):

PR #305, opened 2021-03-05
Hired Participants Are displayed on SiteDetailsView

Zenen-freshworks:
Still a WIP, most logic is in place but still needs to be cleaned up

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ Dockerrun.aws.json
 audit-report.*
 .DS_Store
 .env
+.env.ocp
 cypress/screenshots
 test-results
 npm-debug.log*
--- a/.gitignore
+++ b/.gitignore
@@ -18,8 +18,7 @@ docs
 Dockerrun.aws.json
 audit-report.*
 .DS_Store
-.env
-.env.ocp
+.env*
 cypress/screenshots
 test-results
 npm-debug.log*
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -7,10 +7,10 @@ import { Table } from '../../components/generic';
 
 const columns = [
   { id: 'participantId', name: 'ID' },
-  { id: 'pptName', name: 'Name' },
-  { id: 'status', name: 'Status' },
-  { id: 'emailAddress', name: 'Email Address' },
-  { id: 'phoneNumber', name: 'Phone Number' },
+  { id: 'participantName', name: 'Name' },
+  { id: 'hiredDate', name: 'Hire Date' },
+  { id: 'startDate', name: 'Start Date' },
+  { id: 'nonHCAP', name: 'Position' },
 ];
 
 export default ({ siteId }) => {
@@ -40,14 +40,15 @@ export default ({ siteId }) => {
 
       if (response.ok) {
         const participants = await response.json();
+        console.log(participants);
         const rowsData = participants.map((row) => {
           // Pull all relevant props from row based on columns constant
           const values = {
             participantId: row.participant_id,
-            pptName: `${row.participantJoin.body.firstName} ${row.participantJoin.body.lastName}`,
-            emailAddress: row.participantJoin.body.emailAddress,
-            phoneNumber: row.participantJoin.body.phoneNumber,
-            status: row.status,
+            participantName: `${row.participantJoin.body.firstName} ${row.participantJoin.body.lastName}`,
+            hiredDate: row.data.hiredDate,
+            startDate: row.data.startDate,
+            nonHCAP: row.data.nonHcapOpportunity,
           }
 
           const mappedRow = columns.reduce((accumulator, column) => ({
@@ -95,6 +96,9 @@ export default ({ siteId }) => {
             if (columnId === 'status') {
               const status = String(row['status']);
               return `${status.substring(0,1).toUpperCase()}${status.substring(1)}`
+            }
+            if (columnId === 'nonHCAP') {
+              return (row[columnId]) ? "Non-HCAP" : "HCAP";
             }
             return (row[columnId]);
           }}
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -6,7 +6,7 @@ import store from 'store';
 import { Table } from '../../components/generic';
 
 const columns = [
-  { id: 'pptId', name: 'ID' },
+  { id: 'participantId', name: 'ID' },
   { id: 'pptName', name: 'Name' },
   { id: 'status', name: 'Status' },
   { id: 'emailAddress', name: 'Email Address' },
@@ -43,7 +43,7 @@ export default ({ siteId }) => {
         const rowsData = participants.map((row) => {
           // Pull all relevant props from row based on columns constant
           const values = {
-            pptId: row.participant_id,
+            participantId: row.participant_id,
             pptName: `${row.participantJoin.body.firstName} ${row.participantJoin.body.lastName}`,
             emailAddress: row.participantJoin.body.emailAddress,
             phoneNumber: row.participantJoin.body.phoneNumber,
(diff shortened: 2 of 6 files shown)


server/server.js
```diff
@@ -167,6 +168,28 @@ app.get(`${apiBaseUrl}/participant`,
     return res.json(result);
   }));
 
+app.get(`${apiBaseUrl}/site-participants`,
```
fwkendall, line 171: I would probably actually do this under the existing endpoint, e.g.

/employer-sites/:id/participants

server/server.js
```diff
@@ -167,6 +168,28 @@ app.get(`${apiBaseUrl}/participant`,
...
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health', 'employer'),
```
fwkendall, line 172: employer can't see this yet, should be removed :)

server/server.js
```diff
@@ -167,6 +168,28 @@ app.get(`${apiBaseUrl}/participant`,
...
+    console.log(`we got here, id is ${id}`);
+    const result = await getHiredParticipantsBySite(id);
+    logger.info({
+      action: 'site-participants_get',
+      performed_by: {
+        username: user.username,
+        id: user.id,
+      },
+      on: {
+        id,
```
fwkendall, line 187: We should log the participant IDs in addition to the site ID

server/services/participants.js
```diff
@@ -71,6 +71,22 @@ const setParticipantStatus = async (
...
+  const participants = await dbClient.db[collections.PARTICIPANTS_STATUS].join({
+    participantJoin: {
+      type: 'LEFT OUTER',
+      relation: collections.PARTICIPANTS,
+      decomposeTo: 'object',
+      on: { id: 'participant_id' },
+    },
+  }).find({
+    current: true,
+    'data.site': String(siteID),
```
fwkendall, line 84: We should have the status hired in here, too

I know we don't need it right now because sites are only tracked for a hiring event, but it's very likely that we'll change things later so that participants can have a site indicated sooner in the hiring process - at which point this would regress on us

Better to be explicit about the conditions you're looking for rather than infer them based on another piece of data :)

.gitignore
```diff
@@ -19,6 +19,7 @@ Dockerrun.aws.json
 audit-report.*
 .DS_Store
 .env
+.env.ocp
```
ashtonmeuser, line 22: What exactly is this file for? Shouldn't have to do with HCAP-394.

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -0,0 +1,105 @@
+import React, { useEffect, useState } from 'react';
+import _orderBy from 'lodash/orderBy';
+import Grid from '@material-ui/core/Grid';
+import { Box } from '@material-ui/core';
+import store from 'store';
+import { Table } from '../../components/generic';
+
+const columns = [
+  { id: 'pptId', name: 'ID' },
```
ashtonmeuser, line 9: Nitpick: I think we've been avoiding most ambiguous acronyms and shortenings. Don't think we're using `ppt` for participant anywhere else.

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -0,0 +1,105 @@
...
+    setOrder(isAsc ? 'desc' : 'asc');
+    setOrderBy(property);
+  };
+
+  const sort = (array) => _orderBy(array, [orderBy, 'operatorName'], [order]);
+
+  useEffect(() => {
+    const fetchParticipants = async () => {
+      setLoadingData(true);
+      const response = await fetch(`/api/v1/employer-sites/${siteId}/participants`, {
```
ashtonmeuser, line 36: Previously, @fwkendall mentioned this would be better suited to come with the request for site details.

server/server.js
ashtonmeuser: @Zenen-freshworks is this endpoint still supposed to exist? Noticed most of the work duplicated at :494.

server/tests/employer.form.test.js
```diff
@@ -273,6 +274,13 @@ describe('Server V1 Form Endpoints', () => {
     expect(res1.nonHcapHires).toEqual('1');
   });
 
+  it('checks response from the site participants endpoint', async () => {
+    const [site1] = await getSiteByID(1);
+    const res = await getHiredParticipantsBySite(site1.siteId);
+    // Two participants were added in the previous test
```
ashtonmeuser, line 280: I trust this works but am not a huge fan of using it. Tests should be able to be parallelized and/or run in an arbitrary order. May be another part of the upcoming test refactor.

.gitignore
Zenen-freshworks: Another .env I use for tunneling into the openshift db. I got tired of removing it from my commits.

client/src/pages/private/SiteParticipantsTable.js
ashtonmeuser: I mistook what was being said. Good as is.

.gitignore
fwkendall: Maybe we should go with .env.* ?

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -0,0 +1,105 @@
+import React, { useEffect, useState } from 'react';
+import _orderBy from 'lodash/orderBy';
+import Grid from '@material-ui/core/Grid';
+import { Box } from '@material-ui/core';
+import store from 'store';
+import { Table } from '../../components/generic';
+
+const columns = [
+  { id: 'participantId', name: 'ID' },
+  { id: 'pptName', name: 'Name' },
```
ashtonmeuser, line 10: Nitpick again: naming here
```suggestion
  { id: 'participanttName', name: 'Name' },
```

server/server.js
```diff
@@ -468,6 +469,30 @@ app.get(`${apiBaseUrl}/employer-sites/:id`,
     return res.json(result);
   }));
 
+app.get(`${apiBaseUrl}/employer-sites/:id/participants`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
+  keycloak.getUserInfoMiddleware(),
+  asyncMiddleware(async (req, res) => {
+    // await validate(ParticipantQuerySchema, req.query);
```
ashtonmeuser, line 476: Forgotten comment.

ashtonmeuser (changes requested):

server/server.js
Zenen-freshworks: Ah, I need to add validation! I totally forgot, I'll get back to this on monday

client/src/pages/private/SiteParticipantsTable.js
fwkendall: Yeah i'm also in favour of longer names. It's just too easy to get lost when a new developer joins the project, or when it hits maintenance and you don't look at it for a couple of months

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -0,0 +1,105 @@
...
+  { id: 'status', name: 'Status' },
+  { id: 'emailAddress', name: 'Email Address' },
+  { id: 'phoneNumber', name: 'Phone Number' },
```
fwkendall, line 13: We don't need these three fields - might want to check the ticket's table column list again :)

server/server.js
fwkendall: The validation required is just making sure the user has access to the site, right? should be the same as the site details endpoint?

server/tests/employer.form.test.js
fwkendall: Yeah I would not trust the order at all

Repetitive code is fine, that's where helpers come into play to help each test set up the data/conditions it needs :)

client/src/pages/private/SiteParticipantsTable.js
Zenen-freshworks: Oops! I was looking at the wireframe instead of the ticket :$

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #312, opened 2021-03-10
Migrating string falses to boolean

Zenen-freshworks:
Actually went really smoothly once I stopped overcomplicating it!

### DIFF
--- a/cypress/integration/participantView.spec.js
+++ b/cypress/integration/participantView.spec.js
@@ -11,8 +11,16 @@ describe("Participant View", () => {
     cy.get('ul.MuiMenu-list[role=listbox] > li').should('have.length', 3);
 
     // Testing for freshness
-    cy.contains('a day ago').should('exist');
-    cy.contains('14 days ago').should('exist');
+
+    cy.get('body').then(($body) => {
+      if (Cypress.$('body:contains("a day ago")').length !== 0) {
+        cy.contains('a day ago').should('exist');
+        cy.contains('14 days ago').should('exist');
+      } else {
+        cy.contains('2 days ago').should('exist');
+        cy.contains('15 days ago').should('exist');
+      }
+    });
 
     // Testing Tabs
     cy.contains("My Sites").should('not.exist');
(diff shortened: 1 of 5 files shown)


mauriciofreshw (approved):

mauriciofreshw (dismissed):

cypress/integration/participantView.spec.js
```diff
@@ -11,8 +11,16 @@ describe("Participant View", () => {
     cy.get('ul.MuiMenu-list[role=listbox] > li').should('have.length', 3);
 
     // Testing for freshness
-    cy.contains('a day ago').should('exist');
-    cy.contains('14 days ago').should('exist');
+
+    cy.get('body').then(($body) => {
+      if (Cypress.$('body:contains("a day ago")').length !== 0) {
```
Zenen-freshworks, line 16: The response of moment's .fromNow() function fluctuates by a day based on when the test is run. This should cover both cases. I don't want to spend much more time on writing this test but if it seems important to figure out exactly where the discrepancy is coming from then I'll dig deeper.

ashtonmeuser: Great [lampshading](https://tvtropes.org/pmwiki/pmwiki.php/Main/LampshadeHanging).

ashtonmeuser (approved):

cypress/integration/participantView.spec.js
fwkendall: I bet this is down to the time of the server used by GHA when running the tests (or something like that)

server/server.js
```diff
@@ -237,7 +237,7 @@ app.post(`${apiBaseUrl}/new-hired-participant`,
       [participantInfo.preferredLocation] = user.regions;
       participantInfo.crcClear = 'yes';
       participantInfo.interested = 'yes';
-      participantInfo.callbackStatus = 'false';
+      participantInfo.callbackStatus = false;
```
fwkendall, line 240: Great catch - I wouldn't have been able to see this problem in the UI due to the participant already being hired 😄

fwkendall (approved):

PR #316, opened 2021-03-15
Accessibility fixes

mauriciofreshw:
Added missing labels

### DIFF
--- a/client/src/components/fields/RenderTextField.js
+++ b/client/src/components/fields/RenderTextField.js
@@ -5,13 +5,13 @@ import { ErrorMessage } from 'formik';
 import { InputFieldLabel, InputFieldError } from '../generic';
 
 export const RenderTextField = ({
-  field: { value, ...fieldRest },
+  field: { value, name, ...fieldRest },
   form,
   label,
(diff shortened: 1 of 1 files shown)


fwkendall (approved):

client/src/components/fields/RenderTextField.js
```diff
@@ -20,16 +20,23 @@ export const RenderTextField = ({
 
   return (
     <Fragment>
-      {label && <InputFieldLabel label={label} />}
+      <label htmlFor={fieldRest.name}>
```
ashtonmeuser, line 23: Nitpick: We're using `fieldRest.name` A LOT. We could just pull this out on line 8 via `field: { value, name, ...fieldRest },`.

ashtonmeuser (dismissed):

client/src/components/fields/RenderTextField.js
mauriciofreshw: True, I completely missed that

fwkendall (approved):

ashtonmeuser (approved):

PR #319, opened 2021-03-16
Add site ID to hired participant statuses

ashtonmeuser:
Validation of the XLSX document was loosened a bit to allow the migration to pass on dev and test. Originally, it threw if an expected participant status was not found. Now, it reports that X rows were skipped. We should see that exactly 0 rows are skipped on prod and 110 rows are skipped on dev and test.

This migration also adds the first occurrence of the `history` object on the participant status `data` field. Like elsewhere, the `history` object takes the following shape.

```
{
  timestamp: new Date(),
  changes: [{
    field: 'siteId',
    from: null,
    to: 1234,
  }],
}
```


mauriciofreshw (approved):

mauriciofreshw (dismissed):

fwkendall (approved):

Zenen-freshworks (dismissed):

mauriciofreshw (approved):

server/migrations/1605755216563_initial-schema.js
```diff
@@ -1,12 +1,15 @@
...
+    await dbClient.runRawQuery(schemaItem.definition);
+  }
+  for (const schemaItem of schema.documentTables) {
     await dbClient.createDocumentTableIfNotExists(schemaItem.collection);
     for (const index of schemaItem.indexes) {
-      dbClient.createIndexIfNotExists(schemaItem.collection, index);
+      await dbClient.createIndexIfNotExists(schemaItem.collection, index);
     }
   }
+  await dbClient.reload(); // Required for subsequent migrations to see all tables
```
Zenen-freshworks, line 14: I think this might have been missing from some of my previous work. If you have two migrations that get applied at the same time, will the db need to be reloaded before the second migration can see a table made by the first one?

Zenen-freshworks (approved):

PR #322, opened 2021-03-17
Hcap 381 setup docker kc

fwkendall:


client/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "health-career-access-program-client",
   "version": "0.1.0",
   "private": true,
-  "proxy": "http://server:8080",
+  "proxy": "http://server:8081",
```
Zenen-freshworks, line 5: Is there any particular reason that this was changed?

Zenen-freshworks (dismissed):
I don't really know exactly what I'm looking at in a lot of places here but I believe that it works and I'm happy about that

client/package.json
fwkendall: @will-freshworks I realized this might not end up in your inbox unless I ping you, since I created the PR!

ashtonmeuser: Also curious about this.

will-freshworks: I needed to change the port because Keycloak also uses 8080. Due to the way the docker-compose is set-up you cannot have server and keycloak sharing the same port. This is because when the browser accesses keycloak it goes to `keycloak.local.freshworks.club:8080` which resolves to `localhost:8080`, which would clash with the server unless we moved it. We can't rebind the keycloak port because when the `server` accesses keycloak it also has to use `keycloak.local.freshworks.club:8080`  which resolves to `keycloak:8080` (inside the docker network because of the `alias` and we cannot rebind ports within the docker network, rebinding only works for ports exposed to the external host machine.

This is a little complex I realise. I might do a swamp talk on it...

will-freshworks: Oh, I actually already did a swamp talk on this I remember, it was my first one I think

.docker/keycloak/Dockerfile.test
```diff
@@ -0,0 +1,7 @@
+FROM jboss/keycloak:11.0.2
```
will-freshworks, line 1: You might want to drop this to the latest v9 rather than v11 so it matches the gov keycloak version. They are using RedHat SSO 7.4 which is KC 9.0.11 under the hood: https://access.redhat.com/articles/2342881

server/keycloak.js
```diff
@@ -84,13 +85,15 @@ class Keycloak { // Wrapper class around keycloak-connect
...
   async authenticateServiceAccount() {
+    logger.info('Authenticating Keycloak service account')
     const data = querystring.stringify({
       grant_type: 'password',
       client_id: this.clientNameBackend,
       client_secret: this.clientSecretBackend,
       username: this.serviceAccountUsername,
       password: this.serviceAccountPassword,
     });
+    logger.info(data)
```
will-freshworks, line 96: AHHHHH this is a debug line I left in by mistake. Pls remove.

will-freshworks (changes requested):

mauriciofreshw (approved):

mauriciofreshw (dismissed):

will-freshworks (dismissed):

ashtonmeuser (commented):
I believe this will fail as we do not define the `SERVER_PORT` env var in OpenShift. With this PR, it will default to 8081 which is unaligned with the service in OpenShift.

fwkendall:
Closing this so that Mauricio can re-open a new one, with himself as the author so that I can approve it :)

PR #323, opened 2021-03-17
casting value as string

Zenen-freshworks:


mauriciofreshw (approved):

server/validation.js
```diff
@@ -238,7 +238,7 @@ const ParticipantBatchSchema = yup.array().of(
       // Contact info
       firstName: yup.string().required(errorMessageIndex(index, indexName)),
       lastName: yup.string().required(errorMessageIndex(index, indexName)),
-      phoneNumber: yup.mixed().required(errorMessage).test({ name: 'isPhone', test: ((value) => !!value.match(/\d{10}/)) }),
+      phoneNumber: yup.mixed().required(errorMessage).test({ name: 'isPhone', test: ((value) => !!String(value).match(/\d{10}/)) }),
```
ashtonmeuser, line 241: This is fine as is but in the future, if we want to test against regex without storing the resulting match groups, etc., we can use `RegExp.prototype.test()`. Docs [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test). That would avoid the use of `!!`.
```suggestion
      phoneNumber: yup.mixed().required(errorMessage).test({ name: 'isPhone', test: (value) => /\d{10}/.test(String(value)) }),
```

PR #327, opened 2021-03-19
migration and endpoint are working

Zenen-freshworks:
Migration is in working order, participant intake from batch adds coordinates.


Zenen-freshworks:
const promises = rows.map((row) => dbClient.db.saveDoc(collections.PARTICIPANTS, objectMap(row)));
  const results = await Promise.allSettled(promises);

I found the above lines in services/participants, maybe integrating the Promise.allSettled() function might solve the migration problem?

server/migrations/1616112764336_populate-geocodes-table.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  readInterface.on('line', async (line) => {
+    try {
+      await dbClient.db[collections.GEOCODES].save(objectMap(line));
+    } catch (error) {
+      // 23505 is a duplication error
+      if (error.code !== 23505) throw error;
+    }
+  });
```
ashtonmeuser, line 36:
```suggestion
  const items = [];
  readInterface.on('line', (item) => items(objectMap(item)));
  await dbClient.db[collections.GEOCODES].insert(items);
  });
```

server/db/schema.js
```diff
@@ -32,6 +33,18 @@ const schema = {
         "data" jsonb
         )`,
     },
+    {
+      definition: `CREATE TABLE IF NOT EXISTS ${collections.GEOCODES} (
+			  id serial primary key, 
+        "country_code" varchar(2) not null, 
+        "postal_code" varchar(20) not null,
```
ashtonmeuser, line 40: Might wanna add a unique index to this guy.

mauriciofreshw:
Might be safer if we wrap those multiple writing db calls in a transaction to avoid inconsistencies in case of a chunk failure.

server/server.js
```diff
@@ -101,6 +102,14 @@ app.get(`${apiBaseUrl}/employer-form/:id`,
     return res.json(result);
   }));
 
+app.get(`${apiBaseUrl}/coords`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
+  asyncMiddleware(async (req, res) => {
+    const { postalCode } = req.query;
```
mauriciofreshw, line 108: Missing validation

server/server.js
```diff
@@ -101,6 +102,14 @@ app.get(`${apiBaseUrl}/employer-form/:id`,
     return res.json(result);
   }));
 
+app.get(`${apiBaseUrl}/coords`,
+  keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health'),
```
fwkendall, line 106: should be available to employers too, although I suspect you'll catch this in your next feature :)

server/server.js
fwkendall: I think this actually happens in getPointsFromPostalCodes - which returns null if it doesn't get a hit on the postal code table after handling codes with/without spaces

server/services/geocodes.js
```diff
@@ -0,0 +1,52 @@
...
+
+    if (request) {
+      // Happy path, full match on postal code
+      return {
+        lat: request.latitude,
+        lng: request.longitude,
+        match: request.postal_code.replace(' ', ''),
+      };
+    }
+    const approxRequest = await dbClient.db[collections.GEOCODES].findOne({
```
fwkendall, line 20: this is a style nitpick but I would move the newline above this, as well as the next comment explaining what this branch of the code is doing :) just keeps things a little more nicely broken up

fwkendall: like the comment from line 25 becomes line 21, line 20 is blank, and this line (20) moves to line 22

server/migrations/1616443559929_coordinate-sites-and-participants.js
```diff
@@ -0,0 +1,28 @@
...
+  const sites = await dbClient.db[collections.EMPLOYER_SITES].findDoc(
+    { coords: undefined },
+  );
+
+  const postalCodes = participants.concat(sites).map((entry) => entry.postalCode);
+  const coords = await getPointsFromPostalCodes(postalCodes);
+
+  const promises = participants.map((participant) => dbClient.db[collections.PARTICIPANTS].updateDoc(
+    { id: participant.id },
+    { coords: coords[participant.postalCode] || { match: null } },
```
fwkendall, line 19: I'm just trying to understand how the potential space in the postal codes is being handled here - I think we would be handling it on line 15 with the call to getPointsFromPostalCodes but then here we're doing an exact lookup?

Zenen-freshworks: I think I might have missed handling it here, good catch. None of the entries in the coords object have spaces, but the postal code attached to participant can.

Zenen-freshworks:
It seems like the tests are failing because they're applying the migration and all of the keys are duplicates.

mauriciofreshw (dismissed):

server/package.json
```diff
@@ -44,6 +43,7 @@
     "yup": "0.29.3"
   },
   "devDependencies": {
+    "cors": "^2.8.5",
```
mauriciofreshw, line 46: package not locked

server/services/employers.js
```diff
@@ -45,13 +46,19 @@ const updateSite = async (id, site) => {
     return { ...acc, [field]: to };
   }, { history: site.history, userUpdatedAt: new Date().toJSON() });
 
-  return dbClient.db[collections.EMPLOYER_SITES]
-    .updateDoc({ id }, changes);
+  if (changes.postalCode !== undefined) {
```
mauriciofreshw, line 49: for such comparisons, you can also simply do `if (changes.postalCode)`

server/package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "axios": "0.21.1",
     "body-parser": "1.19.0",
-    "cors": "2.8.5",
+    "cors": "^2.8.5",
```
mauriciofreshw, line 24: forgot to lock the package

server/db/schema.js
```diff
@@ -32,6 +34,26 @@ const schema = {
         "data" jsonb
         )`,
     },
+    {
+      definition: `CREATE TABLE IF NOT EXISTS ${collections.PARTICIPANTS_DISTANCE} (
+        id serial primary key,
+        "participant_id" integer not null,
+        "site_id" integer not null,
+        "distance" integer not null
```
mauriciofreshw, line 42: I think we should make distance nullable for now until we have a final decision if we should round unexpected postal codes or just leave it empty

server/services/geocodes.js
```diff
@@ -0,0 +1,89 @@
...
+  const point = await queryPoint(postalCodes);
+  return { [postalCodes]: point };
+};
+
+const getParticipantCoords = async (participantID) => {
+  const res = await dbClient.runRawQuery(`SELECT ST_AsText(coords) FROM ${collections.PARTICIPANTS} where id=${Number(participantID)};`);
+  return res;
+};
+
+const updateParticipantCoords = async (participantID) => {
```
mauriciofreshw, line 56: We don’t need this function anymore since the coords is not a column anymore and will be handled by the triggers

server/tests/geocodes.test.js
```diff
@@ -0,0 +1,76 @@
...
+      interested: 'yes',
+      nonHCAP: 'yes',
+      crcClear: 'yes',
+      preferredLocation: 'Fraser',
+    };
+
+    const res = await makeParticipant(participant);
+    const coords = await getParticipantCoords(res.id);
+
+    console.log(coords);
```
arranfw, line 42: Is this log + the commented code intended to be included?

server/tests/geocodes.test.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable no-restricted-syntax, no-await-in-loop */
+const { v4 } = require('uuid');
```
arranfw, line 2: Seems like there are some unused variables in this doc due to the commented block

server/tests/geocodes.test.js
arranfw: Shouldn't there be an `expect` somewhere in here?

server/server.js
```diff
@@ -462,6 +473,7 @@ app.patch(`${apiBaseUrl}/employer-sites/:id`,
     await validate(EditSiteSchema, req.body);
     const user = req.hcapUserInfo;
     try {
+      req.body.coords = await getPointsFromPostalCodes(req.body.postalCode);
```
arranfw, line 476: What does assigning to the request do here?

arranfw (commented):
A few small comments, can't speak too much to the majority of the implementation but I can confirm that it runs on my machine and things seem to still work. There aren't meant to be any UI changes here is that correct?

server/server.js
Zenen-freshworks: This is actually obsolete code, good catch. The req.body is passed to the updateSite function with the coords that are derived from the postal code, but we're now updating the coords via database trigger.

server/tests/geocodes.test.js
Zenen-freshworks: Yep! Never finished writing this test, apparently 😅

mauriciofreshw (dismissed):

ashtonmeuser (approved):

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #329, opened 2021-03-19
Added post import migration

mauriciofreshw:
For all participant records where the postal code is Z1Z1Z1, blank out the postal code entirely

Where the value of a field has been saved in the DB as NULL, replace a blank string or actual null - fields impacted: non-HCAP, CRC Clear. I didn’t find anything else on prod data

### DIFF
--- a/server/migrations/1616190016615_post-import-cleanup.js
+++ b/server/migrations/1616190016615_post-import-cleanup.js
@@ -1,27 +1,31 @@
-/* eslint-disable camelcase */
+/* eslint-disable camelcase, no-console */
 const { dbClient, collections } = require('../db');
 
 exports.up = async () => {
   await dbClient.db.withTransaction(async (tx) => {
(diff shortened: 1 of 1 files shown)


Zenen-freshworks (approved):

fwkendall (dismissed):

server/migrations/1616190016615_post-import-cleanup.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable camelcase */
+const { dbClient, collections } = require('../db');
+
+exports.up = async () => {
```
fwkendall, line 4: Suggestion - could we add some basic logging to this just to let us know how many records it updated? We included that in our last few migrations and it proved super useful 🙌

Zenen-freshworks (approved):

fwkendall (approved):

PR #330, opened 2021-03-24
HCAP-389: Single site POST - BE

fwkendall:
- Discovered nothing was using the existing sites POST endpoint - all our scripts to create sites go directly via the DB using saveSites in employers.js
- Repurposed the existing endpoint to be a single site POST endpoint, limited to MOH and with new validation
- The main validation difference between POST/PATCH is that single site POST requires Site ID, and also supports Operator Name - these are both already supported in the batch DB operation (which has its own validation as well)

Incidentally this is how I'm noticing we forgot to put the Operator Name in the single site view FE. 🙃 

Other cleanup:
- Split the employer form (EEOI) tests vs. site tests into separate files
- Added new tests for single site POST

### DIFF
--- a/server/services/employers.js
+++ b/server/services/employers.js
@@ -12,23 +12,8 @@ const getEmployers = async (user) => {
 const getEmployerByID = async (id) => dbClient.db[collections.EMPLOYER_FORMS].findDoc({ id });
 
 const saveSingleSite = async (siteJson) => {
-  const res = dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
+  const res = await dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
   return res;
-
-  // const res = await dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
-  // const { siteId } = res;
-  // switch (res.status) {
-  //   case 'fulfilled':
-  //     res.push({ siteId, status: 'Success' });
-  //     break;
-  //   default:
-  //     if (res.reason.code === '23505') {
-  //       res.push({ siteId, status: 'Duplicate' });
-  //     } else {
-  //       res.push({ siteId, status: 'Error', message: res.reason });
-  //     }
-  // }
-  // return res;
 };
 
 const saveSites = async (sitesArg) => {
--- a/server/services/employers.js
+++ b/server/services/employers.js
@@ -12,20 +12,23 @@ const getEmployers = async (user) => {
 const getEmployerByID = async (id) => dbClient.db[collections.EMPLOYER_FORMS].findDoc({ id });
 
 const saveSingleSite = async (siteJson) => {
-  const res = await dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
-  const { siteId } = res;
-  switch (res.status) {
-    case 'fulfilled':
-      res.push({ siteId, status: 'Success' });
-      break;
-    default:
-      if (res.reason.code === '23505') {
-        res.push({ siteId, status: 'Duplicate' });
-      } else {
-        res.push({ siteId, status: 'Error', message: res.reason });
-      }
-  }
+  const res = dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
   return res;
+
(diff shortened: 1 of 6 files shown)


server/tests/employer.site.test.js
```diff
@@ -0,0 +1,227 @@
...
+    operatorContactLastName: 'CCC',
+    operatorEmail: 'test@hcpa.fresh',
+    operatorPhone: '2219909090',
+    siteContactFirstName: 'NNN',
+    siteContactLastName: 'PCP',
+    siteContactPhoneNumber: '2219909091',
+    siteContactEmailAddress: 'test.site@hcpa.fresh',
+  };
+
+  it('Create new single site, receive success', async () => {
```
fwkendall, line 79: This test and the one below are new, everything else is copied from employer form tests (should be separate commits for easy review)

server/tests/employer.form.test.js
```diff
@@ -214,142 +155,19 @@ describe('Server V1 Form Endpoints', () => {
...
-  it('fetches the list of employers', async () => {
+  it('fetches the list of employer expression of interest forms', async () => {
     const res = await getEmployers({
       isSuperUser: false, isMoH: false, isHA: true, regions: ['Vancouver Island'],
     });
     expect(res).toEqual([]);
   });
 
-  it('fetches a single employer', async () => {
+  it('fetches a single employer expression of interest form', async () => {
```
fwkendall, line 165: Only "real" change in this file is updating the descriptions on lines 158 & 165 - the rest has just been moved to the sites test file

server/services/employers.js
```diff
@@ -11,6 +11,23 @@ const getEmployers = async (user) => {
 
 const getEmployerByID = async (id) => dbClient.db[collections.EMPLOYER_FORMS].findDoc({ id });
 
+const saveSingleSite = async (siteJson) => {
+  const res = await dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
+  const { siteId } = res;
+  switch (res.status) {
```
fwkendall, line 17: this is where things are falling over, but I can't debug locally - might need to change the way I'm attempting to save the record on line 15. I do want to be able to distinguish duplicates from other failure reasons and then have server.js return error codes the FE can use. I'm think we'll want to be able to let the user know while they're still in the modal (so if it's a dupe, we can keep the modal open with a failure for the dupe in case they just made a typo)

server/tests/employer.site.test.js
```diff
@@ -0,0 +1,248 @@
...
+    emailAddress: 'freddy@example.com',
+    interested: 'yes',
+    nonHCAP: 'yes',
+    crcClear: 'yes',
+    preferredLocation: 'Fraser',
+  };
+
+  // Used for single site POST
+  // siteId must be assigned by the test
+  const siteBaseFields = {
```
fwkendall, line 61: this is new to support single site POST

server/tests/employer.site.test.js
```diff
@@ -0,0 +1,248 @@
...
+  it('Get sites, receive all successfully', async () => {
+    const res = await getSites();
+    expect(res).toEqual(
+      expect.arrayContaining(
+        [site].map((item) => (expect.objectContaining(item))),
+      ),
+    );
+  });
+
+  it('Gets a single site', async () => {
```
fwkendall, line 138: Updated this test so it creates the site it's loading, instead of assuming which site has which ID

server/services/employers.js
```diff
@@ -11,6 +11,11 @@ const getEmployers = async (user) => {
 
 const getEmployerByID = async (id) => dbClient.db[collections.EMPLOYER_FORMS].findDoc({ id });
 
+const saveSingleSite = async (siteJson) => {
+  const res = await dbClient.db.saveDoc(collections.EMPLOYER_SITES, siteJson);
```
Zenen-freshworks, line 15: Woah, cool! I wasn't aware of this syntax.

Zenen-freshworks (approved):
Looks good! I really appreciate the work put into refactoring some of the tests to make them less interdependent.

mauriciofreshw (approved):

PR #335, opened 2021-03-27
Added hiring report download

mauriciofreshw:


server/server.js
```diff
@@ -115,6 +116,33 @@ app.get(`${apiBaseUrl}/milestone-report`,
     return res.json({ data: result });
   }));
 
+// Get hired report
+app.get(`${apiBaseUrl}/milestone-report/csv/hired`,
+  keycloak.allowRolesMiddleware('ministry_of_health'),
```
fwkendall, line 121: I noticed this was missing getUserInfoMiddleware and wondered why - then I realized we need to add logging lol

I think just logging the user id + action is sufficient here, we shouldn't need to capture any other details

server/server.js
```diff
@@ -115,6 +116,33 @@ app.get(`${apiBaseUrl}/milestone-report`,
     return res.json({ data: result });
   }));
 
+// Get hired report
+app.get(`${apiBaseUrl}/milestone-report/csv/hired`,
```
fwkendall, line 120: love it - looks very flexible for future needs 💯

fwkendall:
Looks great, nice work! Just spotted that we should add logging to the endpoint. I haven't pulled the branch to look at styling yet either so I'll check that on Monday 👍

server/server.js
```diff
@@ -115,6 +116,33 @@ app.get(`${apiBaseUrl}/milestone-report`,
...
+  keycloak.allowRolesMiddleware('ministry_of_health'),
+  asyncMiddleware(async (req, res) => {
+    res.attachment('report.csv');
+    const csvStream = csv.format({ headers: true });
+    csvStream.pipe(res);
```
Zenen-freshworks, line 125: How does this work? Piping into the response, and when the csvStream is closed then it sends the response as a blob?

server/server.js
mauriciofreshw: Totally missed that 🤦🏻 thanks!!

server/server.js
mauriciofreshw: It actually sends as it goes, not when it finishes, AFAI.

Zenen-freshworks: Oh, interesting. I'm trying to wrap my head around how it would know when to actually return the response blob to the client, but if it works then it isn't a big deal.

Zenen-freshworks (approved):

fwkendall (approved):

PR #340, opened 2021-03-31
Added fully e2e endpoint tests

mauriciofreshw:
POC for HCAP-369 and GH Actions test.


.github/workflows/test.yml
```diff
@@ -32,33 +32,15 @@ jobs:
...
+
 
-      - name: NPM Test
-        working-directory: ./server
-        env:
-          POSTGRES_DB: db_test
-          POSTGRES_USER: admin
-          POSTGRES_PASSWORD: admin
-        run: |
-          npm ci
```
fwkendall, line 63: we should make this this is added to the makefile local-server-tests if we're removing it from here 😄

.docker/keycloak/users.json
```diff
@@ -0,0 +1,98 @@
+[
```
fwkendall, line 1: curious how this was initially created - to add more users would I just mock them out here and trust the import/export scripts to manage things from there, seeing as the IDs aren't used during import?

.github/workflows/test.yml
mauriciofreshw: _Dockerfile.dev_ is already doing this work for us

.docker/keycloak/users.json
mauriciofreshw: You can do it either this way or via UI, just make sure that all the ids under the `roles` field are correct. And if you choose to change via UI don’t forget to `make kc-export-users` afterward.

fwkendall: Gotcha, so first time you add new users would mean following the README steps to get into the local KC and set them up for export 😄  makes sense!

fwkendall (approved):

ashtonmeuser (approved):

PR #342, opened 2021-04-06
Added need help info

mauriciofreshw:


client/src/components/participant-form/Fields.js
```diff
@@ -263,6 +263,18 @@ export const Fields = ({ isDisabled }) => {
             Use the submit button to complete your expression of interest. We will contact you within 3
             weeks with more information.
           </Typography>
+          <Typography variant="body1" gutterBottom>
```
fwkendall, line 266: I was hoping for a closer match to what we did on the EEOI - this "need help" is at the top of the page there, and is in an alert `<Alert severity="info">`

mauriciofreshw: Ah, good catch, totally missed that

fwkendall (approved):

ashtonmeuser (approved):

PR #345, opened 2021-04-24
OCP4 Pipeline

ashtonmeuser:


server/server.js
```diff
@@ -35,9 +35,7 @@ const keycloak = require('./keycloak.js');
 const apiBaseUrl = '/api/v1';
 const app = express();
 
-if (process.env.NODE_ENV === 'local') {
-  app.use(cors());
-}
+app.use(cors());
```
ashtonmeuser, line 38: This should be reverted once FE and BE are on same OCP cluster.

openshift/patroni.bc.yml
```diff
@@ -13,7 +13,7 @@ parameters:
   - name: GIT_URI
     description: Source for Postgres repository with Dockerfile
     required: true
-    value: https://github.com/BCDevOps/platform-services.git
+    value: https://github.com/ashtonmeuser/platform-services.git
```
ashtonmeuser, line 16: This is problematic.

server/server.js
arranfw: Is this true now?

.github/workflows/promote.yml
```diff
@@ -7,8 +7,8 @@ on:
       - prod
 
 env:
-  CLUSTER: https://console.pathfinder.gov.bc.ca:8443
-  AUTH_TOKEN: ${{ secrets.AUTH_TOKEN }}
+  CLUSTER: https://api.silver.devops.gov.bc.ca:6443
+  AUTH_TOKEN: ${{ secrets.AUTH_TOKEN_OCP4 }}
```
arranfw, line 11: When OCP3 is gone are we reverting this change? Is there a ticket made?

vyasworks (approved):

Makefile
```diff
@@ -102,9 +102,14 @@ tag-prod:
 
 # OpenShift Aliases
 
-server-prep: # TODO: Move role binding command to a server prep template
+add-role:
+	@oc policy add-role-to-user admin system:serviceaccount:$(TARGET_NAMESPACE):default -n $(TOOLS_NAMESPACE)
```
arranfw, line 106: As a newcomer should I know what these commands are used for? Perhaps there's some room for documentation here?

arranfw (approved):

server/server.js
ashtonmeuser: This is not yet true. GDX has one DNS record left to swap over. There's a ticket that captures this but I don't see it as super high priority.

Makefile
ashtonmeuser: Lots of room for documentation. We were hoping for these to be roughly self-explanatory but failed on that front. Many of these were added as we found gaps between our Makefile targets and what was required to fully stand up our infra.

PR #348, opened 2021-04-28
HCAP-467 Update PEOI Contact

arranfw:
https://freshworks.atlassian.net/browse/HCAP-467

- [x] Tested on dev

This PR is a good opportunity to talk about styles :D 

I've hardcoded colors here based on what I saw in the [ticket](https://freshworks.slack.com/archives/C011G18PW5S/p1618267252263300?thread_ts=1618248466.243700&cid=C011G18PW5S), normally I'd like all of these colors to be defined in the theme. What do you think?

Also, do you go pixel-perfect with designs or do you use the closest approximation with MUI components? Ideally our MUI theme should be 1-1 with designs and there wouldn't be too much hand-CSS to write.

Hex or RGB?

### DIFF
--- a/client/src/components/participant-form/Fields.js
+++ b/client/src/components/participant-form/Fields.js
@@ -6,12 +6,29 @@ import Box from '@material-ui/core/Box';
 import Typography from '@material-ui/core/Typography';
 import { makeStyles } from '@material-ui/core/styles';
 import { FastField } from 'formik';
+import { SpeakerNotesOutlined } from '@material-ui/icons';
+
 
 import { Card, Divider } from '../generic';
 import { RenderCheckbox, RenderCheckboxGroup, RenderTextField, RenderRadioGroup } from '../fields';
-import Alert from '@material-ui/lab/Alert';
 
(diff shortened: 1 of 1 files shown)


client/src/components/participant-form/Fields.js
```diff
@@ -25,18 +42,13 @@ export const Fields = ({ isDisabled }) => {
...
-            <b>Need Help?</b><br />
-            Contact a Health Career Access Program agent.
-            <br />
-            <Link href="tel:+18773740463">
-              Dial 1-877-374-0463
-            </Link>
-            <b> Service is available from 8:00 am - 4:30 pm Pacific Time</b>
-          </Typography>
-        </Alert>
+      <Box mb={2} py={1} px={2} display="flex" alignItems="center" className={classes.info}>
```
arranfw, line 45: I'm using MUI props here in addition to a class for some other styles. Do we prefer to use all props or all  classes?

ashtonmeuser (approved):

ashtonmeuser (approved):

mauriciofreshw (approved):

PR #349, opened 2021-04-28
Logo links to home page

ashtonmeuser:


arranfw:
Should we use Link from react-router now that this is internal?

mauriciofreshw (approved):

arranfw (approved):

ashtonmeuser:
> Should we use Link from react-router now that this is internal?

Sigh... You're right. Let's do it right. Done.

arranfw:
> > Should we use Link from react-router now that this is internal?
> 
> Sigh... You're right. Let's do it right. Done.

Wait is this just a styled <a> from MUI? Or do they use react-router under the hood

mauriciofreshw (approved):

PR #354, opened 2021-05-04
Added timming to seach fields

dbayly-freshworks:
Added trimming to search fields. 
Added checks so that the page will not refresh if the text does not change after refresh

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -173,36 +173,36 @@ export default () => {
   };
 
   const handleFsaFilter = (value) => {
-    if(value !== value.trim()){
+    if(fsaText && fsaText.trim() === value.trim()){
       return
     }
     setPagination(oldPagination => ({
       ...oldPagination,
       currentPage: 0,
     }));
-    setFsaFilter(value);
+    setFsaFilter(value.trim());
   };
 
   const handleLastNameFilter = (value) => {
-    if(value !== value.trim()){
+    if(lastNameText && lastNameText.trim() === value.trim()){
       return
     }
     setPagination(oldPagination => ({
       ...oldPagination,
       currentPage: 0,
     }));
-    setLastNameFilter(value);
+    setLastNameFilter(value.trim());
   };
 
   const handleEmailFilter = (value) => {
-    if(value !== value.trim()){
+    if(emailText && emailText.trim() === value.trim()){
       return
     }
     setPagination(oldPagination => ({
       ...oldPagination,
       currentPage: 0,
     }));
-    setEmailFilter(value);
+    setEmailFilter(value.trim());
   };
 
   const filterData = (data, columns) => {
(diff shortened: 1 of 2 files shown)


Zenen-freshworks (approved):
Looks good and works as intended! We really gotta revisit those filter functions...

arranfw:
This seems to prevent me from adding spaces to filters at all, which could be problematic when searching for names with spaces

image

dbayly-freshworks:
I've removed the trimming onChange. 
Now the only trimming occurs when a query is sent. This should allow for spaces in between words. It will not search if the space added doesn't change the value when trimmed. 
Screen Shot 2021-05-05 at 9 43 52 AM

arranfw (dismissed):
Much better

client/src/pages/private/ParticipantTable.js
```diff
@@ -19,7 +19,8 @@ import {
 import { Table, CheckPermissions, Button, Dialog } from '../../components/generic';
 import { ProspectingForm, InterviewingForm, RejectedForm, HireForm, NewParticipantForm, EditParticipantForm } from '../../components/modal-forms';
 import { useToast } from '../../hooks';
-import { DebounceTextField } from '../../components/generic/DebounceTextField';
+import { DebounceTextField } from '../../components/generic/DebounceTextField'
+;
```
arranfw, line 23: Extra new-line here

client/src/pages/private/ParticipantTable.js
```diff
@@ -173,27 +173,36 @@ export default () => {
...
       currentPage: 0,
     }));
-    setFsaFilter(value);
+    setFsaFilter(value.trim());
   };
 
   const handleLastNameFilter = (value) => {
-    setPagination(prev => ({
-      ...prev,
+    if(lastNameText && lastNameText.trim() === value.trim()){
```
arranfw, line 187:
```suggestion
    if(lastNameText?.trim() === value?.trim()){
```

Little bit less wordy option

client/src/pages/private/ParticipantTable.js
```diff
@@ -173,27 +172,33 @@ export default () => {
...
-    setFsaFilter(value);
+    setFsaFilter(value.trim());
   };
 
   const handleLastNameFilter = (value) => {
-    setPagination(prev => ({
-      ...prev,
+    if(lastNameText?.trim() === value?.trim()) return;
+    
+    setPagination(oldPagination => ({
```
arranfw, line 187: I'm doing a bad job reviewing this... I'm so sorry, these `old____` variables were recently changed to `prev` for consistency, I think this is an unintentional change

dbayly-freshworks: All good. I'll make the change.

arranfw (approved):

Zenen-freshworks (approved):
nice, this is way more concise

PR #362, opened 2021-05-06
WIP: Hcap 400 add use reducer -- Feedback welcome

dbayly-freshworks:
I'm looking for some feedback here. I'm not sure if this PR is in line with the ticket. 

I've replaced all the Text and Filter `setState` functions with a single `useReducer`.
I've move as many of the constants out of the `participantsTable` file and into a constants file. 

If this is inline with the intent of the ticket, I would like to expand this reducer.  Almost all of our handler functions are copied code, so we could really cut down on our line count here. 

Anyways, let me know what you think.


client/src/pages/private/ParticipantTable.js
```diff
@@ -123,19 +74,33 @@ export default () => {
...
-  const [emailText, setEmailText] = useState(null);
   const [hideLastNameAndEmailFilter, setHideLastNameAndEmailFilter] = useState(true);
   const [actionMenuParticipant, setActionMenuParticipant] = useState(null);
   const [anchorElement, setAnchorElement] = useState(false);
   const [activeModalForm, setActiveModalForm] = useState(null);
   const [tabValue, setTabValue] = useState(null);
-
   const [locations, setLocations] = useState([]);
+
+  const textReducer = (state,action)=>{
```
arranfw, line 84: I typically declare reducers outside of the component so it isn't recreated on re-render, small improvement what do you think?

dbayly-freshworks: Seems like an improvement to me.
Thanks!

client/src/constants/participantTableConstants.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+export const defaultTableState = {
+    fsaText:'',
+    emailText:'',
+    lastNameText:'',
+    fsaFilter:'',
+    emailFilter:'',
+    lastNameFilter:''
+  }
```
Zenen-freshworks, line 93: great call on getting this pulled out into a separate file!

client/src/pages/private/ParticipantTable.js
```diff
@@ -175,36 +144,6 @@ export default () => {
     }));
   };
```
Zenen-freshworks, line 145: would it make sense to bring these functions into the reducer as well? I kind of want everything that uses setOrder or setPagination to sit in the same place, could that work effectively?

client/src/pages/private/ParticipantTable.js
```diff
@@ -302,34 +241,7 @@ export default () => {
...
-            status: ToastStatus.Info,
-            message: `${firstName} ${lastName} has been archived`,
-          },
-          already_hired: {
-            status: ToastStatus.Info,
-            message: `${firstName} ${lastName} is already hired by someone else`,
-          }
-        };
-
+        const toasts = makeToasts(firstName,lastName,ToastStatus);
```
Zenen-freshworks, line 244: Do we need to import ToastStatus here? Can we import it in the constants doc and remove it as a parameter in the makeToasts function?

client/src/pages/private/ParticipantTable.js
```diff
@@ -500,7 +412,7 @@ export default () => {
       });
     };
     runAsync();
-  }, [pagination.currentPage, locationFilter, siteSelector, fsaFilter, lastNameFilter, emailFilter, order, tabValue]);
+  }, [pagination.currentPage, locationFilter, siteSelector, textState.emailFilter, textState.locationFilter, textState.lastNameFilter,  textState.fsaFilter, order, tabValue]);
```
Zenen-freshworks, line 415: Is there any reason not to just pass in textState here instead of all the individual objects?

Zenen-freshworks (approved):
off to a good start! I think that this is going to make the ParticipantTable way more usable when its done, and also provide a good base for using reducers in other parts of our code.

client/src/pages/private/ParticipantTable.js
dbayly-freshworks: I'm all for it!

client/src/pages/private/ParticipantTable.js
arranfw: This effect needs a ticket of it's own doesn't it 👀

client/src/pages/private/ParticipantTable.js
```diff
@@ -767,44 +703,62 @@ export default () => {
...
+                  value={reducerState.emailText}
+                  disabled={isLoadingData}
+                  onDebounce={(text) => dispatch({type:'updateFilter',key:'emailFilter', value:text})}
+                  onChange={({ target }) => dispatch({type:'updateKey',key:'emailText',  value:target.value})}
+                  placeholder='Email'
+                />}
+              </Box>
+            </Grid>
+            <Grid item style={{ 'marginLeft': 20 }}>
+              <Typography>Site for distance calculation: </Typography>
```
arranfw, line 791: This input got duplicated at some point

client/src/pages/private/ParticipantTable.js
```diff
@@ -387,25 +285,28 @@ export default () => {
...
-      order.direction,
-      siteSelector,
-      tabs[tabValue].statuses
+      reducerState.locationFilter,
+      reducerState.fsaFilter,
+      reducerState.lastNameFilter,
+      reducerState.emailFilter,
+      reducerState.order.field,
+      reducerState.order.direction,
+      reducerState.siteSelector,
```
arranfw, line 300: I'd be partial to destructuring these values at the start of the function to avoid the repitition, bit of a nit though.

client/src/pages/private/ParticipantTable.js
```diff
@@ -475,23 +377,26 @@ export default () => {
...
-        lastNameFilter,
-        emailFilter,
-        order.field,
-        order.direction,
-        siteSelector,
-        tabs[tabValue].statuses
+        reducerState.locationFilter,
+        reducerState.fsaFilter || '',
+        reducerState.lastNameFilter || '',
+        reducerState.emailFilter || '',
```
arranfw, line 387: Same nit here

arranfw (changes requested):
One change and a couple nits, this is awesome work though, love it!

We'll also need to look into why that test is failing, all seem to be passing locally which is strange

client/src/pages/private/ParticipantTable.js
dbayly-freshworks: I played around with deconstructing these, but I ran into an issue where the useEffect required adding the entire reducer state to the it's listener. The result was an infinite loop. 

I think I am going to hold off implementing this for now, but I am all onboard with refactoring the useEffect later.

arranfw: Oooh gotcha, yeah let's save this for another day then 👍

client/src/pages/private/ParticipantTable.js
dbayly-freshworks: It must have been during the rebase, I've removed it.

Zenen-freshworks (approved):
Seems like Arran's nitpicks were resolved and nothing else stands out to me!

arranfw (approved):

PR #367, opened 2021-05-10
firstly, fixed farness fathoming. Furthermore, finalized fuzzy matching

Zenen-freshworks:
Basically just rewrote the update_postal_code triggers to add a thing or two.

### DIFF
--- a/server/migrations/1620672944865_update-triggers.js
+++ b/server/migrations/1620672944865_update-triggers.js
@@ -1,0 +1,105 @@
+/* eslint-disable camelcase */
+
+const { collections } = require('../db');
+
+exports.up = async (pgm) => {
+  await pgm.dropTrigger(collections.EMPLOYER_SITES, 'update_postal_code_s', { ifExists: true });
+  await pgm.createTrigger(collections.EMPLOYER_SITES, 'update_postal_code_s', {
+    when: 'BEFORE',
+    operation: ['INSERT', 'UPDATE OF body'],
+    language: 'plpgsql',
+    level: 'ROW',
+    replace: true,
+  },
+  `
+  DECLARE
+   loc record;
+   participant record;
+  BEGIN
+    IF NEW.body->>'postalCode' = OLD.body->>'postalCode' THEN
+      RETURN NEW;
+    END IF;
+
+    SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE REPLACE(postal_code, ' ', '')=REPLACE(NEW.body->>'postalCode', ' ', '');
+
+    IF NOT FOUND THEN
+      SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE SUBSTRING(REPLACE(postal_code, ' ', ''),1,5)=SUBSTRING(REPLACE(NEW.body->>'postalCode', ' ', ''),1,5);
+    END IF;
+
+    IF NOT FOUND THEN
+      IF NEW.body->>'location' IS NOT NULL THEN
+        NEW.body = NEW.body::jsonb - 'location';
+        DELETE FROM participants_distance where site_id = OLD.id;
+      END IF;
+      RETURN NEW;
+    END IF;
+
+    NEW.body = jsonb_set(NEW.body::jsonb, '{location}', ('{"type":"Point","coordinates":[' || loc.longitude || ',' || loc.latitude || ']}')::jsonb);
+    IF OLD IS NULL OR OLD.body->>'location' IS NULL THEN
+      FOR participant IN
+        SELECT * FROM participants WHERE body->>'location' IS NOT NULL
+      LOOP
+        INSERT INTO participants_distance (participant_id, site_id, distance) VALUES
+        (participant.id, (NEW.body->>'siteId')::int, ST_DistanceSphere(ST_GeomFromGeoJSON(NEW.body->>'location'), ST_GeomFromGeoJSON(participant.body->>'location')));
+      END LOOP;
+    ELSE
+      FOR participant IN
+        SELECT * FROM participants
+      LOOP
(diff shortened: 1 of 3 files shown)


server/migrations/1620672944865_update-triggers.js
```diff
@@ -0,0 +1,105 @@
...
+  BEGIN
+    IF NEW.body->>'postalCode' = OLD.body->>'postalCode' THEN
+      RETURN NEW;
+    END IF;
+
+    SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE REPLACE(postal_code, ' ', '')=REPLACE(NEW.body->>'postalCode', ' ', '');
+
+    IF NOT FOUND THEN
+      SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE SUBSTRING(REPLACE(postal_code, ' ', ''),1,5)=SUBSTRING(REPLACE(NEW.body->>'postalCode', ' ', ''),1,5);
+    END IF;
```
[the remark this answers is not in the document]
arranfw: Small performance consideration where we're doubling this query for users with invalid postal codes. Not a huge concern as I don't think that case represents most of the database.

server/migrations/1620672944865_update-triggers.js
```diff
@@ -0,0 +1,105 @@
...
+
+    IF NOT FOUND THEN
+      IF NEW.body->>'location' IS NOT NULL THEN
+        NEW.body = NEW.body::jsonb - 'location';
+        DELETE FROM participants_distance where site_id = OLD.id;
```
arranfw, line 32: Is this logic wrong? If the new location is not null, we delete and remove location?

Also it seems strange to check for NEW.body and then use OLD.id

arranfw (changes requested):

server/migrations/1620672944865_update-triggers.js
Zenen-freshworks: OLD.id and NEW.id are semantically equal. The logic overall states that if the postal code was updated with a value that wasn't found in the geocodes database, check if the new entry is holding a location value. If it is, then it is no longer relevant and we remove it along with the related entries in the distance table.

server/migrations/1620672944865_update-triggers.js
```diff
@@ -0,0 +1,105 @@
...
+      RETURN NEW;
+    END IF;
+
+    SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE REPLACE(postal_code, ' ', '')=REPLACE(NEW.body->>'postalCode', ' ', '');
+
+    IF NOT FOUND THEN
+      SELECT longitude::text,latitude::text INTO loc FROM geocodes WHERE SUBSTRING(REPLACE(postal_code, ' ', ''),1,5)=SUBSTRING(REPLACE(NEW.body->>'postalCode', ' ', ''),1,5);
+    END IF;
+
+    IF NOT FOUND THEN
```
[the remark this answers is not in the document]
arranfw: Ahh totally, nice

server/migrations/1620672944865_update-triggers.js
```diff
@@ -0,0 +1,105 @@
...
+      IF NEW.body->>'location' IS NOT NULL THEN
+        NEW.body = NEW.body::jsonb - 'location';
+        DELETE FROM participants_distance where site_id = OLD.id;
+      END IF;
+      RETURN NEW;
+    END IF;
+
+    NEW.body = jsonb_set(NEW.body::jsonb, '{location}', ('{"type":"Point","coordinates":[' || loc.longitude || ',' || loc.latitude || ']}')::jsonb);
+    IF OLD IS NULL OR OLD.body->>'location' IS NULL THEN
```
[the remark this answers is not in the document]
arranfw: This is the case when a postal code is updated from being invalid to being valid.

We don't want to enter into the 'update' block because it relies on the records existing in the distance table ✔️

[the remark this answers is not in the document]
arranfw: To clarify again for future readers, this is the case when a postal code is updated from being invalid to valid

arranfw:
Tough to see the recent changes on a force-push :P what changed?

arranfw (approved):

server/migrations/1620672944865_update-triggers.js
ashtonmeuser, line 32: I think this should be referencing `body->>'siteId'`, no? For participants `id` is used but for sites we're not using the default `id` column but rather the `siteId` in the JSONB.

arranw: This is the regular `site_id` column on `participants_distance` not a jsonb right?

arranfw: This is the regular site_id column on participants_distance not a jsonb right?

Zenen-freshworks: Ah yes, this is valid. I really don't have a strong understanding of the difference between how id and siteId are used in sites. I think that you're probably right to suggest this correction here, I'll update

server/migrations/1620672944865_update-triggers.js
```diff
@@ -0,0 +1,105 @@
...
+const { collections } = require('../db');
+
+exports.up = async (pgm) => {
+  await pgm.dropTrigger(collections.EMPLOYER_SITES, 'update_postal_code_s', { ifExists: true });
+  await pgm.createTrigger(collections.EMPLOYER_SITES, 'update_postal_code_s', {
+    when: 'BEFORE',
+    operation: ['INSERT', 'UPDATE OF body'],
+    language: 'plpgsql',
+    level: 'ROW',
+    replace: true,
```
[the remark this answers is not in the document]
Zenen-freshworks: Yea, I can't find it anywhere actually. I'll try removing it and see if it breaks anything.

ashtonmeuser (approved):

arranfw (approved):

PR #368, opened 2021-05-11
[HCAP-69] Version End point - Git SHA

vyasworks:
provide ARG for OCR docker build 

https://freshworks.atlassian.net/browse/HCAP-69

In order to make version endpoint return git version of deployed code, we need to pass this to the ENV. 

we could do this using OC CLI method we are already using 

OR 

recreate K8S resources for BC / DC using this parameter in the build config. 

(https://docs.openshift.com/container-platform/4.2/builds/build-strategies.html#builds-strategy-docker-build-arguments_build-strategies)

The former is easier change and meets our requirement

NOTE: 

this commit ID is not reliable in DEV env since developers can directly tag and could later do a `force push`, so the commit id will be untraceable in master. 

However it is very well stable for any commit that has landed in `master`

This problem is now solved with a richer version message: 

```
{
    "version": "5243b15, v-y-a-s, ungraft, HEAD, tag: refs/tags/dev, refs/remotes/origin/version_ep"
}
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          ref: 'dev'
+          fetch-depth: 0
 
       - name: Cache OpenShift CLI
         id: cache-oc
@@ -55,6 +58,7 @@ jobs:
         run: |
           cd "$GITHUB_WORKSPACE"
           oc login --token="$AUTH_TOKEN" --server="$CLUSTER"
+          make print-status
           make server-build
           make server-deploy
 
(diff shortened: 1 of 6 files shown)


vyasworks:
> Is this PR doing more than just adding the version endpoint? I seemed like the version endpoint was working with the one-line change this morning

The oneline change earlier had just the git sha, as opposed to what we have now : 

```"version": "5243b15, v-y-a-s, ungraft, HEAD, tag: refs/tags/dev, refs/remotes/origin/version_ep"```

arranfw:
> > Is this PR doing more than just adding the version endpoint? I seemed like the version endpoint was working with the one-line change this morning
> 
> The oneline change earlier had just the git sha, as opposed to what we have now :
> 
> `"version": "5243b15, v-y-a-s, ungraft, HEAD, tag: refs/tags/dev, refs/remotes/origin/version_ep"`

🤦 Didn't see the updated post body, thanks!

arranfw (approved):

PR #370, opened 2021-05-11
don't click on the disabled button

Zenen-freshworks:
I spent some time trying to figure out what was causing it to fail now and not before, but eventually decided that clicking on the button wasn't necessary for testing either way. Simple fix.


arranfw (approved):
If the goal of the test is to ensure the dropdown is disabled and we consider having the `Mui-disabled` class to be synonymous with it not being actionable then I can see this being appropriate.

I'm assuming line 32 confirms that the button exists

PR #371, opened 2021-05-11
Hcap 536 filter participants for email blast

dbayly-freshworks:
Added a simple script to get all unique email addresses from participants that are: 
- 6 weeks old or older
- Still active in the program 
- At least one application is not hired

### DIFF
--- a/server/scripts/get-email-list.js
+++ b/server/scripts/get-email-list.js
@@ -16,11 +16,8 @@ require('dotenv').config({ path: '../.env' });
         AND (CAST(p.body ->> 'interested' AS TEXT) = 'yes'  
             OR CAST(p.body->> 'interested' AS TEXT) IS NULL)
         AND (
-                (p.updated_at < CURRENT_TIMESTAMP - interval '6 weeks'
-                AND p.updated_at IS NOT NULL)
-                OR (p.updated_at IS NULL 
-                    AND p.created_at < CURRENT_TIMESTAMP - interval '6 weeks')
-	     )`);
+            to_timestamp(p.body->>'userUpdatedAt', 'YYYY-MM-DD') < CURRENT_TIMESTAMP - interval '6 weeks'
+	      );
   console.log(res);
   process.exit(0);
 })();
--- a/server/scripts/get-email-list.js
+++ b/server/scripts/get-email-list.js
@@ -17,7 +17,8 @@ require('dotenv').config({ path: '../.env' });
             OR CAST(p.body->> 'interested' AS TEXT) IS NULL)
         AND (
             to_timestamp(p.body->>'userUpdatedAt', 'YYYY-MM-DD') < CURRENT_TIMESTAMP - interval '6 weeks'
-	      );
+        )`
+  );
   console.log(res);
   process.exit(0);
 })();
--- a/server/scripts/get-email-list.js
+++ b/server/scripts/get-email-list.js
@@ -4,7 +4,7 @@ require('dotenv').config({ path: '../.env' });
 (async function exec() {
   await dbClient.connect();
   const res = await dbClient.db.query(`
-    SELECT p.body ->> 'emailAddress' AS email
+    SELECT DISTINCT p.body ->> 'emailAddress' AS email
     FROM participants p 
     WHERE 
         p.body ->> 'emailAddress' NOT IN (
--- a/server/scripts/get-email-list.js
+++ b/server/scripts/get-email-list.js
@@ -1,0 +1,14 @@
+const { dbClient } = require('../db');
+require('dotenv').config({ path: '../.env' });
+(async function exec(){
+    await dbClient.connect();
+    const res = await dbClient.db.query(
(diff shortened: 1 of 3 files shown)


server/scripts/get-email-list.js
```diff
@@ -0,0 +1,18 @@
+const { dbClient } = require('../db');
```
arranfw, line 1:
```suggestion
/* eslint-disable no-console */
const { dbClient } = require('../db');
```

server/scripts/get-email-list.js
```diff
@@ -0,0 +1,18 @@
...
+    await dbClient.connect();
+    const res = await dbClient.db.query(
+   `SELECT DISTINCT body->> 'emailAddress' AS email 
+    FROM participants p 
+    JOIN participants_status ps 
+        ON  p.id = ps.participant_id AND ps.status != 'hired'
+    WHERE 
+        CAST(p.body ->> 'interested' AS TEXT) = 'yes'  
+        OR CAST(p.body->> 'interested' AS TEXT) IS NULL
+        AND p.created_at < CURRENT_TIMESTAMP - interval '6 weeks'`)
```
arranfw, line 13: @ashtonmeuser  or @Zenen-freshworks can confirm but I think we want `body->>'userUpdatedAt'` here based on AC: 
> Send to PEOIs with an updated at date past a specific window (must be able to enter 6 weeks or a date that reflects records older than 6 weeks)

server/scripts/get-email-list.js
```diff
@@ -0,0 +1,18 @@
+const { dbClient } = require('../db');
+require('dotenv').config({ path: '../.env' });
+(async function exec(){
+    await dbClient.connect();
+    const res = await dbClient.db.query(
+   `SELECT DISTINCT body->> 'emailAddress' AS email 
+    FROM participants p 
+    JOIN participants_status ps 
+        ON  p.id = ps.participant_id AND ps.status != 'hired'
```
Zenen-freshworks, line 9: This is an inner join by default, right? We would want that because we only want the status and participants that are connected to one another by ID and meet the join criteria?

arranfw (commented):
could you run  `npm run format` 😄

server/scripts/get-email-list.js
```diff
@@ -0,0 +1,18 @@
...
+    WHERE 
+        CAST(p.body ->> 'interested' AS TEXT) = 'yes'
```
Zenen-freshworks, line 11: is p.body->>'interested'::text = 'yes' another option instead of using the CAST function? I think they're equivalent, I just haven't seen this one before.

Zenen-freshworks (dismissed):

server/scripts/get-email-list.js
dbayly-freshworks: Sounds good, I'll make the change!

server/scripts/get-email-list.js
```diff
@@ -0,0 +1,26 @@
...
+            ON p2.id = ps.participant_id AND ps.status = 'hired'
+        )
+        AND (CAST(p.body ->> 'interested' AS TEXT) = 'yes'  
+            OR CAST(p.body->> 'interested' AS TEXT) IS NULL)
+        AND (
+                (p.updated_at < CURRENT_TIMESTAMP - interval '6 weeks'
+                AND p.updated_at IS NOT NULL)
+                OR (p.updated_at IS NULL 
+                    AND p.created_at < CURRENT_TIMESTAMP - interval '6 weeks')
+	     )`);
```
arranfw, line 23:
```suggestion
        AND (
            to_timestamp(p.body->>'userUpdatedAt', 'YYYY-MM-DD') < CURRENT_TIMESTAMP - interval '6 weeks'
	      );
```

server/scripts/get-email-list.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable no-console */
+const { dbClient } = require('../db');
+require('dotenv').config({ path: '../.env' });
+(async function exec() {
+  await dbClient.connect();
+  const res = await dbClient.db.query(`
+    SELECT p.body ->> 'emailAddress' AS email
+    FROM participants p 
+    WHERE 
+        p.body ->> 'emailAddress' NOT IN (
```
ashtonmeuser, line 10: This seems like a kinda convoluted way of only grabbing those participants without `hired` statuses. Would a standard join do the trick? BRB with a suggestion.

arranfw (approved):

server/scripts/get-email-list.js
dbayly-freshworks: I initially went with a standard join, but it doesn't account for the case where a user has multiple `participants_status`, with one of them being 'hired'. It will still join with the non-hired `participants_status`. 

Putting the exclusion in the where clause isn't the most efficient, but it's the simplest way of excluding an email.

arranfw (approved):

Zenen-freshworks (approved):
Logic looks sound to me!

PR #372, opened 2021-05-12
www. is now a valid subdomain for redirection

Zenen-freshworks:
Simplified the logic behind how URL matching is performed. 

Resolving the 'invalid parameter: redirect_uri' involved altering one of the client values in keycloak to a wildcard match, which may need to be specified further in the case of non-local deployments.

We could also use a more intricate regex to match all of the hostnames for dev, test, and prod as well but brain hurt.

### DIFF
--- a/client/src/constants/routes.js
+++ b/client/src/constants/routes.js
@@ -1,7 +1,8 @@
+/* eslint-disable no-useless-escape */
 export default Object.freeze({
   // Hostname
-  ParticipantHostname: 'hcapparticipants.*',
-  EmployerHostname: 'hcapemployers.*',
+  ParticipantHostname: RegExp('^(www\.)?hcapparticipants\..*'),
+  EmployerHostname: RegExp('^(www\.)?hcapemployers\..*'),
 
   // Public routes
   Login: '/login',
--- a/client/src/constants/routes.js
+++ b/client/src/constants/routes.js
@@ -1,7 +1,6 @@
-/* eslint-disable no-useless-escape */
 export default Object.freeze({
   // Hostname
-  ParticipantHostname: RegExp('^(www\.)?hcapparticipants\..*'),
+  ParticipantHostname: RegExp('^(www\\.)?hcapparticipants\\..*'),
   EmployerHostname: RegExp('^(www\\.)?hcapemployers\\..*'),
 
   // Public routes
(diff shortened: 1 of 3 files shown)


client/src/constants/routes.js
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable no-useless-escape */
 export default Object.freeze({
   // Hostname
-  ParticipantHostname: 'hcapparticipants.*',
-  EmployerHostname: 'hcapemployers.*',
+  ParticipantHostname: RegExp('^(www\.)?hcapparticipants\..*'),
+  EmployerHostname: RegExp('^(www\.)?hcapemployers\..*'),
```
Zenen-freshworks, line 5: As it stands, these urls contain a wildcard match at the end, meaning they are valid for any domain provided it starts in the right way? Is that a security issue or are we not concerned with that at this level?

client/src/constants/routes.js
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable no-useless-escape */
```
ashtonmeuser, line 1: I think we should keep this rule. Can be misleading otherwise.

client/src/constants/routes.js
ashtonmeuser, line 5: Because the linting rule is disabled, I'm not sure this will have its intended effect. Double backslash outta do it.
```suggestion
  EmployerHostname: RegExp('^(www\\.)?hcapemployers\\..*'),
```

dbayly-freshworks (approved):

client/src/constants/routes.js
Zenen-freshworks: It triggers on escaping the `.` in our RegExp, I don't see any better way around it.

client/src/constants/routes.js
Zenen-freshworks: It has the correct effect, the linting rule doesn't change code functionality as far as I'm aware. Wouldn't escaping the backslash then match on hostnames that contained the literal backslash char, followed by any character? i.e. `www\Zhcapemployers.*`

arranfw: I believe `'^(www\\.)?hcapemployers\\..*'` evaluates to '^(www\.)?hcapemployers\..*' and seems to have the correct effect (doesn't allow `www*hcap`, just `www.hcap`) without the need to disable the lint rule, if there's a way to accomplish while making the linter happy we should go that direction

ashtonmeuser (approved):

arranfw (approved):

PR #374, opened 2021-05-12
Prettier pre-commit hook

arranfw:
This is just for prettier, ended up running into some weirdness with the linters/lint-staged so I'm going to keep it brief for now. Would love to pick this back up and get the whole suite running pre-commit.

Removed lint-staged from server for streamlining (and I don't think I've seen it activate for me yet?)

### DIFF
--- a/.husky/_/husky.sh
+++ b/.husky/_/husky.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+if [ -z "$husky_skip_init" ]; then
+  debug () {
+    [ "$HUSKY_DEBUG" = "1" ] && echo "husky (debug) - $1"
+  }
+
+  readonly hook_name="$(basename "$0")"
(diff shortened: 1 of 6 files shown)


.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
ashtonmeuser, line 1: Is this introduced by the new husky version?

ashtonmeuser (approved):

.husky/_/husky.sh
ashtonmeuser: Looks like this is the recommended way to define husky config where before we had defined it in package.json.

arranfw: Yeah that's right, it's the result of `npx husky install`

Zenen-freshworks (approved):

PR #375, opened 2021-05-12
speedy gonzales bugfix

Zenen-freshworks:
gotta cast siteID as an integer


ashtonmeuser (approved):

arranfw:
I see `error: invalid input syntax for type integer: "siteId"` locally

PR #377, opened 2021-05-13
Confirm Interest Table + Migration

Zenen-freshworks:


fwkendall:
I remember originally when we were discussing the design for this the plan was to actually use a separate table and delete rows from it to invalidate the OTP. Curious if we can discuss the change here to put it in the participant jsonb? Just wondering the pro/con we ran into 😄

Zenen-freshworks:
Sure! There wasn't much of a discussion. I think that the main point behind the decision was "we're already using the json body for everything participant related so we might as well keep everything in one place". I think that we can quite easily remove the value from the jsonb to invalidate it in the same way. Do you have any concerns about putting it here rather than in a separate table?

fwkendall:
Really just performance of the lookup, keeping in mind we'll have one OTP associated with multiple participant records that the FE is going to be requesting when the link is visited. So we might need an index on it.

Zenen-freshworks:
Okay, I definitely forgot to add an index here so I can patch that up. Should the OTP not be unique for each participant then, and instead correspond to distinct email addresses?

fwkendall:
Yes I think it's one unique email == one unique OTP, not unique to each record

Zenen-freshworks:
So, your suggestion would be to make an OTP table with a column for email address and another for the code? I feel like both should be indexed, right?

arranfw:
I'm in favour of the table solution rather than having to manage duplicate OTPS across PEOIs with the same email. The output of [HCAP-536](https://freshworks.atlassian.net/browse/HCAP-536) should end up 1:1 with records in that table.

Zenen-freshworks:
Alright, I will tackle this tomorrow!

arranfw (approved):

server/migrations/1620932907414_add-otps.js
```diff
@@ -0,0 +1,31 @@
...
+  await dbClient.db.query(
+    `CREATE INDEX IF NOT EXISTS otp_index ON ${collections.CONFIRM_INTEREST}(otp);`
+  );
+
+  await dbClient.db
+    .query(`INSERT INTO ${collections.CONFIRM_INTEREST} (email_address)
+  SELECT DISTINCT p.body ->> 'emailAddress' AS email
+      FROM participants p 
+      WHERE 
+          p.body ->> 'emailAddress' NOT IN (
```
ashtonmeuser, line 20: Seems like we could do this without the `NOT IN` subquery thing. Suggestion incoming.

ashtonmeuser: Can we try something like this?
```
SELECT DISTINCT participants.body->>'emailAddress' AS email
    FROM participants
    JOIN participants_status 
    ON participants.id = participants_status.participant_id
    AND participants_status.status != 'hired'
    AND (
        (participants.body->>'interested')::TEXT = 'yes'  
        OR (participants.body->>'interested')::TEXT IS NULL
    );
```

server/migrations/1620932907414_add-otps.js
```diff
@@ -0,0 +1,27 @@
...
+      DISTINCT p.body ->> 'emailAddress' AS email
+    FROM
+      ${PARTICIPANTS} as p FULL OUTER JOIN
+      ${PARTICIPANTS_STATUS} as ps ON p.id = ps.participant_id
+      AND ps.status != 'hired'
+      AND (
+          (p.body ->> 'interested') :: TEXT = 'yes'
+          OR (p.body ->> 'interested') :: TEXT IS NULL
+      )
+    WHERE p.body ->> 'emailAddress'  IS NOT NULL
```
arranfw, line 24: This query was returning the first row as null so I added this 🤔

arranfw (approved):

dbayly-freshworks (approved):

ashtonmeuser (approved):

PR #383, opened 2021-05-17
HCAP-582 Confirm Interest Endpoint

arranfw:
https://freshworks.atlassian.net/browse/HCAP-582

### DIFF
--- a/client/src/pages/public/ConfirmInterest.js
+++ b/client/src/pages/public/ConfirmInterest.js
@@ -14,6 +14,7 @@ import {
   confirmInterestDefault,
   confirmInterestError,
   confirmInterestSuccess,
+  confirmInterestLoading,
 } from '../../constants/confirmInterestConstants';
 
 const useStyles = makeStyles((theme) => {
@@ -47,7 +48,7 @@ const useStyles = makeStyles((theme) => {
 });
 
 export default (props) => {
-  const [state, setState] = useState(null);
+  const [state, setState] = useState(confirmInterestLoading);
   const classes = useStyles();
   // Get the query string from the url
   // Remove the questionmark from the search
@@ -58,11 +59,13 @@ export default (props) => {
       const res = await fetch(`${API_URL}/api/v1/participants/confirm-interest?id=${query.id}`, {
         method: 'GET',
       });
-      if (res.ok) {
-        setState(confirmInterestDefault);
-      } else {
-        setState(confirmInterestError);
-      }
+      setTimeout(() => {
+        if (res.ok) {
+          setState(confirmInterestDefault);
+        } else {
+          setState(confirmInterestError);
+        }
+      }, 500);
     };
     if (!query.id) {
       props.history.push('/');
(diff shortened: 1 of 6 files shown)


server/services/participants.js
```diff
@@ -125,6 +125,52 @@ const updateParticipant = async (participantInfo) => {
   return participant;
 };
 
+const validateConfirmationId = (id) =>
+  dbClient.db[collections.CONFIRM_INTEREST].findOne({ otp: id });
```
Zenen-freshworks, line 129: Is this necessary? If the OTP is invalid then the relatedParticipants query would be null. Calling it a nitpick as removing this would require catching it elsewhere.

server/server.js
```diff
@@ -183,6 +186,33 @@ app.get(
...
 );
 
+app.get(
+  `${apiBaseUrl}/participants/confirm-interest`,
+  asyncMiddleware(async (req, res) => {
+    const confirmationId = req.query.id;
+    const isValid = uuidValidate(confirmationId) && (await validateConfirmationId(confirmationId));
+
+    if (isValid) {
+      return res.status(isValid ? 200 : 400).send();
```
Zenen-freshworks, line 196: this ternary operator is redundant.
```suggestion
      return res.status(200).send();
```

server/services/participants.js
arranfw: This is the bit that we call to see if we should show the invalid message, are you saying we could just call the other function below to get the same effect?

Zenen-freshworks: It only confirms that the uuid is in the table, when I would have just queried the table and checked afterwards to see if it hit anything.

Zenen-freshworks (dismissed):
looks good, works well! Nitpick redacted

server/services/participants.js
```diff
@@ -125,6 +125,52 @@ const updateParticipant = async (participantInfo) => {
...
+        ],
+        timestamp: now,
+      },
+      ...(participant.body.history ? participant.body.history : []),
+    ],
+  }));
+
+  await Promise.all(
+    updatedParticipantFields.map((fields) => {
+      const result = dbClient.db[collections.PARTICIPANTS].updateDoc({ id: fields.id }, fields);
```
ashtonmeuser, line 164: Hmmm... `fields` does include `id` which means that `id` (PK!!!) is being updated for each participant.

ashtonmeuser, line 164: Ah ha! This outta do it.
```suggestion
    updatedParticipantFields.map(({id, ...fields}) => {
      const result = dbClient.db[collections.PARTICIPANTS].updateDoc({ id }, fields);
```

client/src/pages/public/ConfirmInterest.js
```diff
@@ -46,19 +47,43 @@ const useStyles = makeStyles((theme) => {
...
+      await fetch(`${API_URL}/api/v1/participants/confirm-interest?id=${query.id}`, {
+        method: 'POST',
+      });
+      setState(confirmInterestSuccess);
+    } catch (error) {
       setState(confirmInterestError);
     }
   };
+
+  if (!state) return null;
```
dbayly-freshworks, line 85: Total nitpick, but it might be nice to show a loading icon or something here.

server/services/participants.js
Zenen-freshworks: Ah, so the `fields` object contained an `id` that was updating the pk of the entry, and {id, ...fields} expands the object to pull out the id as a separate value from the rest of `fields`?

dbayly-freshworks (dismissed):

dbayly-freshworks (approved):

ashtonmeuser (approved):

PR #388, opened 2021-05-21
HCAP-587 - add jsx a11y eslint plugin 2/3

arranfw:
+ fix warning about lint/react version

Removed vestigial onClick from logo

### DIFF
--- a/client/src/components/generic/Header.js
+++ b/client/src/components/generic/Header.js
@@ -76,8 +76,6 @@ export const Header = ({ hideEmployers = false }) => {
   const classes = useStyles();
   const [keycloak] = useKeycloak();
 
-  const handleLogoClick = () => history.push(Routes.EmployerForm);
-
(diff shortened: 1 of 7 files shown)


client/src/components/generic/Header.js
```diff
@@ -76,8 +76,6 @@ export const Header = ({ hideEmployers = false }) => {
   const classes = useStyles();
   const [keycloak] = useKeycloak();
 
-  const handleLogoClick = () => history.push(Routes.EmployerForm);
```
ashtonmeuser, line 79: What is the intent here?

arranfw: This was the hacky way of handling the logo before, we added a link around the image so this is unnecessary and impossible to actually activate

arranfw: Also it was a failure for the jsx-a11y ruleset

PR #389, opened 2021-05-21
HCAP-587 - Add lint-staged 3/3

arranfw:


ashtonmeuser (approved):

package.json
```diff
@@ -18,12 +17,16 @@
     "dotenv": "8.2.0",
     "eslint": "^7.14.0",
     "husky": "6.0.0",
+    "lint-staged": "^11.0.0",
```
dbayly-freshworks, line 20: Maybe lock down some of these packages.

dbayly-freshworks (approved):
One nitpick, otherwise looks good.

package.json
arranfw: Totally

PR #390, opened 2021-05-21
confirm interest migration: revert to previous query

arranfw:
looks like new edition of this query didn't give the exact results we were after, going to ignore the performance hit as this is only going to be run once.


dbayly-freshworks (approved):
Looks good to me!

ashtonmeuser (approved):

PR #397, opened 2021-05-27
Hcap 520 filter refresh

dbayly-freshworks:
Fixed issue where after 5 minutes, the sites component would re-render, and clear the filter. 

Also contains some missing requirements from HCAP-527

### DIFF
--- a/server/keycloak.js
+++ b/server/keycloak.js
@@ -89,6 +89,7 @@ class Keycloak {
   }
 
   async authenticateServiceAccount() {
+    console.log('Type 1')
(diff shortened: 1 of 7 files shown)


server/keycloak.js
```diff
@@ -89,6 +89,7 @@ class Keycloak {
   }
 
   async authenticateServiceAccount() {
+    console.log('Type 1')
```
arranfw, line 92: 😃

dbayly-freshworks:
This PR needs more work and will not be included in the current release

PR #401, opened 2021-06-01
Lock Versions + Rerun npm i

arranfw:
- Locks package versions
- Updates package-lock files (they were out of step)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -32,7 +32,7 @@
         "yup": "0.29.3"
       },
       "devDependencies": {
-        "eslint-plugin-jsx-a11y": "^6.4.1",
+        "eslint-plugin-jsx-a11y": "6.4.1",
         "react-scripts": "4.0.0"
       }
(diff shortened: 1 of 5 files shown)


dbayly-freshworks (approved):
I get locking down the packages in package.json, but why lockdown packages inside the package-lock?

arranfw:
> I get locking down the packages in package.json, but why lockdown packages inside the package-lock?

That's the auto-output from npm, I think I missed it when finishing up the last prettier/eslint PR

dbayly-freshworks:
> > I get locking down the packages in package.json, but why lockdown packages inside the package-lock?
> 
> That's the auto-output from npm, I think I missed it when finishing up the last prettier/eslint PR, if you have a moment could you pull this branch and verify you don't get a diff after running `npm run update-all`?

I just did this and there was no diff. 👍

fwpushan (approved):
👍

PR #406, opened 2021-06-02
Confirm interest Logic update

dbayly-freshworks:
* fixed issue where the confirm-interest submit route was checking only records that had not been updated in the past 6 weeks. 
* Fixed issue where the change to the interested column is changed without logging the change. 
* Will now fail if no otp is deleted or no rows were updated. 

https://freshworks.atlassian.net/browse/HCAP-610

### DIFF
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -1,4 +1,3 @@
-const dayjs = require('dayjs');
 const readXlsxFile = require('node-xlsx').default;
 const { validate, ParticipantBatchSchema, isBooleanValue } = require('../validation.js');
 const { dbClient, collections } = require('../db');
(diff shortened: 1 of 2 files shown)


server/services/participants.js
```diff
@@ -141,9 +140,7 @@ const confirmParticipantInterest = async (id) => {
     })
     .find({
       'body.interested IS DISTINCT FROM': 'withdrawn', // "IS DISTINCT FROM" = "!=" but includes null
-      'body.userUpdatedAt::TIMESTAMP <': sixWeeksAgo,
```
arranfw, line 144: I would think we want to keep this check, why is it removed?

dbayly-freshworks: The updated at column can be changed without the users action. It is a terrible way of measuring interest. 
It leads to edge cases in testing as well.  Having a valid OTP should be evidence enough.

arranfw (approved):

fwpushan (approved):
👍

PR #407, opened 2021-06-03
Mongo Transferral Process

Zenen-freshworks:
Okay, phew! Finally got this whole process documented and tested on dev.

The new mongo instance has 18953 records, which seems like a pretty good number to me though I didn't confirm whether or not that is exactly what we're looking for. Looking forward to trying this out on test!

### DIFF
--- a/mongodb-recreate-and-restore.md
+++ b/mongodb-recreate-and-restore.md
@@ -64,7 +64,8 @@ Storage -> Persistent Volume Claims.
 With the new space freed up, we can begin to spin up the new Mongo deployment
 with helm. Run the following command from the root directory:
 
-$ helm install hcap-mongo bitnami/mongodb -f openshift/mongo.helm.yml
+$ helm install hcap-mongo bitnami/mongodb -f openshift/mongo.helm.yml --set
+auth.username <desired username> --set auth.database <desired db name>
 
 Once it completes, we will have 2 new mongo pods to work with and we can shift
 logging duties over there with the server-new-mongo deployment config.
--- a/mongodb-recreate-and-restore.md
+++ b/mongodb-recreate-and-restore.md
@@ -114,8 +114,8 @@ $ oc port-forward service/hcap-mongo-headless 27017
 With the above command running, we can run the mongorestore command to load our
 backups onto the new deployment.
 
-$ mongorestore --authenticationDatabase hcap -d hcap -u hcap -p LlHeh2drDp ocp3-mongo-dump/hcap
-$ mongorestore --authenticationDatabase hcap -d hcap -u hcap -p LlHeh2drDp ocp4-mongo-dump/hcap
+$ mongorestore --authenticationDatabase hcap -d <desired db name> -u <desired username> -p <generated user password> ocp3-mongo-dump/hcap
+$ mongorestore --authenticationDatabase hcap -d <desired db name> -u <desired username> -p <generated user password> ocp4-mongo-dump/hcap
 
 Once these commands have completed, the restoration is complete.
 Congratulations!
--- a/mongodb-recreate-and-restore.md
+++ b/mongodb-recreate-and-restore.md
@@ -64,8 +64,7 @@ Storage -> Persistent Volume Claims.
 With the new space freed up, we can begin to spin up the new Mongo deployment
 with helm. Run the following command from the root directory:
 
-$ helm install hcap-mongo bitnami/mongodb -f openshift/mongo.helm.yml --set
-auth.username <desired username> --set auth.database <desired db name>
+$ helm install hcap-mongo bitnami/mongodb -f openshift/mongo.helm.yml --set auth.username=<desired username> --set auth.database=<desired db name>
 
 Once it completes, we will have 2 new mongo pods to work with and we can shift
 logging duties over there with the server-new-mongo deployment config.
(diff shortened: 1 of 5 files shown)


openshift/server.dc.yml
```diff
@@ -119,24 +119,18 @@ objects:
...
-                  value: ${APP_NAME}-mongodb
+                  value: ${APP_NAME}-mongo-headless
                 - name: MONGO_PORT
                   value: '27017'
                 - name: MONGO_DB
-                  valueFrom:
-                    secretKeyRef:
-                      key: database
-                      name: ${APP_NAME}-mongodb
+                  value: hcap
```
arranfw, line 126: Is it worth using the app name variable here?

```suggestion
                  value: ${APP_NAME}
```

openshift/server.dc.yml
```diff
@@ -119,24 +119,18 @@ objects:
                       key: sa-password
                       name: ${APP_NAME}-keycloak
                 - name: MONGO_HOST
-                  value: ${APP_NAME}-mongodb
+                  value: ${APP_NAME}-mongo-headless
```
arranfw, line 122: Do we have more than 1 mongo name? Why are we specifying headless in the name?

openshift/mongo.helm.yml
```diff
@@ -0,0 +1,25 @@
+fullnameOverride: hcap-mongo
+persistence:
+  size: 4Gi
+
+podSecurityContext:
+  enabled: false
+  fsGroup: 1008111001
```
arranfw, line 7: Are these ids or do they have a specific meaning?

mongodb-recreate-and-restore.txt
```diff
@@ -0,0 +1,125 @@
...
+
+Once helm is installed, we need to log into OCP4 with the creds provided
+by the web console, and then add the bitnami repo with the following command:
+
+$ oc login ...
+$ helm repo add bitnami https://charts.bitnami.com/bitnami
+
+Now that bitnami is registered, we can use their helm chart to build a new
+mongo deployment on Openshift. Make space by scaling down the existing
+hcap-mongodb StatefulSet to one pod, and then delete the associated PVC in
```
arranfw, line 61: Do we need to backup data before this step?

arranfw (approved):
Awesome work, this is so great to have done

My understanding
1. Backup OCP3
2. Scale down old Mongo
3. Create new Mongo StatefulSet 
5. Point server to new Mongo
6. Backup OCP4 
7. Restore OCP3 and OCP4 to new Mongo

Seems like a great plan and shouldn't result in any log gaps or downtime, nicely done!

Quick questions:
- How does this relate to @ashtonmeuser 's open PR renaming Mongo?
- Does this impact the local dev environment at all?

openshift/server.dc.yml
Zenen-freshworks: hcap-mongo-headless is the name of the service that helm spins up.

openshift/mongo.helm.yml
Zenen-freshworks: The ids can be removed, I believe. We're not enabling them anyways.

mongodb-recreate-and-restore.txt
Zenen-freshworks: I would prefer to make our backups one we've stopped using this statefulset for logging to avoid data loss. As far as I know, every PVC will hold the same data - we can consult the rest of the team to be sure.

arranfw: Ahh roger, do we run out of space if we don't take the PVC down?

mongodb-recreate-and-restore.txt
```diff
@@ -0,0 +1,125 @@
+This text document is for outlining the process necessary on a given
```
fwpushan, line 1: Please rename file as md

openshift/mongo.helm.yml
```diff
@@ -0,0 +1,21 @@
...
+arbiter:
+  podSecurityContext:
+    enabled: false
+
+architecture: replicaset
+image:
+  registry: "public.ecr.aws"
+
+auth:
+  username: hcap
```
fwpushan, line 20: This should be connected to any secret. This is a public repository, this is a big security issue. or at-least connect these auth attribute with input parameter

openshift/server.dc.yml
```diff
@@ -119,24 +119,18 @@ objects:
...
                 - name: MONGO_PORT
                   value: '27017'
                 - name: MONGO_DB
-                  valueFrom:
-                    secretKeyRef:
-                      key: database
-                      name: ${APP_NAME}-mongodb
+                  value: ${APP_NAME}
```
fwpushan, line 126: This should be connected to any secret. This is a public repository, this is a big security issue. or at-least connect these auth attribute with input parameter

fwpushan (changes requested):

mongodb-recreate-and-restore.txt
Zenen-freshworks: Yep, not having enough storage space was what got us into this situation in the first place. This should resolve it though.

mongodb-recreate-and-restore.md
```diff
@@ -0,0 +1,128 @@
...
+
+Now that we have a service to direct traffic, we can use that to forward the
+port.
+
+$ oc port-forward service/hcap-mongo-headless 27017
+
+With the above command running, we can run the mongorestore command to load our
+backups onto the new deployment.
+
+$ mongorestore --authenticationDatabase hcap -d hcap -u hcap -p LlHeh2drDp ocp3-mongo-dump/hcap
```
fwpushan, line 117: Is this a dummy password?

fwpushan (commented):
Please check comment regarding one more security concern.

dbayly-freshworks (approved):

fwpushan (approved):

PR #408, opened 2021-06-04
HCAP-661 Global auth provider 

arranfw:
https://freshworks.atlassian.net/browse/HCAP-661

Rather than focus on the loading state, I'm addressing the fact we reload user so often. This is the action that will occur on app load, next step is to remove the user request from the individual components.

This PR only handles the global state, which fires a request on app load to fill the auth context with the user object, apart from the extra request and the failure handling, there isn't any impact on other pages


fwpushan (approved):
👍

Zenen-freshworks (approved):

PR #414, opened 2021-06-08
[openshift][workflow][Makefile][HCAP-#585] Deploy config changes

fwpushan:
----
1. Create make alias for testing deploy config change
2. Create make alias for process and apply deployment config change
3. Create new github workflow config.

Ticket: [HCAP-585](https://freshworks.atlassian.net/browse/HCAP-585)

- [x] All secrets configured.
- [x] Passing all checks

### DIFF
--- a/.github/workflows/config-prod.yml
+++ b/.github/workflows/config-prod.yml
@@ -17,17 +17,23 @@ jobs:
     runs-on: 'ubuntu-latest'
     environment:
       name: prod
-    if: ${{ secrets.SA_TOKEN }}
     env:
       OS_NAMESPACE_SUFFIX: prod
       AUTH_TOKEN: ${{ secrets.SA_TOKEN }}
     steps:
+      - name: Print Status
+        if: ${{ secrets.SA_TOKEN }}
+        run: |
+          echo OC CLI Version: $(oc version)
+          echo Service Account token configured
       - name: checkout
         uses: actions/checkout@v2
       - name: Login
+        if: ${{ secrets.SA_TOKEN }}
         run: |
           oc login --token="$AUTH_TOKEN" --server="$CLUSTER"
       - name: Test and Deploy
+        if: ${{ secrets.SA_TOKEN }}
         run: |
           cd "$GITHUB_WORKSPACE"
           make server-config-test
--- a/.github/workflows/config-prod.yml
+++ b/.github/workflows/config-prod.yml
@@ -7,8 +7,6 @@ on:
       - '.github/**'
     tags:
       - prod
-    branches:
-      - hcap-585-deploy-config-change
 
 env:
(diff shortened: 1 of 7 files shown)


Zenen-freshworks (dismissed):

will-freshworks (changes requested):
The code looks good, but I am wondering why we have a single `config` tag for pushing to all three environments? Seems risky. Why is the config not bundled along with the rest of the `prod`/`test`/`dev` code?

I am probably missing something and need more context.

fwpushan:
> The code looks good, but I am wondering why we have a single `config` tag for pushing to all three environments? Seems risky. Why is the config not bundled along with the rest of the `prod`/`test`/`dev` code?
> 
> I am probably missing something and need more context.

Initially. I thought about putting config update as part of dev/test/prod general deployment.
But as per current process, team separate config change from build promotion.
Currently, config change is manual process, dev manually config the env before promote.
Deployment has image trigger (unlike Court), so tagging image automatically start deployment to respective env
So, introducing config update in same  flow will trigger rollout in respective env twice (redundant) one for config change and one for image tagging.
For above reason, I also kept it separate from general image promotion.
Now, what is your recommendation?

.github/workflows/config.yml
```diff
@@ -0,0 +1,66 @@
...
+      OS_NAMESPACE_SUFFIX: test
+    needs: test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Apply changes
+        run: |
+          oc login --token="$AUTH_TOKEN" --server="$CLUSTER"
+          cd "$GITHUB_WORKSPACE"
+          make server-config
```
vyasworks, line 66: Is reapplying `server-config` going to trigger a deployment config? I thought DC resource would change in K8S and wait for the actual trigger configured in DC or forced via `server-deploy`  my understanding was different. I might be wrong :)

fwpushan: Now using same tag to update config. Please review

Zenen-freshworks (dismissed):
looks good to me

dbayly-freshworks (approved):

.github/workflows/config-prod.yml
```diff
@@ -0,0 +1,41 @@
+name: Test and Deploy Config on Prod
+
+on:
+  push:
+    paths:
+      - 'openshift/*.yml'
```
will-freshworks, line 6: I think `openshift/**` is safer

will-freshworks (changes requested):
Needs the change we discussed so dev and test can be deployed separately.

will-freshworks (approved):

Zenen-freshworks (approved):

dbayly-freshworks (approved):

PR #417, opened 2021-06-14
Fixed filter

dbayly-freshworks:
It turns out that for some reason, up doesn't display validation errors on fields that are undefined. 
Before the active participant is assigned, if the postalCode is undefined, I set it to the empty string so Yup has a value to check.


Zenen-freshworks (dismissed):

Zenen-freshworks (approved):

fwpushan (approved):
👍

PR #423, opened 2021-06-17
participant constants + context (HCAP-660 #2) 

arranfw:
Context for the participant table/view. This will eventually hold all state for the component. For now it's starting with the tab/column state.


fwpushan (approved):
Very good starting

PR #431, opened 2021-06-24
Hcap 412 add withdraw to site details view

dbayly-freshworks:
Added missing AC for 412. 

Health Authorities can now withdraw candidates from the Site details page

Screen Shot 2021-06-24 at 2 47 55 PM
Screen Shot 2021-06-24 at 2 47 47 PM

### DIFF
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -3,24 +3,48 @@ import _orderBy from 'lodash/orderBy';
 import Grid from '@material-ui/core/Grid';
 import { Box } from '@material-ui/core';
 import store from 'store';
-import { Table } from '../../components/generic';
-import { API_URL } from '../../constants';
+import { Table, Button, Dialog } from '../../components/generic';
+import { checkPermissions , getDialogTitle} from '../../utils';
+import { AuthContext } from '../../providers';
+import {
+  ToastStatus,
+  API_URL,
+  makeToasts,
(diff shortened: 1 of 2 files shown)


fwpushan (approved):
LGTM

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -31,6 +49,48 @@ export default ({ siteId }) => {
 
   const sort = (array) => _orderBy(array, [orderBy, 'operatorName'], [order]);
 
+  const forceReload = async () => {
```
Zenen-freshworks, line 52: is this code taken from the useEffect on line 94? I think we would be better off encapsulating it.

Zenen-freshworks (commented):
Just one thing I'd like to see changed, otherwise it's looking like a solid PR

client/src/pages/private/SiteParticipantsTable.js
dbayly-freshworks: I know this is an anti pattern, but we're kinda forced into it. We need to be able to reload the form after we submit. 
If we just have it declared within the useEffect, then we can't call it elsewhere. If we declare it outside, it becomes a dependancy of the useEffect and breaks. 
I don't know if there is a more elegant way to handle this issue. 

We use this pattern exactly in the participants table.

Zenen-freshworks: Ah, fair enough! I forgot about the whole dependency thing.

PR #436, opened 2021-06-28
Hcap 696 final showdown

dbayly-freshworks:
Fixed HCAP-715 - Issue where the incorrect key and shape were used to check if a user was MoH in the SiteParticipants table. 

Improved the function of  HCAP-695 
- When navigating between tabs on the participant table, if you are on a page with a page number that is greater than the page total for the tab you are navigating to, you will be shown the first  page of that tab. The number will still be wrong as I haven't figured out why the component isn't updating. This also causes the page switch feature to respond improperly ( The component is set to the wrong page, so it wont update correctly until the user uses the controls to update it to the correct page. 

Fixed HCAP-696 

When a MoH withdraws a hired participant, their participant statuses were not being updated. 

Now when a MoH worker withdraws a user, a new participant status is added for each EOI. The previous status is stored in that data field and if that user's interest is switched back to 'yes', the previous status is restored. A record of this switch will exist in the database. 

Updated the Participants Table to show archived users to MoH

### DIFF
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -6,7 +6,13 @@ import store from 'store';
 import { Table, Button, Dialog } from '../../components/generic';
 import { getDialogTitle } from '../../utils';
 import { AuthContext } from '../../providers';
-import { ToastStatus, API_URL, makeToasts, ArchiveHiredParticipantSchema } from '../../constants';
+import {
+  ToastStatus,
+  API_URL,
+  makeToasts,
+  ArchiveHiredParticipantSchema,
+  participantStatus,
+} from '../../constants';
 import { ArchiveHiredParticipantForm } from '../../components/modal-forms';
 import { useToast } from '../../hooks';
 import moment from 'moment';
@@ -152,7 +158,7 @@ export default ({ siteId, onArchiveParticipantAction }) => {
       const index = rows.findIndex((row) => row.participantId === participantId);
       const { participantName } = rows[index];
       const toasts = makeToasts(participantName, '');
-      openToast(toasts['archived']);
+      openToast(toasts[participantStatus.ARCHIVED]);
       setActionMenuParticipant(null);
       setActiveModalForm(null);
     } else {
@@ -243,7 +249,7 @@ export default ({ siteId, onArchiveParticipantAction }) => {
             }}
             validationSchema={ArchiveHiredParticipantSchema}
             onSubmit={async (values) => {
-              await handleEngage(actionMenuParticipant.id, 'archived', values);
+              await handleEngage(actionMenuParticipant.id, participantStatus.ARCHIVED, values);
               if (onArchiveParticipantAction) {
                 onArchiveParticipantAction();
               } else {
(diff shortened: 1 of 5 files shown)


server/services/participants.js
```diff
@@ -133,6 +134,20 @@ const updateParticipant = async (participantInfo) => {
     },
     { history: participantInfo.history, userUpdatedAt: new Date().toJSON() }
   );
+  if(changes.interested !== undefined){
```
fwkendall, line 137: nitpick - the spacing here is inconsistent. I thought our linting should catch this - any idea if it didn't run / isn't looking for this?

dbayly-freshworks: For whatever reason my linting doesn't run on commit and I forgot to run the formatter. Good catch!

server/services/participants-helper.js
```diff
@@ -56,6 +56,34 @@ const addDistanceToParticipantFields = (raw, siteDistanceJoin) =>
     }),
   }));
 
+// Find the previous status for each org, create a copy of it 
+const  revertToOldStatus = async ( participantStatuses,setParticipantStatus)=>{
```
fwkendall, line 60: I think this is actually a change in the existing behaviour, but that's fine as long as we document it (I think before the participant would always go back to "prospecting" - my understanding here is they would go back to any prior status at all, although I believe the button is disabled for withdrawn archived participants so going back to "hired" would never be possible)

server/services/participants-helper.js
```diff
@@ -56,6 +56,34 @@ const addDistanceToParticipantFields = (raw, siteDistanceJoin) =>
...
+}
+// Withdraw the participant 
+const insertWithdrawalParticipantStatus = async (participantStatuses,setParticipantStatus) =>{ 
+  participantStatuses.forEach( async (status)=>{
+    // Prevent locking the user into a loop of archived statuses
+    if(status.status && status.status ==='archived'){
+      return;
+    }else{
+      await setParticipantStatus(status.employer_id,status.participant_id, 'archived',{
+        final_status: 'Withdrawn by MoH',
```
fwkendall, line 78: Should we make this more generic? Thinking we'll probably reuse this method when participants can withdraw their own interest. Maybe just "Withdrawn from HCAP"?

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -31,9 +31,9 @@ export default ({ siteId, onArchiveParticipantAction }) => {
   const { auth } = AuthContext.useAuth();
   const { openToast } = useToast();
   const roles = auth.user?.roles || [];
-  const isHA = checkPermissions(roles, ['health_authority']);
+  const isHA = roles.includes('health_authority');
   if (!isHA) {
-    columns = columns.filter((col) => col.id !== 'archived');
+    columns = columns.filter((col) => col.id !== 'archive');
```
fwkendall, line 36: gross, we really need a constants file somewhere to prevent this kind of thing

fwpushan (commented):
LGTM for me. Please check some minor comments. And also check some action is failing.

Zenen-freshworks (commented):
Yep, looks like we already have good suggestions on this PR - I have nothing to add.

server/services/participants-helper.js
dbayly-freshworks: It's pretty clear this was a case that wasn't thought of when it was first implemented. 

The previous function was to set the status of the `participant` to withdrawn but not update any `participants_status`, hence the bug.  I added this part to ensure that the ability to re-engage a user after they're withdrawn by a MoH user was kept.

This all might warrant further discussion.

fwkendall: Discussed offline - MOH undoing a withdrawal is not a real use case. For now we are documenting this behaviour but it should be disabled in future.

fwkendall (approved):

PR #441, opened 2021-07-06
HCAP-512 BC Services card 

dbayly-freshworks:
Screen Shot 2021-07-06 at 8 28 16 AM

PR is ready to go 

We may need to chat about the link to the BCSC virtual card set up. The link in the designs goes nowhere. The one I added is the result of a quick google search.

### DIFF
--- a/client/src/pages/public/ParticipantLogin.js
+++ b/client/src/pages/public/ParticipantLogin.js
@@ -1,0 +1,88 @@
+import React from 'react';
+import Grid from '@material-ui/core/Grid';
+import Typography from '@material-ui/core/Typography';
+import { Page, Button  } from '../../components/generic';
+import Box from '@material-ui/core/Box';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles((theme) => ({
+  button:{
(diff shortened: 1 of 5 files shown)


client/src/pages/public/ParticipantLogin.js
```diff
@@ -0,0 +1,124 @@
...
+    height: 'calc(100vh - 68px)',
+  },
+}));
+
+export default () => {
+  const classes = useStyles();
+  const [keycloak] = useKeycloak();
+  const history = useHistory();
+  const redirectToLogin = () => {
+    store.set('REDIRECT', true);
```
Zenen-freshworks, line 39: How does setting REDIRECT work here? I would expect it to hold a url rather than a boolean value, but I've also never used it before.

dbayly-freshworks: I copied this over from Login.js, but it looks like it doesn't do anything here. I will remove it.

fwkendall (approved):

fwkendall (approved):

Zenen-freshworks (approved):
Yep, this looks good to me

PR #444, opened 2021-07-08
[HCAP-517][HCAP-738] Linking participants with user after first time login

fwpushan:
---
1. Fix bug
2. Improve efficiency
3. Add more test items
---

Ticket: [HCAP-517](https://freshworks.atlassian.net/browse/HCAP-517) | [HCAP-738](https://freshworks.atlassian.net/browse/HCAP-738)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,4 +174,4 @@ db-mongo-tunnel:
 
 db-postgres-tunnel:
 	@oc project $(TARGET_NAMESPACE)
-	@oc port-forward $(APP_NAME)-patroni-0 5432
+	@oc port-forward svc/$(APP_NAME)-patroni 5432
--- a/Makefile
+++ b/Makefile
@@ -174,4 +174,8 @@ db-mongo-tunnel:
 
 db-postgres-tunnel:
 	@oc project $(TARGET_NAMESPACE)
+	@oc port-forward $(APP_NAME)-patroni-0 5432
+
+db-postgres-rw-tunnel:
+	@oc project $(TARGET_NAMESPACE)
 	@oc port-forward svc/$(APP_NAME)-patroni 5432
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -466,8 +466,8 @@ const makeParticipant = async (participantJson) => {
   return res;
 };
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -174,4 +174,4 @@ db-mongo-tunnel:
 
 db-postgres-tunnel:
 	@oc project $(TARGET_NAMESPACE)
-	@oc port-forward $(APP_NAME)-patroni-0 5432
+	@oc port-forward svc/$(APP_NAME)-patroni 5432
```
fwkendall, line 177: What does adding this do?

For running reports on prod I usually deliberately point at a readonly replica, so want to make sure I understand this in case I need to do that in future :)

fwpushan: @fwkendall , This will pointing to service and through service directly to leader pod. But 100% it will be pointing to RW replica pod. I do it because I want to update some record in dev/db to test certain things .... but if this is alarming for prod, I can create one more alias.

fwkendall: Hmm the main thing is I'll need to update my instructions for the operations team on how to interact with prod.

Maybe two aliases makes sense - keep this one pointing to 0, so it provides more specificity, and create a new one db-postgres-service-tunnel with svc? For dev/test that's absolutely better in all case 🙌

fwpushan: OK

fwpushan: DONE

Zenen-freshworks (approved):

server/services/participants.js
```diff
@@ -466,8 +466,8 @@ const makeParticipant = async (participantJson) => {
   return res;
 };
 
-const createParticipantUserMap = async (userId, email) => {
-  const participants = await dbClient.db[collections.PARTICIPANTS]
+const createParticipantUserMap = async (userId, email, tnx) => {
```
dbayly-freshworks, line 469: this is a bit of a nit, but we might want a more descriptive variable name than tnx. I have no idea what it stands for.

dbayly-freshworks (approved):

server/services/participants.js
fwpushan: transaction, sorry for this, I will update this in my next PR, I'm currently working on same pice of code

PR #452, opened 2021-07-19
Fixed validation on date picker

dbayly-freshworks:
Updated front end validation for the archive candidate date picker
Screen Shot 2021-07-19 at 12 30 43 PM
Screen Shot 2021-07-19 at 12 30 33 PM
Screen Shot 2021-07-19 at 12 30 20 PM

### DIFF
--- a/client/src/constants/validation.js
+++ b/client/src/constants/validation.js
@@ -1,6 +1,10 @@
 import * as yup from 'yup';
 import mapValues from 'lodash/mapValues';
-import { archiveReasonOptions, archiveStatusOptions } from './archiveParticipantsConstants';
+import {
+  archiveReasonOptions,
+  archiveStatusOptions,
+  maxDateString,
+} from './archiveParticipantsConstants';
 
 const healthRegions = ['Interior', 'Fraser', 'Vancouver Coastal', 'Vancouver Island', 'Northern'];
 
@@ -23,7 +27,7 @@ const validateDateIsInThePast = (s) => {
 };
 const validateDateIsReasonable = (d) => {
   try {
-    return Date.parse(d) >= Date.parse('1899/12/31');
+    return Date.parse(d) >= Date.parse(maxDateString);
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


client/src/constants/validation.js
```diff
@@ -18,8 +18,15 @@ const validatePastDateString = (s) => {
...
   return Date.parse(s) <= new Date();
 };
-const validateDateIsInThePast = (d) => {
-  return Date.parse(d) <= new Date();
+const validateDateIsInThePast = (s) => {
+  return Date.parse(s) <= new Date();
+};
+const validateDateIsReasonable = (d) => {
+  try {
+    return Date.parse(d) >= Date.parse('1899/12/31');
```
arranfw, line 26: Nit: It would be nice to store this date as a constant and then it can be used here for comparison as well as in the error message so it doesn't need to be changed in multiple places in the future

(pretty unlikely this date will change though)

docker-compose.dev.yml
```diff
@@ -27,7 +27,7 @@ services:
       context: ./server
       dockerfile: Dockerfile.dev
       args:
-        VERSION: ${LAST_COMMIT}
+        VERSION: '1'
```
arranfw, line 30: Is this supposed to be in this PR?

arranfw (approved):

docker-compose.dev.yml
dbayly-freshworks: As of this morning, whenever I try to build this locally, it will fail unless I have a hardcoded string for the Version. I'm not sure whats going on here. I'll remove it for now.

client/src/constants/validation.js
fwkendall: Do we have a tech debt ticket logged for constants? Let's get this tracked!

fwkendall (dismissed):

arranfw (approved):

fwkendall (approved):

client/src/constants/validation.js
dbayly-freshworks: Tech debt tracked in HCAP 762

PR #455, opened 2021-07-20
HCAP-761 | Add required validation for PEOI update

fwpushan:
-----------------------------------------------
1. Add required validation for PEOI update
2. Fix Docker(v20.*) env issue with `LAST_COMMIT`. Change to `COMMIT_SHA`

Ticket: [HCAP-761](https://freshworks.atlassian.net/browse/HCAP-761)

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -27,7 +27,7 @@ services:
       context: ./server
       dockerfile: Dockerfile.dev
       args:
-        VERSION: ${LAST_COMMIT}
+        VERSION: ${COMMIT_SHA}
     environment:
       - POSTGRES_HOST=${POSTGRES_HOST}
       - POSTGRES_DB=${POSTGRES_DB}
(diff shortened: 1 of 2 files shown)


Zenen-freshworks (approved):

docker-compose.dev.yml
```diff
@@ -27,7 +27,7 @@ services:
       context: ./server
       dockerfile: Dockerfile.dev
       args:
-        VERSION: ${LAST_COMMIT}
+        VERSION: ${COMMIT_SHA}
```
Zenen-freshworks, line 30: Does this work as a long-term solution to the Docker Engine error?

fwkendall (approved):

docker-compose.dev.yml
dbayly-freshworks: I think it might have too. I have a theory that a length limit has been applied to text fields, and the `LAST_COMMIT` variable is too long. `COMMIT_SHA` is shorter and should be unique.

PR #461, opened 2021-07-21
HCAP-774 | Fix base image pull limit issue.

fwpushan:
Switch to **registry.access.redhat.com/ubi8/nodejs-14:1** as base image to build client/server.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,26 +97,6 @@ jobs:
         with:
           ref: 'dev'
           fetch-depth: 0
-
-      - name: Cache OpenShift CLI
-        id: cache-oc
-        uses: actions/cache@v2
-        with:
-          path: /usr/local/bin/oc
-          key: ${{ runner.os }}-oc
-
-      - name: Install OpenShift CLI
-        if: steps.cache-oc.outputs.cache-hit != 'true'
-        run: |
-          OC_VERSION=3.11.0
-          sudo apt-get update
-          sudo apt-get -y install wget
-          wget --quiet -O oc.tar.gz "https://github.com/openshift/origin/releases/download/v${OC_VERSION}/openshift-origin-client-tools-v${OC_VERSION}-0cbc58b-linux-64bit.tar.gz"
-          FILE=$(tar -tf oc.tar.gz | grep '/oc$')
-          tar -zxf oc.tar.gz "$FILE"
-          sudo mv "$FILE" /usr/local/bin/oc
-          rm -rf oc.tar.gz openshift-origin-client-tools-v*
-
       - name: Build & Deploy
         run: |
           cd "$GITHUB_WORKSPACE"
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,13 @@ FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS client
 
 # Build client
 # RUN apk add --no-cache git python g++ make
+USER root
 WORKDIR /opt/app-root/src/app/client
 COPY client/package*.json ./
 RUN npm set progress=false && npm ci --no-cache
 COPY client/. .
 RUN INLINE_RUNTIME_CHUNK=false npm run build
-
+USER 1001
 # Server
 FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS server
 
@@ -19,12 +20,13 @@ ENV NODE_ENV production
 
 # Configure server
 # RUN apk add --no-cache git
+USER root
 COPY --from=client /opt/app-root/src/app/client/build /opt/app-root/src/app/client/build/.
 WORKDIR /opt/app-root/src/app/server
 COPY server/package*.json ./
 RUN npm set progress=false && npm ci --no-cache
 COPY server/. .
-
+USER 1001
 # Run app
 EXPOSE 8080
 CMD [ "npm", "run", "start" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,9 @@ COPY client/package*.json ./
 RUN npm set progress=false && npm ci --no-cache
 COPY client/. .
 RUN INLINE_RUNTIME_CHUNK=false npm run build
+
 # Server
 FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS server
-
 # Static env vars
 ARG VERSION
 ENV VERSION $VERSION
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,28 +1,32 @@
 # Client
-FROM public.ecr.aws/bitnami/node:14.17.3 AS client
+FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS client
```
arranfw, line 2: Does using this image mean we don't need to download the CLI in the build step?

arranfw (dismissed):

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
-WORKDIR /client
+# RUN apk add --no-cache git python g++ make
+USER root
+WORKDIR /opt/app-root/src/app/client
 COPY client/package*.json ./
 RUN npm set progress=false && npm ci --no-cache
 COPY client/. .
 RUN INLINE_RUNTIME_CHUNK=false npm run build
-
+USER 1001
```
arranfw, line 12: I haven't seen this before do you have a link to what this means?

.github/workflows/build.yml
```diff
@@ -93,26 +93,6 @@ jobs:
         with:
           ref: 'dev'
           fetch-depth: 0
-
-      - name: Cache OpenShift CLI
```
vyasworks, line 97: doesnt `make server-build` and `make server-deploy` need the OC CLI?

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
 # Static env vars
 ARG VERSION
 ENV VERSION $VERSION
 ENV NODE_ENV production
 
 # Configure server
-COPY --from=client /client/build /client/build/.
-WORKDIR /server
+# RUN apk add --no-cache git
+USER root
```
vyasworks, line 23: How are we allowing root for build? Does this apply only for the `tools` workspace? 

Can you point to the relevant config or add more details on how to do it as comments on the code? :)

.github/workflows/build.yml
fwpushan: OC CLI is preinstall in ubuntu latest ....

Dockerfile
fwpushan: USER 1001 is default user of rhel container, I find this in many CSI-Lab project but let me find appropriate document.

fwpushan: [Link](https://developers.redhat.com/blog/2020/10/26/adapting-docker-and-kubernetes-containers-to-run-on-red-hat-openshift-container-platform#)

Dockerfile
fwpushan: 1. This is container specific user (by default Redhat containers are non-root container)
2. Are you talking about commented code?  we are not using those applications  and there is no support for apk package manager in Redhat images (**yum**)

Dockerfile
fwpushan: nope CLI is required in pipeline. But that is already preinstalled in `ubuntu-latest` runner.

Dockerfile
vyasworks: I meant, we can add the point 1 with more details as a comment for future readers on why we have enabled root permissions here.

Dockerfile
```diff
@@ -1,28 +1,32 @@
 # Client
-FROM public.ecr.aws/bitnami/node:14.17.3 AS client
+FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS client
 
 # Build client
-WORKDIR /client
+# RUN apk add --no-cache git python g++ make
+USER root
```
fwkendall, line 6: I would leave this out

Dockerfile
fwkendall: I would leave root disabled, and only enable it for specific commands where needed

It's disabled for a reason, right? To protect us from breaking things 😓

Dockerfile
fwpushan: The commented code, right? 
For root user case.... our build script required some additional privileges. But at the end we switch non-root user.

Dockerfile
fwpushan: By default all RHEL images run as non-root user. Here we build app as root but in the end we switch to non-root user.

Dockerfile
fwkendall: I think we should be very deliberate about which commands require root, just as a matter of best practice—like I would enable it for only specific commands vs a batch of them where only one requires it. Feel free to make a case otherwise, I know it's more annoying but I think it creates better clarity around what root is actually required for and prevents accidentally adding commands later with more privilege than they need

Also we should remove the commented-out code lol

Dockerfile
fwpushan: And regular alpine/bitmani or ubuntu based images (like current one) run container as `root`. Please check this [blog](https://engineering.bitnami.com/articles/running-non-root-containers-on-openshift.html)

fwkendall: I think it's a bad practice 🤷‍♀️ but I also think the risk is low, so I'm fine with this if no one agrees with me!

Dockerfile
fwpushan: Currently, build script try to create some dir or file and getting access denied. I checked, those items are part of `WORKDIR`. So, quick solution will be give just write access to non-root user then switch back to non-root user

Dockerfile
Zenen-freshworks: I'm on-board with upholding best practices around permissions, it's a good habit and I can't see any downsides.

fwpushan:
I update Dockerfile to use minimal root context.

Dockerfile
fwkendall: Looks great Pushan thank you!

fwkendall (approved):

Dockerfile
```diff
@@ -1,28 +1,42 @@
...
 
 # Server
-FROM public.ecr.aws/bitnami/node:14.17.3 AS server
-
+FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS server
 # Static env vars
 ARG VERSION
 ENV VERSION $VERSION
 ENV NODE_ENV production
+ENV HOME_SEVER /opt/app-root/src/app/server
```
arranfw, line 23:
```suggestion
ENV HOME_SERVER /opt/app-root/src/app/server
```

Dockerfile
```diff
@@ -2,31 +2,41 @@
...
+
 # Server
 FROM registry.access.redhat.com/ubi8/nodejs-14:1 AS server
-
 # Static env vars
 ARG VERSION
 ENV VERSION $VERSION
 ENV NODE_ENV production
+ENV HOME_SEVER /opt/app-root/src/app/server
+ENV HOME_CLIENT /opt/app-root/src/app/client
```
arranfw, line 24: Is this declared twice? Once here and once on L5

arranfw: ahh it's a different container

arranfw (approved):

arranfw (approved):

Dockerfile
fwpushan: Yes

Zenen-freshworks (approved):

PR #462, opened 2021-07-21
[HCAP-700] Node Engines, Enforce NPM version via husky

vyasworks:
We want to enforce both Node and NPM version for all packages we use. 

The straight forward and easy approach is to use `engines` option provided by NPM. 

I have picked the least minor version of  NPM 7. 

We should see an error like this on npm install in the base folder. 

```
vyas@arch:hcap $ npm install
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: undefined
npm ERR! notsup Not compatible with your version of node/npm: undefined
npm ERR! notsup Required: {"node":">=14.0","npm":">=7.20.0"}
npm ERR! notsup Actual:   {"npm":"7.19.0","node":"v14.16.1"}
```

In addition I have added checks in pre-commit hook (probably redundant, but we can extend this for custom checks later on)


dbayly-freshworks (dismissed):

fwpushan (approved):
Nice

Zenen-freshworks (dismissed):
nice

.github/workflows/format.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/setup-node@v2
```
vyasworks, line 13: Github actions ships with NPM V6 by default, we use NODE 15 so that the bundles NPM V7 can be used

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
vyasworks, line 1: This is what actually enforces the checks or else we just get a warning and not `enforced errors`

.github/workflows/format.yml
fwpushan: we are not running in node inside pipeline os but docker inside pipeline os.

fwpushan (approved):

Zenen-freshworks (approved):
If this does what we want it to, I'm happy 👍

PR #464, opened 2021-07-22
Hcap 788 Added proper logging to the reconfirm interest action

dbayly-freshworks:


Zenen-freshworks (approved):
👍

arranfw:
Is this right? I understand the ticket as adding the action to the `participantBody.history` object like what's done here https://github.com/bcgov/hcap/blob/master/server/services/participants.js#L153

dbayly-freshworks:
> Is this right? I understand the ticket as adding the action to the `participantBody.history` object like what's done here https://github.com/bcgov/hcap/blob/master/server/services/participants.js#L153

It looks like I got logging and history mixed up, so the ticket is mislabled. 

The history object is correctly being updated, however, the action was no properly logged.

arranfw (approved):

PR #469, opened 2021-07-26
Update husky hooks

arranfw:
- Force husky to use bash for shell scripts
- Add `--staged` flag for pretty-quick [formats and then restages changes](https://github.com/azz/pretty-quick#--staged-only-git)
- rename script


Zenen-freshworks (approved):
Looks effective 👍

PR #483, opened 2021-07-29
Hcap 720 page skip issue

dbayly-freshworks:
* When you change tabs, the current page is set to zero 
* removed the Table components internal page state because it was causing issues. 
* Moved changing page from the main reducer to the tab reducer because changing tabs now changes page state.


Zenen-freshworks (approved):
This seems to be a good simplification of an overly complicated component. Happy to see it merged

fwpushan (approved):
LGTM

PR #488, opened 2021-08-03
Hcap 723 notification on archival

dbayly-freshworks:
When a user is withdrawn, their status is no-longer changed. 
Removed my old overly complicated withdrawn status patch for 696. 

NOTE: While initially I intended to address HCAP-722 here, after talking with Surabhi, more clarification is needed, so I have flagged the ticket and put it on hold. This PR is only for 723.

### DIFF
--- a/server/services/participants-helper.js
+++ b/server/services/participants-helper.js
@@ -66,40 +66,6 @@ const addDistanceToParticipantFields = (raw, siteDistanceJoin) =>
     }),
   }));
 
-// Find the previous status for each org, create a copy of it
-const revertToOldStatus = async (participantStatuses, setParticipantStatus) => {
-  await participantStatuses.forEach(async (status) => {
-    if (status?.data?.previousStatus !== participantStatus.ARCHIVED) {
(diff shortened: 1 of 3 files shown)


server/services/participants-helper.js
```diff
@@ -66,40 +66,6 @@ const addDistanceToParticipantFields = (raw, siteDistanceJoin) =>
...
-        {
-          final_status: 'Withdrawn from HCAP',
-          previousStatus: status.status,
-          previousData: status.data,
-        }
-      );
-    }
-  });
-};
-
```
Zenen-freshworks, line 102: Is this code totally obsolete then? It might relate to HCAP-804

Zenen-freshworks (approved):
Seems like a two-for one of resolving bugs and slimming the codebase.

fwpushan (approved):

server/services/participants-helper.js
dbayly-freshworks: Yes it is obsolete. And yes it should impact HCAP-804.

PR #493, opened 2021-08-03
HCAP-823 | Calculating Available Seats from Open Cohorts

Zenen-freshworks:
filtering out closed cohorts before summing remaining seats.

Addressing the following ticket: 
https://freshworks.atlassian.net/browse/HCAP-823?atlOrigin=eyJpIjoiZjg0NGZjODMwZGY0NGQ3ZThhMWRlZDg5YWI2MjU1MjYiLCJwIjoiaiJ9

### DIFF
--- a/client/src/pages/private/PSIView.js
+++ b/client/src/pages/private/PSIView.js
@@ -108,14 +108,15 @@ export default () => {
         const data = await response.json();
         const mappedData = data.map((row) => {
           const rowCohorts = cohorts.filter((cohort) => cohort.psi_id === row.id);
+          // To calculate available_seats, we filter out the expired cohorts and
+          // then sum the remaining seats of the current ones
           return {
             ...row,
             id: row.id,
             cohorts: rowCohorts.length,
-            available_seats: rowCohorts.reduce(
-              (sum, cohort) => sum + (cohort.cohort_size - cohort.participants.length),
-              0
-            ),
+            available_seats: rowCohorts
+              .filter((cohort) => new Date(cohort.end_date) > new Date())
+              .reduce((sum, cohort) => sum + (cohort.cohort_size - cohort.participants.length), 0),
           };
         });
         setPSIs(mappedData);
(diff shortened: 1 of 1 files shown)


client/src/pages/private/PSIView.js
```diff
@@ -108,14 +108,15 @@ export default () => {
         const data = await response.json();
         const mappedData = data.map((row) => {
           const rowCohorts = cohorts.filter((cohort) => cohort.psi_id === row.id);
```
arranfw, line 110: Unrelated to this PR, but this looks like we're getting all cohorts from the API and then reducing to the current PSI, why not pass the PSI ID to the API and only retrieve the records we want? It looks like there's logic written on the API to retrieve a specific PSI's cohorts.

client/src/pages/private/PSIView.js
```diff
@@ -108,14 +108,15 @@ export default () => {
...
+          // then sum the remaining seats of the current ones
           return {
             ...row,
             id: row.id,
             cohorts: rowCohorts.length,
-            available_seats: rowCohorts.reduce(
-              (sum, cohort) => sum + (cohort.cohort_size - cohort.participants.length),
-              0
-            ),
+            available_seats: rowCohorts
```
arranfw, line 117: I feel like the API could be used a bit more here as well, the API could return the list of cohorts as one attribute and the available seats as another.

IMO this is too much processing to be done on the FE particularly in useEffect and could be handled easier on the API

arranfw (approved):
Looks good, I left a couple comments that are mostly unrelated to this PR.

client/src/pages/private/PSIView.js
fwpushan, line 117: To improve efficiency we can declare new Date() in top scope. So, that help us to avoid every time creation new Date(). Just a thought sharing with you.

fwpushan (approved):

client/src/pages/private/PSIView.js
Zenen-freshworks: In the case of the PSIView, we actually do want all of the cohorts since we're using all of the PSIs. It could probably be done more efficiently but nothing really jumped out at me.

fwpushan (approved):

arranfw (approved):

PR #497, opened 2021-08-05
Hcap 720 page skip fix

dbayly-freshworks:
It looks like stripping out pagination from all the other tables fixes 822, so I am re-adding my changes for 720. 
I've tested this out manually a fair bit and it looks. good. 
I have some concerns about this though, mainly around performance if we have large amounts of data on prod. 

Please let me know what you think.


arranfw (approved):
Works locally, looks great

fwpushan (approved):
LGTM

PR #508, opened 2021-08-10
Hcap 849 notification on withdrawal

dbayly-freshworks:
This PR handles two cases. 

* When a hired user is archived by MoH, a warning icon will show up in the status column, and there's an option to move them to archives 
* When a HA archives a user, a separate status 'pending_acknowledgement' will be created to inform the related employer that this person has been archived. It has an Acknowledge button that dismisses the status. 

Screen Shot 2021-08-10 at 1 58 51 PM


client/src/pages/private/ParticipantTable.js
```diff
@@ -244,7 +245,31 @@ export default () => {
       });
     }
   };
-
+  const handleAcknowledge = async (id) => {
```
fwpushan, line 248: For now this is Ok, but in future we should remove all api call logic to service layer. This file is also growing massively.

server/services/participants.js
```diff
@@ -211,6 +149,146 @@ const withdrawParticipant = async (participantInfo) => {
...
+      status,
+      current: true,
+      data,
+    });
+
+    const participant = await tx[collections.PARTICIPANTS].findDoc({
+      id: participantId,
+    });
+    // Now check if current status is archived then set interested flag
+    if (status === 'archived') {
```
fwpushan, line 265: Now there is a dedicated method to archive participant. we should remove this status check. And put code under if to archive participant method. Any thought.

fwpushan (commented):
LGTM. Just one clarification require, please check my comments.

client/src/pages/private/ParticipantTable.js
dbayly-freshworks: Agreed!

server/services/participants.js
dbayly-freshworks: I have mixed feelings about this. One of the benefits of using the changeStatus route for archiving is that the user can only archive their own participants.

fwpushan (dismissed):

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -247,17 +247,22 @@ export default ({ id, siteId, onArchiveParticipantAction }) => {
     };
 
     fetchParticipants();
-  }, [siteId, setRows, setFetchedRows, setFetchedWithdrawnRows, setLoadingData]);
+  }, [id, setRows, setFetchedRows, setFetchedWithdrawnRows, setLoadingData]);
```
fwkendall, line 250: I think this siteId vs id naming was deliberate because site entities have both keys in the DB... this change could introduce some confusion, we might want to revert to the previous name in a subsequent PR

fwkendall: unless this WAS using the id and not the siteId, in which case this is a good fix!

fwkendall (dismissed):

client/src/pages/private/SiteParticipantsTable.js
fwkendall: I think the key thing is that id != siteId

So I just want to make sure we're still functional and using the correct term in the code, to reduce future confusion about this...

fwkendall (approved):

client/src/pages/private/SiteParticipantsTable.js
dbayly-freshworks: I removed siteId from the component to prevent confusion. I would rename the variable as well, but because it's a prop this really is an issue with the parent component, and I don't want to tack a rework on to this PR, it's big enough.

fwkendall: Sounds good to me - if there's still cleanup that would make sense to do, let's track as tech debt

fwpushan (approved):

PR #511, opened 2021-08-13
[HCAP-843]  emit JSON logs in non local env

vyasworks:
https://freshworks.atlassian.net/browse/HCAP-843

- emit logs as json on non local 
- Update some log statements

### DIFF
--- a/server/logger.js
+++ b/server/logger.js
@@ -9,6 +9,10 @@ const formatWithTimestamp = printf(({ level, message, label: _label, timestamp: 
   return `${_timestamp} | [${_label || 'console'}] ${level}: ${newMessage}`;
 });
 
+// ============
+// TODO: Mongo logs to be removed, containers, bc, dc, pvc destroyed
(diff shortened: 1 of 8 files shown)


server/logger.js
```diff
@@ -26,12 +30,18 @@ if (dbServer) {
   );
 }
 
-if (process.env.NODE_ENV !== 'test') {
+if (process.env.NODE_ENV === 'local') {
```
dbayly-freshworks, line 33: Why was this check switch?

vyasworks: We want to emit JSON logs in OCP so that elastic search can index them. 

While only in local we want to use a more human readable format.

dbayly-freshworks: Makes sense. 👍

dbayly-freshworks (approved):

fwpushan (approved):
LGTM. Want to see ELK in action asap.

PR #520, opened 2021-08-17
HCAP-850 Migration

dbayly-freshworks:
Migration to remove duplicate archived or rejected statuses. 

This is causing the pagination to display incorrect results on the Archived Candidates tab. (HCAP - 850)

### DIFF
--- a/server/migrations/1629228977039_remove-duplicates.js
+++ b/server/migrations/1629228977039_remove-duplicates.js
@@ -1,0 +1,30 @@
+/* eslint-disable camelcase */
+
+const { dbClient } = require('../db');
+
+exports.up = async () => {
+  const duplicatedIds = await dbClient.db.query(`
+            SELECT 
+            id
+            from participants_status as p1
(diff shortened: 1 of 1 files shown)


server/migrations/1629228977039_remove-duplicates.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable camelcase */
+
+const { dbClient } = require('../db');
+
+exports.up = async () => {
```
fwpushan, line 5: We can achieve this using single query. Try using `With` clause of PGSQL.

fwpushan (changes requested):
Example:
```sql
with to_update as (
      select id,
      from participant
      where body->>interested = 'no';
    )
    update participant
    set body =body(resource, path::text[], to_jsonb(value))
    from to_update
    where participant.id = to_update.id
```

Zenen-freshworks (approved):
Cool! I've never known how to use `with` subclauses before.

fwpushan (approved):

PR #526, opened 2021-08-19
Fixed various small issues.

dbayly-freshworks:
- Removed pending withdrawal status from employer actions route
- Added siteId to useEffect for the selected tab so that on load, the default tab will be the site details tab.
- Added HA to the acknowledge route to cover the case where a HA withdraws another HA's participant


fwpushan (approved):
Just require one clarification. When `pending_acknowledgement` is added ?

dbayly-freshworks:
> Just require one clarification. When `pending_acknowledgement` is added ?

In one niche case. When an HA withdraws a participant that they specifically did not hire from the site details screen. 
This is handled by the `employer-details/archive` route.

Zenen-freshworks (approved):

PR #527, opened 2021-08-19
HCAP-732 | seeding participants and sites

Zenen-freshworks:
This is a general QoL improvement for testing as well as on-boarding new developers to the project. By seeding the database with sites and participants when the initial migrations are run, we can run tests on an environment more similar to the production deployment (instead of empty tables everywhere).

Addresses the following ticket:
https://freshworks.atlassian.net/browse/HCAP-732?atlOrigin=eyJpIjoiN2ViMDhmNjc0YjE0NGU2ODhmZmU3ZDNhMWU0NWJhMjgiLCJwIjoiaiJ9


server/migrations/1629329748833_feed-database.js
```diff
@@ -0,0 +1,67 @@
...
+  'EOI - VCHA': 'vancouverCoastal',
+  'EOI - VIHA': 'vancouverIsland',
+  'CB1: Still Interested': 'interested',
+  'CB8: Non-HCAP Opportunities': 'nonHCAP',
+  'CB13: CRC Clear': 'crcClear',
+};
+
+exports.up = async () => {
+  const nodeEnv = process.env.NODE_ENV;
+  if (nodeEnv === 'test' || nodeEnv === 'local' || nodeEnv === 'dev') {
```
dbayly-freshworks, line 50: We should be careful here.  This will seed for local and in tests just fine, but for all deployed environments NODE_ENV will be `production`. If we plan on doing any seeding on `dev`/`test`, we should wait for HCAP-870 to be complete so we have an accurate environment specific variable.

Zenen-freshworks: Fully willing to shelve this to wait for HCAP-870. In hindsight, however, it's only really needed for the 'local' and 'test' environments, the latter being what gets run within the pipeline.

fwpushan:
@Zenen-freshworks Can you check why test is failing?

fwpushan (changes requested):
1. I prefer calling this as pure  seeding script. Migration is more like database versioning. So, please don't use migration.
2. Factories are ideal for seeding. You can explore that 
3. Just run automated script to populate data in pipeline.

Zenen-freshworks:
restarting to avoid seeding via migration.

PR #530, opened 2021-08-19
HCAP- 906 / 907 hotfix

dbayly-freshworks:
Fixed 906/907, looks like it was a site id issue
Added an error if the archived user could not be found.

### DIFF
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -275,7 +275,7 @@ export default ({ id, siteId, onArchiveParticipantAction, stale, setStale }) => 
       },
       body: JSON.stringify({
         participantId,
-        site: parseInt(id),
+        site: parseInt(siteId),
         data: additional,
         status: 'archived',
       }),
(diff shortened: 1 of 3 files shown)


client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -275,7 +275,7 @@ export default ({ id, siteId, onArchiveParticipantAction, stale, setStale }) =>
       },
       body: JSON.stringify({
         participantId,
-        site: parseInt(id),
+        site: parseInt(siteId),
```
Zenen-freshworks, line 278: I found out from Pushan that `+siteId` also works instead of passing it through parseInt.. just a lil bit of sytactic sugar

Zenen-freshworks (approved):

client/src/pages/private/SiteParticipantsTable.js
dbayly-freshworks: Neat. Every day I learn something new. I'll implement it.

fwkendall (approved):

PR #544, opened 2021-08-26
Hcap 880 peoi landing updates

dbayly-freshworks:
I have implemented everything in the designs except for the `View more` link. I need to confer with Kendall about what we're expecting there. Also I think the wording needs updating. 

Changes: 
* Updated styling to match closer to the designs. 
* Added a route to withdraw from all PEOIs 
* Updated the fetch route to also get the hired participants_statuses 

Screen Shot 2021-08-26 at 4 39 06 PM


fwpushan (approved):
👍

fwkendall (approved):

PR #555, opened 2021-09-03
[HCAP-929] Sanitize outputs to avoid XSS via inputs

vyasworks:


fwpushan (approved):
Possibly using of middleware like https://express-validator.github.io/docs/filter-api.html is more appropriate.

arranfw (approved):

PR #561, opened 2021-09-23
HCAP-970 | Fix security warnings

fwpushan:
----
1. Add or update header to disable security warning
2. Block options and trace method in express
3. Skip one UI test because of insufficient data setup the test is failing intermittently. ([HCAP-974](https://freshworks.atlassian.net/browse/HCAP-974))

Ticket: [HCAP-970](https://freshworks.atlassian.net/browse/HCAP-970)


server/server.js
```diff
@@ -25,27 +25,45 @@ app.use(
         'default-src': ["'self'"],
         'connect-src': [
           "'self'",
-          'https://*.apps.gov.bc.ca',
```
fwkendall, line 28: hmmm do we have a sense of what this was originally added to support? might be worth a git-blame and checking the PR

server/server.js
```diff
@@ -25,27 +25,45 @@ app.use(
...
         ],
         'base-uri': ["'self'"],
         'block-all-mixed-content': [],
-        'font-src': ["'self'", 'https:', 'data:'],
+        'font-src': ["'self'"],
         'frame-ancestors': ["'self'"],
         'img-src': ["'self'", 'data:'],
         'object-src': ["'none'"],
-        'script-src': ["'self'", 'https://*.gov.bc.ca'],
+        'script-src': ["'self'"],
```
fwkendall, line 39: this change looks good - we likely had the gov address enabled to support analytics, but we never ended up adding any gov analytics

server/server.js
fwpushan: I check line history this was added as part of initial helmet config.

dbayly-freshworks (dismissed):
This looks ok for now. I do feel a bit weird skipping a test though.

dbayly-freshworks (approved):
Looks ok, though I still have mixed feelings about skipping a test. We should look at fixing it later.

fwpushan:
> Looks ok, though I still have mixed feelings about skipping a test. We should look at fixing it later.

100% agree with you, thats why I added a task to handle that.

PR #568, opened 2021-11-03
HCAP-979 | disable haproxy router cookie

arranfw:
https://freshworks.atlassian.net/browse/HCAP-979


arranfw:
currently tagged on dev, I don't see any regression in routing behaviour

dbayly-freshworks (approved):

PR #575, opened 2021-11-08
HCAP-983 | overwrite non-object histories

arranfw:
If the current history value isn't an object then overwrite it with the given entry

https://freshworks.atlassian.net/browse/HCAP-983

### DIFF
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -123,9 +123,11 @@ const createChangeHistory = (participantBody, changes) => {
     timestamp: new Date(),
     changes: [...changeDetails],
   };
-  newBody.history = participantBody.history
-    ? [newHistory, ...participantBody.history]
-    : [newHistory];
+  // handle histories that have bad data/aren't objects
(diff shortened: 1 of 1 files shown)


server/services/participants.js
```diff
@@ -123,9 +123,11 @@ const createChangeHistory = (participantBody, changes) => {
     timestamp: new Date(),
     changes: [...changeDetails],
   };
-  newBody.history = participantBody.history
-    ? [newHistory, ...participantBody.history]
-    : [newHistory];
+  // handle histories that have bad data/aren't objects
```
fwpushan, line 126: The previous code is perfectly alright.

arranfw: It is if prod doesn't have integer histories, otherwise this code needs changing doesn't it?

arranfw:
Closing because the data errors don't exist in prod, they can be fixed via manual prod edits so this bug won't require a code change

PR #581, opened 2021-11-09
fix blocker XSS sonarcloud errors

arranfw:


fwpushan (commented):
LGTM. But you are sanitising response, do we need that. For response, source of truth is our application.

PR #584, opened 2021-11-12
HCAP-971 | Disabled form redesign

arranfw:
https://freshworks.atlassian.net/browse/HCAP-971

- There's a bottom navbar in the designs but I didn't put it in because it feels like a separate ticket and it isn't present in other screens
- Didn't add custom styling for the email input to stay consistent with the rest of the app (can be added if we want a second input design in the app)
- Designs were a bit complex with the responsive behaviour so the CSS is a little cumbersome

image
image
image


client/src/components/participant-form/DisabledForm.js
```diff
@@ -0,0 +1,226 @@
...
+    [theme.breakpoints.down('sm')]: {
+      width: '100%',
+    },
+  },
+}));
+
+export const DisabledForm = () => {
+  const classes = useStyles();
+  const handleSubmit = async (values, { setSubmitting }) => {
+    console.log('submitted', { values });
```
dbayly-freshworks, line 147: Stay console log here

client/src/components/participant-form/DisabledForm.js
```diff
@@ -0,0 +1,226 @@
...
+    await new Promise((resolve) => setTimeout(resolve, 500));
+    setSubmitting(false);
+  };
+
+  return (
+    <Box className={classes.container}>
+      <img src={heroBackground} alt='' className={classes.hero} />
```
dbayly-freshworks, line 154: alt is empty here.

dbayly-freshworks (approved):
Looks ok, two nits.

arranfw:
@dbayly-freshworks 
Empty alt is by design, decorative images should use empty alt tags to explicitly designate them as decorative https://www.w3.org/WAI/tutorials/images/decorative/

Console log is removed in next PR, are you cool with merging?

PR #586, opened 2021-11-15
Relabeled button plus fixed some issues with Indigenous identity popup

dbayly-freshworks:
Changed the label of the `Go back` button after Reconfirm interest. 

I had some records that had a number for their history object. This was preventing the `Are you Indigenous` popup from closing, and locked me out of the PEOI portal. Included are some minor fixes that helped with that issue, but we should probably add the ability to close the popup to give the user the ability to get out if this issue shows up on prod.

### DIFF
--- a/client/src/pages/private/ParticipantWithdrawConfirm.js
+++ b/client/src/pages/private/ParticipantWithdrawConfirm.js
@@ -36,7 +36,7 @@ export default () => {
           variant='contained'
           onClick={() => history.push(Routes.ParticipantLanding)}
         >
-          Return Home
+          Go Back
         </Button>
       </Box>
     </Page>
--- a/client/src/pages/private/ParticipantWithdrawConfirm.js
+++ b/client/src/pages/private/ParticipantWithdrawConfirm.js
@@ -36,7 +36,7 @@ export default () => {
           variant='contained'
           onClick={() => history.push(Routes.ParticipantLanding)}
         >
-          Go Back
+          Return Home
         </Button>
       </Box>
     </Page>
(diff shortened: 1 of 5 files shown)


arranfw:
@fwpushan Decided against making code changes for the history-as-integer issue as we didn't see any occurences of that data in prod.

I think leaving the form open with an error popping up is okay behaviour and updating the integers to nulls in your local db is the best option

client/src/pages/private/ParticipantWithdrawConfirm.js
```diff
@@ -36,7 +36,7 @@ export default () => {
           variant='contained'
           onClick={() => history.push(Routes.ParticipantLanding)}
         >
-          Return Home
```
arranfw, line 39: I think we want to leave this as-is, this page only returns to the homepage whereas the other one returns to the PEOI

arranfw (changes requested):

server/services/participants.js
```diff
@@ -132,7 +132,7 @@ const createChangeHistory = (participantBody, changes) => {
 const updateParticipant = async (participantInfo) => {
   // The below reduce function unpacks the most recent changes in the history
   // and builds them into an object to be used for the update request
-  const changes = participantInfo.history[0].changes.reduce(
+  const changes = participantInfo.history[0]?.changes.reduce(
```
arranfw, line 135: Do we need the changes in this file? If they aren't open bugs or blockers I'd like to avoid changing them, optional chaining can hide other issues

dbayly-freshworks: Sure, I can remove this.

arranfw (approved):

PR #593, opened 2021-12-13
HCAP-995 | Fix bug: Type error | Remove depreciated escape method

fwpushan:


server/utils.js
```diff
@@ -29,6 +29,6 @@ const patchObject = (source, patchableFields) =>
     {}
   );
 
-const sanitize = (input) => input.trim().escape();
```
arranfw, line 32: Wonder if we should use a library for sanitizing

arranfw (approved):

dbayly-freshworks (approved):
Looks fine to me

PR #594, opened 2021-12-13
HCAP-941 | Add content for interest = blank

arranfw:
When `interested === null` the following status is shown

https://freshworks.atlassian.net/browse/HCAP-941

image

### DIFF
--- a/client/src/pages/private/ParticipantLanding.js
+++ b/client/src/pages/private/ParticipantLanding.js
@@ -242,6 +242,9 @@ export default () => {
           } else if (item.hired?.length > 0) {
             color = '#17d149';
             status = 'Hired';
+          } else if (!item.interested) {
+            color = '#735613';
+            status = 'Pending';
           }
 
           return (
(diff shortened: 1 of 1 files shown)


client/src/pages/private/ParticipantLanding.js
```diff
@@ -242,6 +242,9 @@ export default () => {
           } else if (item.hired?.length > 0) {
             color = '#17d149';
             status = 'Hired';
+          } else if (!item.interested) {
+            color = '#735613';
```
dbayly-freshworks, line 246: Does Hcap have themes?

dbayly-freshworks (approved):
Looks good 👍

fwpushan (approved):

client/src/pages/private/ParticipantLanding.js
arranfw: There is a theme (although it's out of date when considering recent designs), rather than either making something inconsistent with previous lines or refactoring the surrounding code I opted to extend the existing pattern. I don't think the effort vs. reward is worth it here considering our timelines

PR #596, opened 2021-12-13
Fixed 923 and 691

dbayly-freshworks:
Updated the BCSC login with correct link 

Fixed issue where when viewing an EEOI the values were not being populated. 
The tldr is that the initial values of a formik form are only checked once. Since we were fetching the EEOI form values, the default values were already checked before the fetch resolved. I added a cheap and simple workaround.

### DIFF
--- a/client/src/components/employer-form/index.js
+++ b/client/src/components/employer-form/index.js
@@ -13,6 +13,7 @@ import Hidden from '@material-ui/core/Hidden';
 import KeyboardArrowLeft from '@material-ui/icons/KeyboardArrowLeft';
 import KeyboardArrowRight from '@material-ui/icons/KeyboardArrowRight';
 import Typography from '@material-ui/core/Typography';
+import CircularProgress from '@material-ui/core/CircularProgress';
 
 import { API_URL, EmployerFormSchema, Routes, ToastStatus } from '../../constants';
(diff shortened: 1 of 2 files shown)


client/src/components/employer-form/index.js
```diff
@@ -202,6 +202,23 @@ export const Form = ({ hideCollectionNotice, initialValues, isDisabled }) => {
     }
   };
 
+  if (!initialValues && isDisabled) {
```
arranfw, line 205: This prevents setting initial values twice? ie:

1st render: `initialValues = undefined`
2nd render: `initialValues = {...initialValues}` // <-- wouldn't do anything because formik doesn't look for updates to initialValues?

If so, another low-weight solution could be the [enableReinitialize](https://formik.org/docs/api/formik#enablereinitialize-boolean) prop

arranfw (approved):

client/src/components/employer-form/index.js
dbayly-freshworks: Neat! I had no idea that formik had already thought of this. I'll give it a try.

fwpushan (dismissed):
LGTM

arranfw (approved):

arranfw (approved):

vyasworks (approved):

PR #597, opened 2021-12-16
Added copy change

dbayly-freshworks:
Screen Shot 2021-12-15 at 4 58 37 PM
Last minute copy change for the portal

### DIFF
--- a/client/src/pages/public/ParticipantLogin.js
+++ b/client/src/pages/public/ParticipantLogin.js
@@ -90,10 +90,11 @@ export default () => {
             <Typography variant='h2'>Welcome to the Health Career Access Program</Typography>
           </Box>
           <Typography mb={2}>
-            COVID-19 has increased the need for health care assistants in long-term care and
-            assisted living settings across the province. The Health Career Access Program provides
-            a path for applicants with no health care experience to get hired and receive paid
-            employer sponsored health care assistant training as part of their employment.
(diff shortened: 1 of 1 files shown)


client/src/pages/public/ParticipantLogin.js
```diff
@@ -90,10 +90,11 @@ export default () => {
             <Typography variant='h2'>Welcome to the Health Career Access Program</Typography>
           </Box>
           <Typography mb={2}>
-            COVID-19 has increased the need for health care assistants in long-term care and
-            assisted living settings across the province. The Health Career Access Program provides
-            a path for applicants with no health care experience to get hired and receive paid
-            employer sponsored health care assistant training as part of their employment.
+            COVID-19 has increased the need for health care assistances in long-term care and
```
fwkendall, line 93: I noticed the change here from "assistants" to "assistances" and confirmed that while it isn't bolded, it is *definitely* present in the copy provided via email

In context I think it makes sense to reflect the broadening scope of the program, so I think we're all good to proceed with this

fwkendall: And by bolded I mean in terms of what was highlighted to draw my attention with respect to copy changes (not that it should literally be bolded as part of our changes 😅 )

fwkendall (approved):

fwpushan (approved):

PR #606, opened 2022-01-18
EHPR-1008 | Don't install packages in containers

arranfw:
We already had `node_modules` mounted as a volume for client and server so this is just removing the redundant `npm i` step from our dev dockerfiles.


package.json
```diff
@@ -9,9 +9,10 @@
     "format:lint-client": "eslint --ext js,jsx -c ./client/.eslintrc ./client --fix",
     "format:lint-server": "eslint --ext js,jsx -c ./server/.eslintrc ./server --fix",
     "update-all": "run-s 'update:*'",
+    "update-apps": "run-p 'update:client' 'update:server'",
     "update:root": "npm i",
-    "update:client": "npm i --prefix client",
-    "update:server": "npm i --prefix server",
+    "update:client": "npm ci --prefix client",
```
arranfw, line 14: `npm ci` will be slower for local development, we should use `npm i` here

fwpushan: npm i is updating package-lock.json .....

arranfw: then package-lock is out of date for some reason 🤔

fwpushan: yes...thats true

fwpushan: Reverted.

dbayly-freshworks (approved):
Looks good to me

PR #615, opened 2022-01-19
HCAP-1022 | Merge local and context reducers

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1022

### DIFF
--- a/client/src/components/generic/DebounceTextField.js
+++ b/client/src/components/generic/DebounceTextField.js
@@ -17,7 +17,7 @@ export const DebounceTextField = ({ time, onDebounce, onChange, ...props }) => {
 
   const handleOnChange = (element) => {
     setText(element.target.value);
-    onChange(element);
+    onChange && onChange(element);
   };
 
   return <TextField {...props} onChange={(element) => handleOnChange(element)} />;
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useReducer, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import Grid from '@material-ui/core/Grid';
 import { Box, Menu, MenuItem, Link } from '@material-ui/core';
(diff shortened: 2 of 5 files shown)


client/src/components/generic/DebounceTextField.js
```diff
@@ -17,7 +17,7 @@ export const DebounceTextField = ({ time, onDebounce, onChange, ...props }) => {
 
   const handleOnChange = (element) => {
     setText(element.target.value);
-    onChange(element);
+    onChange && onChange(element);
```
dbayly-freshworks, line 20: Are there issues with onChange not being defined?

client/src/pages/private/ParticipantTable.js
```diff
@@ -282,22 +257,21 @@ const ParticipantTable = () => {
...
-        reducerState.emailFilter || '',
-        reducerState.order.field,
-        reducerState.order.direction,
-        reducerState.siteSelector,
+      const { data, pagination: newPagination } = await fetchParticipants(
+        pagination.page * pageSize,
+        filter.locationFilter?.value || '',
+        filter.fsaFilter?.value || '',
+        filter.lastNameFilter?.value || '',
+        filter.emailFilter?.value || '',
```
dbayly-freshworks, line 265: Threre's still a fair bit of duplication here. I wonder if creating a prop type could be useful here.

client/src/components/generic/DebounceTextField.js
fwpushan: Yes, I've same question

fwpushan (dismissed):
LGTM

client/src/components/generic/DebounceTextField.js
arranfw: This component handle it's own textfield state so the onchange/value props aren't really needed. I removed those values from the implementations of this component and only use `defaultValue`/`onDebounce` props. Removing `onChange` cause an error here

client/src/pages/private/ParticipantTable.js
arranfw: Totally, this gets alleviated a bit in https://github.com/bcgov/hcap/pull/617

dbayly-freshworks (approved):
Perfectly balanced as all PRs should be.

PR #619, opened 2022-01-28
HCAP - 1245 Fixed callback issue with edit participant component

dbayly-freshworks:
It looks like we were trying to call an undefined submissionCallback function. 
I added a check to ensure it exists. I also added a fetch data call to the onClose function for the EditParticipant component in the ParticipantDetailsView to load the updated data after it's submitted.

### DIFF
--- a/client/src/pages/private/ParticipantDetailsView.js
+++ b/client/src/pages/private/ParticipantDetailsView.js
@@ -304,7 +304,17 @@ export default () => {
               initialValues={actualParticipant}
               validationSchema={EditParticipantFormSchema}
               onSubmit={onUpdateInfo}
-              onClose={() => setShowEditModal(false)}
+              onClose={() => {
+                setShowEditModal(false);
+                fetchData({
+                  setParticipant,
+                  setPSIList,
+                  setActualParticipant,
+                  setDisableAssign,
+                  setError,
+                  id,
+                });
+              }}
             />
           </Dialog>
         )}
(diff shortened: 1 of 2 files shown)


fwkendall (approved):

client/src/pages/private/ParticipantDetailsView.js
```diff
@@ -304,7 +304,17 @@ export default () => {
               initialValues={actualParticipant}
               validationSchema={EditParticipantFormSchema}
               onSubmit={onUpdateInfo}
-              onClose={() => setShowEditModal(false)}
+              onClose={() => {
+                setShowEditModal(false);
+                fetchData({
```
fwpushan, line 309: 1. Is this section updating the Participant table?

fwpushan (commented):
LGTM, please check my comments.

fwpushan (approved):

client/src/pages/private/ParticipantDetailsView.js
```diff
@@ -304,7 +304,17 @@ export default () => {
               initialValues={actualParticipant}
               validationSchema={EditParticipantFormSchema}
               onSubmit={onUpdateInfo}
-              onClose={() => setShowEditModal(false)}
+              onClose={() => {
```
arranfw, line 307: Looks like I missed an instance of `EditParticipantForm` 🤔 heads up that `validationSchema` and `onSubmit` aren't used by the component anymore, if this page uses the same component but has a different edit function `EditParticipantForm` might need to be looked at again.

Do you know if it's intentional that these two pages have different onSubmits?

dbayly-freshworks:
I am going to merge this, and then put up a separate PR with the related test.

PR #621, opened 2022-01-31
Hcap 789 fix local cypress tests

dbayly-freshworks:
I've fixed cypress tests for me locally (for the most part) .

Some environment variables needed. 
-- AUTH_URL needs to be set to the local keycloak url, http://keycloak.local.freshworks.club:8080/auth
-- The port of the backend url needs to be set to 8081
-- KEYCLOAK_LOCAL_AUTH_URL needs to be set to the same as the auth URL 

Keycloak needs some configuration 
-- The local service account needs to have it's password reset 
-- The local keycloak client secret needs to be set 

I rebuilt the login flow for local by copying the headers/paramteters that my browser used into postman and then translating that into cypress. 

Because the two login processes were quite different, I split them into two separate functions. I tried adding a variable into cypress.json to track this, but for some reason it's not being picked up. For now, just replace `Cypress.env('isLocal')` in the if statement with true if you want to run locally. 

I also included a test for my most recent ticket HCAP-1028. Adding data-cy attributes is best practice and something we should continue doing going forward. 

There are still some local issues, and I hope to fix them in good time. 

Screen Shot 2022-01-31 at 3 17 36 PM

### DIFF
--- a/client/src/components/modal-forms/EditParticipantForm.js
+++ b/client/src/components/modal-forms/EditParticipantForm.js
@@ -54,8 +54,18 @@ export const EditParticipantForm = ({ initialValues, onClose, submissionCallback
       {({ submitForm }) => (
         <FormikForm>
           <Box>
-            <Field name='firstName' component={RenderTextField} label='First Name' />
-            <Field name='lastName' component={RenderTextField} label='Last Name' />
+            <Field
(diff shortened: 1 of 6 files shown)


fwpushan (approved):
👍

PR #624, opened 2022-02-01
HCAP-1043 | HCAP-1030 | Add migration to create new table `participant_post_hire_status`

fwpushan:


server/migrations/1643746660201_create-participant-post-hire-status.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable camelcase */
+const { dbClient, schema, collections } = require('../db');
+
+exports.shorthands = undefined;
+
+exports.up = async () => {
+  await dbClient.runRawQuery(schema.postHireRelationTables[0].definition);
```
dbayly-freshworks, line 7: Instead of running a raw query, it would be better practice to use the query builder like we did for the `capture-waitlist-emails` migration

server/migrations/1643746660201_create-participant-post-hire-status.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable camelcase */
+const { dbClient, schema, collections } = require('../db');
+
+exports.shorthands = undefined;
```
dbayly-freshworks, line 4: What does this do?

fwpushan: This is part of autogenerated code.

server/migrations/1643746660201_create-participant-post-hire-status.js
fwpushan: Updated please check

dbayly-freshworks (approved):

PR #625, opened 2022-02-02
HCAP-1047 | HCAP-1030 | Create Service for Post Hire Flow and Update existing services.

fwpushan:
- Create service module with create and getter 
- Update existing `getParticipants` to include post-hire-statuses in response 
- Add and update tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ yarn-error.log*
 .vscode
 .eslintcache
 key
+coverage
 
 sysdig_users
 .scannerwork
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -5,6 +5,7 @@ const { validate, ParticipantBatchSchema, isBooleanValue } = require('../validat
 const { dbClient, collections } = require('../db');
 const { createRows, verifyHeaders } = require('../utils');
 const { ParticipantsFinder } = require('./participants-helper');
+const { getPostHireStatusesForParticipant } = require('./post-hire-flow');
 const logger = require('../logger.js');
 
 const deleteParticipant = async ({ email }) => {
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -32,6 +32,7 @@ yarn-error.log*
 .vscode
 .eslintcache
 key
+coverage
```
dbayly-freshworks, line 35: What is in the coverage folder? Do I need up update my local?

fwpushan: Test coverage report folder. This will address the issue.

server/tests/participant.service.test.js
```diff
@@ -1023,4 +1026,44 @@ describe('Participants Service', () => {
...
+
+    const participants = await getParticipants(
+      { isEmployer: true, id: employerAId, regions },
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
+      null,
```
dbayly-freshworks, line 1064: This feels like a bit of a code smell here. We may want to refactor the parameters of the getParticipants function.

fwpushan: we may need to refactor getParticipants func. But this is not in the scope of this sprint.

server/services/participants.js
```diff
@@ -420,6 +421,17 @@ const getParticipants = async (
     total: Number(await table.count(criteria || {})),
   };
 
+  // HCAP:1030: Get participants post-hire statuses
+  participants = await Promise.all(
```
dbayly-freshworks, line 425: What are your thoughts around adding post hire statuses to the response  body of the main getParticipants call?

fwpushan: in participant view we have to show graduation status ....

dbayly-freshworks (approved):

PR #630, opened 2022-02-07
HCAP-1052 | HCAP-1030 | Show graduation completion status on participant list view

fwpushan:
- [x] Display Graduation status for MoH/SuperUser
- [x] Display Graduation Status For HA/Employer

### DIFF
--- a/client/src/constants/participantTableConstants.js
+++ b/client/src/constants/participantTableConstants.js
@@ -129,9 +129,10 @@ const columns = {
   callbackStatus: { id: 'callbackStatus', name: 'Callback Status', sortOrder: 14 },
   userUpdatedAt: { id: 'userUpdatedAt', name: 'Last Updated', sortOrder: 15 },
   engage: { id: 'engage', name: null, sortOrder: 16 },
-  edit: { id: 'edit', name: null, sortOrder: 17 },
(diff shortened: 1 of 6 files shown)


client/src/constants/participantTableConstants.js
```diff
@@ -129,9 +129,10 @@ const columns = {
   callbackStatus: { id: 'callbackStatus', name: 'Callback Status', sortOrder: 14 },
   userUpdatedAt: { id: 'userUpdatedAt', name: 'Last Updated', sortOrder: 15 },
   engage: { id: 'engage', name: null, sortOrder: 16 },
-  edit: { id: 'edit', name: null, sortOrder: 17 },
+  edit: { id: 'edit', name: null, sortOrder: 21 },
```
dbayly-freshworks, line 132: total nit, maybe shift this value to the end of the list

dbayly-freshworks (approved):
One nit, otherwise looks great

client/src/constants/participantTableConstants.js
fwpushan: This is in object format ... technically position is irrelevant but for consistency purpose we can change this.

dbayly-freshworks (approved):

PR #633, opened 2022-02-09
Hcap 1031 task for failed workflow

dbayly-freshworks:
Added post hired status of 'failed_cohort' on participant withdrawal.

### DIFF
--- a/client/src/services/participant.js
+++ b/client/src/services/participant.js
@@ -6,11 +6,15 @@ const getCohortName = (cohort = {}) =>
     ? `${cohort.cohort_name} / ${cohort.psi?.institute_name}`
     : 'Not Assigned';
 
-export const getPostHireStatusLabel = ({ status, data = {} } = {}) => {
-  if (status === postHireStatuses.postSecondaryEducationCompleted) {
-    return `Graduation Completed on - ${data.graduationDate}`;
+const getPostHireStatusLabel = ({ status, data = {} } = {}) => {
+  switch (status) {
+    case postHireStatuses.postSecondaryEducationCompleted:
+      return `Graduation Completed on - ${data.graduationDate}`;
+    case postHireStatuses.failedCohort:
+      return `Failed/incomplete course.`;
+    default:
+      return `Status not recorded`;
   }
-  return `Status not recorded`;
 };
 
 export const getGraduationStatus = (statuses = []) => {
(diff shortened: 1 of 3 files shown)


server/services/participants.js
```diff
@@ -148,6 +155,17 @@ const updateParticipant = async (participantInfo) => {
...
       { history: participantInfo.history || [], userUpdatedAt: new Date().toJSON() }
     );
+    if (changes.interested === 'withdrawn') {
+      const cohort = await getAssignCohort({ participantId: participantInfo.id });
+      // ensure that a participant has a cohort before adding post hire status
+      if (cohort) {
+        await createPostHireStatus({
+          participantId: participantInfo.id,
+          status: postHireStatuses.failedCohort,
+          data: {},
```
fwpushan, line 165: I would recommend store cohortId in data for future reference.

fwpushan (changes requested):

client/src/services/participant.js
```diff
@@ -6,11 +6,15 @@ const getCohortName = (cohort = {}) =>
...
 
-export const getPostHireStatusLabel = ({ status, data = {} } = {}) => {
-  if (status === postHireStatuses.postSecondaryEducationCompleted) {
-    return `Graduation Completed on - ${data.graduationDate}`;
+const getPostHireStatusLabel = ({ status, data = {} } = {}) => {
+  switch (status) {
+    case postHireStatuses.postSecondaryEducationCompleted:
+      return `Graduation Completed on - ${data.graduationDate}`;
+    case postHireStatuses.failedCohort:
+      return `Failed/incomplete course.`;
```
fwkendall, line 14: Given we pulled in the wording change ticket, should this read "Unsuccessful" instead?

server/services/participants.js
fwkendall: I think it makes sense not to delete the entry, since we'll want that historical info to show on the participant and cohort details in future

Not sure about storing the cohortId, that seems to make sense to me to do - @dbayly-freshworks thoughts?

dbayly-freshworks: If we delete the user from the cohort table, then a slot will open up. Space in cohorts is calculated from the cohorts total minus the number of users in that cohort. Saving the Cohort ID makes sense to me so we have a record of what happened. 

It seems like the system we have to keep track of cohort size isn't serving our needs. @fwkendall  Should we refactor it to add a `remaining allocation` column in the database instead of counting the number of remaining users.

fwkendall: It sounds like what we have now will work, as long as we don't add the deletion

Do you think it's worth a refactor now? Or hold off for a bit and see if we need more flexibility?

dbayly-freshworks: I'm on the fence to be honest. I think let's keep the function as is, but be ready to refactor if requirements change.

PR #644, opened 2022-02-24
HCAP-819 | Web-API | Add new patch method to update cohort

fwpushan:
-------
- [x] Added new service method
- [x] Update existing code
- [x] Update tests

### DIFF
--- a/server/routes/cohorts.js
+++ b/server/routes/cohorts.js
@@ -2,8 +2,16 @@ const express = require('express');
 const keycloak = require('../keycloak');
 const logger = require('../logger.js');
 const { asyncMiddleware, applyMiddleware } = require('../error-handler.js');
-const { getCohorts, getCohort, assignCohort, getAssignCohort } = require('../services/cohorts.js');
+const {
+  getCohorts,
+  getCohort,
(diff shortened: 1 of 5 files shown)


server/routes/cohorts.js
```diff
@@ -121,4 +129,42 @@ router.get(
...
+  asyncMiddleware(async (req, res) => {
+    const { email, user_id: userId, sub: localUserId } = req.user;
+    if (email && (userId || localUserId)) {
+      const { id } = req.params;
+      const { body } = req;
+      if (!id) {
+        return res.status(400).send({ message: 'Missing cohort id' });
+      }
+      // Get Cohort
+      const [cohort] = await getCohort(+id);
```
dbayly-freshworks, line 143: What does the plus here do?

dbayly-freshworks (commented):
One comment, otherwise looks pretty good.

server/routes/cohorts.js
fwpushan: converting string to number .... actually req.params. is key-value : string: string

dbayly-freshworks (approved):

PR #645, opened 2022-02-24
HCAP-819 | Web | Edit Cohort 

fwpushan:
----
- Refactor UI to sync with other views
- Integrate service layer
- Update Cohort form to support Edit
- Add success message for add/edit cohort.

### DIFF
--- a/client/src/pages/private/PSIViewDetails.js
+++ b/client/src/pages/private/PSIViewDetails.js
@@ -1,18 +1,30 @@
 import React, { lazy, useEffect, useState, useMemo } from 'react';
 import { Card, Page, CheckPermissions, Dialog } from '../../components/generic';
 import Button from '@material-ui/core/Button';
-import { Box, Grid, Link, Typography } from '@material-ui/core';
+import { Box, Grid, Link, Typography, makeStyles } from '@material-ui/core';
 import { scrollUp } from '../../utils';
-import store from 'store';
 import routes from '../../constants/routes';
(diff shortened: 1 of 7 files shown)


client/src/pages/private/PSIViewDetails.js
```diff
@@ -54,74 +75,76 @@ export default ({ match }) => {
...
   return (
-    <>
+    <Page isAutoHeight={true}>
       <Dialog
         title={
-          activeModalForm === 'edit-psi' ? `Edit PSI (${psi.instituteName})` : `Create New Cohort`
+          activeModalForm === 'edit-psi'
+            ? `Edit PSI (${psi.instituteName})`
+            : cohort
+            ? `Edit Cohort (${cohort.cohort_name})`
```
fwkendall, line 146: minor nitpick not worth worrying about, but in future I think it would be helpful if we indent nested cases further for clarity :)

fwkendall (approved):

client/src/services/psi-cohort.js
```diff
@@ -1,4 +1,5 @@
 import store from 'store';
+import moment from 'moment';
```
fwkendall, line 2: might want dayjs here instead?

fwpushan: Currently front end is using moment and backend dayjs, one of the first thing in my tech debt backlog is use dayjs for front end

PR #657, opened 2022-03-16
HCAP-1083 | Create UI integration layer with Web API

fwpushan:


client/src/constants/validation.js
```diff
@@ -689,3 +690,22 @@ export const ParticipantAssignCohortSchema = yup
     cohort: yup.number().required(errorMessage),
     institute: yup.number().required(errorMessage),
   });
+
+export const ReturnOfServiceSchema = yup
```
sunghwan-park-fw, line 694: don't we need to do `Past or present date validation`

client/src/services/return-of-service.js
```diff
@@ -0,0 +1,29 @@
...
+
+export const createReturnOfServiceStatus = async ({ participantId, data }) => {
+  // Covert data into date obj
+  const dateObj = moment(data.date, 'YYYY/MM/DD').toDate();
+  const finalBody = {
+    ...data,
+    date: dateObj,
+  };
+  const url = `${API_URL}/api/v1/ros/participant/${participantId}`;
+  const response = await fetch(url, {
```
sunghwan-park-fw, line 13: just curious. why don't we use axios?

sunghwan-park-fw (commented):
It's hard to review because I don't have access to dev site and can't run the app locally.

client/src/services/return-of-service.js
fwpushan: Legacy

client/src/constants/validation.js
fwpushan: Yes

fwpushan: Very good catch I will work on this ....

fwpushan: This may affect all our existing app validations ...

PR #665, opened 2022-03-28
HCAP-1116 | Add CLI logic to invoke service

fwpushan:


server/scripts/services/change-cohort.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  console.log(`Found participant with ${participantId}`);
+
+  try {
+    await changeCohortParticipant({
+      cohortId: oldCohortDbObj.id,
+      participantId,
+      newCohortId: newCohortDbObj.id,
+      meta: {
+        user: options.user || 'system',
```
sunghwan-park-fw, line 48: How about adding user option to the usage?

server/scripts/services/change-cohort.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+    });
+    return { success: true, message: 'Success', usageMessage: '' };
+  } catch (error) {
+    return { success: false, message: `Error: ${error}`, usageMessage: '' };
+  }
+};
+
+module.exports = {
+  changeCohort: serviceMethod,
```
sunghwan-park-fw, line 59: why do we need a different name?

sunghwan-park-fw (approved):

server/scripts/services/change-cohort.js
fwpushan: no need I update it in next PR

server/scripts/services/change-cohort.js
fwpushan: this is optional but I will add

PR #666, opened 2022-03-28
HCAP-1116 | Pod | Create resource to execute service req in OpenShift

fwpushan:
- [x] OpenShift Pod template and invoke code
- [x] Github workflow to execute service code

### DIFF
--- a/.github/workflows/service.run.prod.yml
+++ b/.github/workflows/service.run.prod.yml
@@ -1,0 +1,37 @@
+name: Create and execute Service Pod - Prod
+
+on:
+  workflow_dispatch:
+    inputs:
+      IMAGE_TAG:
+        description: 'Image tag'
+        required: true
+        default: latest
+      CONFIG:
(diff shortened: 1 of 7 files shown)


.github/workflows/service.run.prod.yml
```diff
@@ -0,0 +1,37 @@
...
+
+env:
+  CLUSTER: https://api.silver.devops.gov.bc.ca:6443
+  AUTH_TOKEN: ${{ secrets.AUTH_TOKEN_OCP4 }}
+  SA_TOKEN: ${{ secrets.SA_TOKEN_CONFIG }}
+  PROJECT: hcap
+
+jobs:
+  # Dev
+  service-exe-dev:
```
sunghwan-park-fw, line 22: Couldn't the job name be just `service-exe`?

sunghwan-park-fw (approved):
I used OCP once but don't have deep understanding enough to review IaC on OCP.

PR #672, opened 2022-03-30
HCAP-1116 | Read Me

fwpushan:


akhilfw (commented):
19 +- Select [Env] - Service Pod action from left panel

May be we can add examples here like Test - Service Pod or Prod - Service Pod

akhilfw (commented):
Also we could mention that we should not pass rich text when we use JSON

Same is the case with line 49 - if we can give examples

sunghwan-park-fw (approved):

akhilfw (approved):

PR #673, opened 2022-03-31
HCAP-1143 | Show User Info |  Store user details in db

fwpushan:
- [x] Create Migration and sync service
- [x] Update `/approve-request` and `/user-details`  route to store user details from KC

### DIFF
--- a/server/tests/util/db.js
+++ b/server/tests/util/db.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-console, no-restricted-syntax, no-await-in-loop */
 const { dbClient, collections } = require('../../db');
+const keycloak = require('../../keycloak');
 
 async function clearDB() {
   for (const value of Object.values(collections)) {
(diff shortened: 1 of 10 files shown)


server/tests/util/db.js
```diff
@@ -14,6 +15,7 @@ async function clearDB() {
 async function startDB() {
   await dbClient.connect();
   await clearDB();
+  await keycloak.buildInternalIdMap();
```
sunghwan-park-fw, line 18: It's not in your changes but just curiosity. Why are there two lines of authenticateIfNeed in buildInternalIdMap method?

sunghwan-park-fw (approved):

server/tests/util/db.js
fwpushan: KeyCloak Manager used a service account to interact with KeyCloak server. 
`authenticateIfNeed`  => 
    1. Check expiry of auth token 
    2. Renew auth token if expired.

PR #675, opened 2022-04-04
HCAP-1143 | HCAP-1152 | Sort Employer name column

fwpushan:
- [x] Fix and tests

### DIFF
--- a/server/services/participants-helper.js
+++ b/server/services/participants-helper.js
@@ -161,6 +161,20 @@ class FilteredParticipantsFinder {
         field: joinFieldName,
         direction: pagination.direction || 'asc',
       });
+
+      // To manage employer name column sorting we need to sort by employer name
+      if (sortField === 'employerName') {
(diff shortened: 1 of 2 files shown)


server/services/participants-helper.js
```diff
@@ -161,6 +161,20 @@ class FilteredParticipantsFinder {
         field: joinFieldName,
         direction: pagination.direction || 'asc',
       });
+
+      // To manage employer name column sorting we need to sort by employer name
+      if (sortField === 'employerName') {
```
dbayly-freshworks, line 166: Nit - Might be good to reference a constant/enum here.

dbayly-freshworks (approved):
One nit, otherwise looks good.

server/services/participants-helper.js
fwpushan: yup whole function need to refactor regarding const name ....

PR #677, opened 2022-04-04
Refactoring validation.js

taraepp:
to do: divide schema into smaller files


fwpushan (changes requested):
Remove Schema.js it is causing lots of code duplication

fwpushan (changes requested):
Why this part of the code is added with refactor validation?

fwpushan (approved):

PR #678, opened 2022-04-06
HCAP-1156 | Archive with Multi Org Support

fwpushan:
-----
- [x] Fix archive with muti org support

### DIFF
--- a/server/services/participants.js
+++ b/server/services/participants.js
@@ -34,7 +34,7 @@ const deleteParticipant = async ({ email }) => {
   });
 };
 
-const deleteAcknowledgement = async (participantId) => {
+const deleteAcknowledgement = async (participantId) =>
   dbClient.db.withTransaction(async (tx) => {
     const item = await tx[collections.PARTICIPANTS_STATUS].findOne({
       participant_id: participantId,
@@ -42,7 +42,7 @@ const deleteAcknowledgement = async (participantId) => {
       current: true,
     });
     if (!item) {
-      return {};
+      return { success: false, message: 'No pending acknowledgement found' };
     }
     await tx[collections.PARTICIPANTS_STATUS].update(
       {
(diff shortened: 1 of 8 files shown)


server/services/participants.js
```diff
@@ -300,8 +300,25 @@ const setParticipantStatus = async (
       item?.status !== 'archived'
     )
       return { status: 'invalid_status_transition' };
-    if (status === 'archived' && !item) {
-      return { status: 'Could not find participant' };
+
+    // Handling Hired Status
+    const hiredStatus = items[0];
+    if (
+      ['archived'].includes(status) &&
```
dbayly-freshworks, line 307: Since this is an array of 1, can we simplify this? status === 'archived' ?

server/services/participants.js
```diff
@@ -300,8 +300,25 @@ const setParticipantStatus = async (
...
+      (!item || (hiredStatus && hiredStatus.employer_id !== employerId))
+    ) {
+      if (hiredStatus.data.site && user?.sites.includes(hiredStatus.data.site)) {
```
dbayly-freshworks, line 310: Maybe make this null safe?

server/services/participants.js
```diff
@@ -531,19 +548,43 @@ const getParticipants = async (
         (statusInfo) => statusInfo.status === 'hired' && !user.sites.includes(statusInfo.data.site)
       );
 
+      const hiredBySomeoneInSameOrgStatus = item.statusInfos?.find(
```
dbayly-freshworks, line 551: More a nit than anything, but the function that maps the data in the return statement for this function is 50+ lines long. We should flag this for refactoring. Also all of this business logic is very hard to parse. 

I would recommend adding some more comments explaining exactly what's going on here.

dbayly-freshworks (approved):
Overall ok, but I think participants.js is in dire need of refactoring.

dbayly-freshworks (dismissed):
Looks ok to me.

dbayly-freshworks (approved):
This is a lot, but it all looks ok

PR #679, opened 2022-04-07
HCAP-1155: minor style updates to Typography

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1155

🐛 **bug description**: long names are not truncated when displayed on employer portal causing UI issues

🔧 **possible fixes**: 
1. Add `maxWidth` to `Typography` and display the rest as `...`
❓ do we want to make it consistent across the application in this case? Possibly create a new component?
2. Specify `maxLength` to user `auth.user.name` instead

This PR implements 1st option
Open for discussion 🙂

### DIFF
--- a/client/src/pages/private/Admin.js
+++ b/client/src/pages/private/Admin.js
@@ -11,6 +11,11 @@ const useStyles = makeStyles((theme) => ({
   button: {
     marginTop: theme.spacing(3),
   },
+  welcomeOverflow: {
+    maxWidth: '250px',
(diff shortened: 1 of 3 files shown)


fwpushan (changes requested):
This is fine but we need to address the same issue in `ParticipantTable` > `Hired Participant`  (tab)> `Hired by` (column)

fwpushan (approved):
LGTM

PR #686, opened 2022-04-21
HCAP-1162 | Fix header alignment

taraepp:
Description of problem (and chosen solution): https://freshworks.atlassian.net/browse/HCAP-1162

Put horizontal scroll on overflowing container element

### DIFF
--- a/client/src/pages/private/ParticipantView.js
+++ b/client/src/pages/private/ParticipantView.js
@@ -1,16 +1,24 @@
 import React, { lazy, useState } from 'react';
 import Grid from '@material-ui/core/Grid';
+import { makeStyles } from '@material-ui/core/styles';
 import { Page, CheckPermissions, CustomTab, CustomTabs } from '../../components/generic';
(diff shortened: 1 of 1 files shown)


client/src/pages/private/ParticipantView.js
```diff
@@ -30,7 +30,13 @@ export default () => {
             )}
           </CustomTabs>
         </Grid>
-        <Grid container alignItems='center' justify='flex-start' direction='column'>
+        <Grid
+          container
+          alignItems='center'
+          justify='flex-start'
+          direction='column'
+          style={{ overflowX: 'scroll' }}
```
fwpushan, line 38: try to use class instead of inline-style

fwpushan (commented):
Please check my comments

client/src/pages/private/ParticipantView.js
taraepp: That makes sense! I'll update that.

fwpushan (approved):
LGTM 👍

PR #689, opened 2022-04-21
HCAP-1155: adding maxLength for employers name

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1155

Just increasing max character length for names


taraepp (approved):
Looks good to me

PR #691, opened 2022-04-22
HCAP-1175 | Multiselect anchoring

taraepp:
https://freshworks.atlassian.net/jira/software/c/projects/HCAP/boards/192?selectedIssue=HCAP-1107

- Found that the issue we were experiencing with the select was a known issue with the component: https://github.com/mui/material-ui/issues/19977

- Solution referenced in issue worked, and covers in detail why it happens (TLDR: it recalculates the anchor point when you select something new): https://stackoverflow.com/questions/59785482/multiselect-box-popover-keeps-jumping-when-scroll-or-select-items/59790471#59790471


alva-fresh (approved):
Thank you for digging into and fixing it!

PR #696, opened 2022-04-26
HCAP-1106 | Show last engaged by column

taraepp:
https://freshworks.atlassian.net/browse/HCAP-1106?atlOrigin=eyJpIjoiYjJlYmE0MDU0ZTQ4NGZiNDgwYTAzODI0NGZhZjZmODMiLCJwIjoiaiJ9

Last Engaged By column is between Last Updated and Actions columns.

Not yet sortable

### DIFF
--- a/client/src/constants/participantTableConstants.js
+++ b/client/src/constants/participantTableConstants.js
@@ -162,6 +162,7 @@ const columns = {
   rosStartDate: { id: 'rosStartDate', name: 'Return of Service Start Date', sortOrder: 22 },
   rosSiteName: { id: 'rosSiteName', name: 'Site Name', sortOrder: 23 },
   employerName: { id: 'employerName', name: 'Hired By', sortOrder: 24 },
+  engagedBy: { id: 'engagedBy', name: 'Last Engaged By', sortOrder: 15.5 },
 };
 
 const {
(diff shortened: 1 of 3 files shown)


client/src/constants/participantTableConstants.js
```diff
@@ -162,6 +162,7 @@ const columns = {
   rosStartDate: { id: 'rosStartDate', name: 'Return of Service Start Date', sortOrder: 22 },
   rosSiteName: { id: 'rosSiteName', name: 'Site Name', sortOrder: 23 },
   employerName: { id: 'employerName', name: 'Hired By', sortOrder: 24 },
+  engagedBy: { id: 'engagedBy', name: 'Last Engaged By', sortOrder: 15.5 },
```
alva-fresh, line 165: What's `sortOrder` used for? I'm curious why it's so specific to 15.5

dbayly-freshworks: Sort order determines which columns to show from left to right.  We should stick to whole numbers when using it, and just shift the others down by one.

taraepp: Good point, fixed that!

alva-fresh (approved):
Looks good to me!

PR #700, opened 2022-04-26
HCAP-1106 | Make Last Engaged By column sortable

taraepp:
https://freshworks.atlassian.net/browse/HCAP-1106
- This is the second part of this story, the column has been added previously, now it is sortable.
- On first click, it sorts reverse alphabetical (2 clicks for alphabetical). However, I noticed that the Last Name column does as well.


fwpushan (approved):
Perfect

PR #702, opened 2022-04-28
HCAP-701: fix MoH participant 'Open' status

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-701

This PR fixes the issue with statuses being displayed as `Open` for MoH. Still, I feel like there're a few more tests needed
I'd appreciate any input at this point 🙌 

I'm adding some refactoring as well 🖌️ 

**My questions:** ❔ 
1. I noticed there's a field we're generating on BE called `statusInfo`. Is it used somewhere?
2. When HA archives a participant, they have an option to select: _Withdrawn at participant request_. If they select this option - should the status change to `rejected` or `withdrawn`?
3. What should be the status for archived participants assuming they're not available to any employers?

Screen Shot 2022-04-29 at 11 00 27 AM
Screen Shot 2022-04-29 at 11 07 10 AM

### DIFF
--- a/client/src/utils/prettifyStatus.js
+++ b/client/src/utils/prettifyStatus.js
@@ -14,10 +14,9 @@ const getParticipantStatus = (isMoH, status) => {
   if (status === 'rejected') return 'Archived';
   if (status === 'ros') return 'Return of Service';
 
-  if (isMoH) {
-    if (status === 'prospecting') return 'In Progress';
-    if (status === 'interviewing') return 'In Progress (2)';
-    if (status === 'offer_made') return 'In Progress (3)';
+  if (isMoH && status.startsWith('inprogress')) {
+    const count = status.split('_');
+    return `In Progress (${count[1]})`;
   }
 
   if (status === 'offer_made') return 'Offer Made';
--- a/client/src/utils/prettifyStatus.js
+++ b/client/src/utils/prettifyStatus.js
@@ -3,6 +3,13 @@ import InfoIcon from '@material-ui/icons/Info';
 import { ComponentTooltip } from '../components/generic/ComponentTooltip';
 import { Button } from '../components/generic';
 
+/**
+ * Returns participant stats message based on the first status provided
+ *
+ * @param {boolean} isMoH - is this view displayed for MoH user
+ * @param {string} status - first status defined in row.status
+ * @returns {string} specific status message based on value / capitalized status
+ */
 const getParticipantStatus = (isMoH, status) => {
   if (status === 'rejected') return 'Archived';
   if (status === 'ros') return 'Return of Service';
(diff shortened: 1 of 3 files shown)


fwpushan (changes requested):
LGTM. Just only one comment.
Also, you have to update backed   
In server/services/participant.js line 622.
The withdrawn is determined by 
`['no', 'withdrawn'].includes(item.interested)`

fwkendall:
> 2\. When HA archives a participant, they have an option to select: _Withdrawn at participant request_. If they select this option - should the status change to `rejected` or `withdrawn`?

To expand on Pushan's reply, there's a difference between the status that a participant has with any given user and their underlying status. When an HA/employer is archiving a participant, that's the user-specific status. So the participant is "rejected" in the BE, "archived" in the FE (because the original terminology was determined to be too harsh after initial implementation). They are *not* withdrawn from the program, regardless of the reason selected for archiving/rejecting. *Withdrawn* is a participant-specific status that would be seen by all users who can view the participant - it means no employers can further interact with them as they've withdrawn from the HCAP program entirely.

client/src/utils/prettifyStatus.js
```diff
@@ -3,16 +3,40 @@ import InfoIcon from '@material-ui/icons/Info';
...
-  let firstStatus = status[0];
+/**
+ * Returns participant stats message based on the first status provided
+ *
+ * @param {boolean} isMoH - is this view displayed for MoH user
+ * @param {string} status - first status defined in row.status
+ * @returns {string} specific status message based on value / capitalized status
+ */
+const getParticipantStatus = (isMoH, status) => {
+  if (status === 'rejected') return 'Archived';
```
fwpushan, line 14: `rejected` status is not associated with MoH, consider it `Available` here.

client/src/utils/prettifyStatus.js
```diff
@@ -3,16 +3,40 @@ import InfoIcon from '@material-ui/icons/Info';
...
+  if (status === 'ros') return 'Return of Service';
+
+  if (isMoH) {
+    if (status === 'prospecting') return 'In Progress';
+    if (status === 'interviewing') return 'In Progress (2)';
```
fwpushan, line 19: In Progress (#num) signifies the number of such statuses. e.g. The participant is prospected by two or many users.

fwpushan (changes requested):
Please check my comments.

client/src/utils/prettifyStatus.js
alva-fresh: This is not for MoH. This function returns status for all users. MoH specific logic is captured under `if (MoH)` clause

client/src/utils/prettifyStatus.js
alva-fresh: Thanks! I'll update that

client/src/utils/prettifyStatus.js
fwpushan: perfect

fwpushan (approved):
Good job 👍

PR #710, opened 2022-05-03
HCAP-1106 | Add date to last engaged by column

taraepp:
- date added
- format according to MMM DD, YYYY
- made classes for the cell text in participant table
- added that darker grey colour to theme palette
- did not change the sorting (sorting by date not desired, already sorting by name)

Could have instead used Typography instead of classes? Not sure if we like div but it gets the job done.

Updated:
- investigated using Typography instead: v4 of MUI does not support custom variants, the feature was added in v5. None of our existing variants matched the design (all bold variants were much larger or the button variant has a text-transform). Did not want to modify variants used elsewhere.
- changed div to Box
- renamed that class to Alva's suggestion
- did not follow the design _exactly_ as it had the staff name text larger than the date. The design also had other larger, bold text (column headers) that the implementation currently lacks. It looked out of place.

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -3,6 +3,7 @@ import { useHistory } from 'react-router-dom';
 
 import Grid from '@material-ui/core/Grid';
 import { Box, Menu, MenuItem, Link } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
 
 import {
   ToastStatus,
@@ -149,6 +150,11 @@ const filterData = (data, columns, isMoH = false) => {
         statusWithEmployerDetails[0].employerInfo.firstName,
         maxStrLen
       )} ${addEllipsisMask(statusWithEmployerDetails[0].employerInfo.lastName, maxStrLen)}`;
+
+      const createdAtFormatted = dayUtils(statusWithEmployerDetails[0].createdAt).format(
+        'MMM DD, YYYY'
+      );
+      row.engagedLast = createdAtFormatted;
       row.engagedBy = row.employerName;
     }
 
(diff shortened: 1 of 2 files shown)


client/src/pages/private/ParticipantTable.js
```diff
@@ -401,6 +417,13 @@ const ParticipantTable = () => {
         );
       case 'postHireStatuses':
         return getGraduationStatus(row[columnId] || []);
+      case 'engagedBy':
+        return (
+          <>
+            <div className={classes.cellTextStrong}>{row.engagedBy}</div>
```
alva-fresh, line 423: Can we replace _div_ either with `Box` or `Typography`? I feel like for `Typography` we already have some predefined styles, so there's no need to introduce `cellTextStrong`

client/src/pages/private/ParticipantTable.js
```diff
@@ -190,6 +196,16 @@ const ParticipantTable = () => {
   const isAdmin = isMoH || isSuperUser;
   const isEmployer = roles.includes('health_authority') || roles.includes('employer');
 
+  const useStyles = makeStyles((theme) => ({
+    cellTextStrong: {
+      fontWeight: 'bold',
+    },
+    cellTextWeak: {
```
alva-fresh, line 203: nit: maybe `cellTextGray` sounds beeter?

taraepp: Much better! Thanks for the suggestion, I didn't like like what I had either.

client/src/pages/private/ParticipantTable.js
taraepp: My only concern with Typography is that it won't introduce a line break, but I'll do some experimenting with Typography and Box to confirm and find something that works- I'll get rid of the div.

fwpushan (commented):
1. I don't see any column config here
2. what about sort logic?

alva-fresh (approved):
Looks good!

fwpushan (approved):

PR #714, opened 2022-05-04
HCAP-1190: updating SelectProspectingSite to single select

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1190

Changing `SelectProspectingSite` form to allow pick only 1 site during prospecting for both `Engage` and `Bulk Engage`

Screen Shot 2022-05-04 at 4 11 06 PM
Screen Shot 2022-05-04 at 4 10 44 PM
Screen Shot 2022-05-04 at 4 11 00 PM
Screen Shot 2022-05-04 at 4 10 29 PM

### DIFF
--- a/client/src/pages/private/ParticipantTableDialogues.js
+++ b/client/src/pages/private/ParticipantTableDialogues.js
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 import {
   InterviewingFormSchema,
   RejectedFormSchema,
-  ProspectingSitesSchema,
+  ProspectingSiteSchema,
   participantStatus,
 } from '../../constants';
 import { Dialog } from '../../components/generic';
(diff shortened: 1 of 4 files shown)


client/src/pages/private/ParticipantTableDialogues.js
```diff
@@ -69,8 +69,8 @@ export const ParticipantTableDialogues = ({
         <SelectProspectingSiteForm
           isMultiSelect
           selected={bulkParticipants}
-          initialValues={{ prospectingSites: [] }}
-          validationSchema={ProspectingSitesSchema}
+          initialValues={{ prospectingSite: undefined }}
```
fwpushan, line 72: same as above

client/src/pages/private/ParticipantTableDialogues.js
```diff
@@ -56,8 +56,8 @@ export const ParticipantTableDialogues = ({
     >
       {activeModalForm === 'single-select-site' && (
         <SelectProspectingSiteForm
-          initialValues={{ prospectingSites: [] }}
-          validationSchema={ProspectingSitesSchema}
+          initialValues={{ prospectingSite: undefined }}
```
fwpushan, line 59:
```suggestion
          initialValues={{ prospectingSite: '' }}
```
`prospectingSite` is a string value, so please initiate with an empty string.

fwpushan (commented):
All good, just one single comment to address

client/src/pages/private/ParticipantTableDialogues.js
alva-fresh: `prospectingSite` is a number, form saves siteId as a value

fwpushan: Then use negative numbers and update the schema so it can only accept positive values .....

fwpushan: ok, then no change required

fwpushan (approved):

PR #716, opened 2022-05-06
HCAP-1140 | Enable feature flag | Update backend to support single site prospecting

fwpushan:
- [x] Update service and rest api to support saving of single-site prospecting 
- [x] Add and update tests
- [x] Update frontend


taraepp (approved):
Looks good to me. 
However, I agree with the bot that the participant status function needs refactoring (but was only picked up in this PR as it was moved from one file to another- makes more sense as separate ticket/PR)

PR #721, opened 2022-05-16
HCAP-921 |  Positions overview count

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-921

In the table's handleArchive I took out the call to forceReload and replaced it with a call to fetch the site's data and dispatch the site's new status through context provider. This was the only reference to forceReload, so I got rid of it, and the onArchiveParticipantAction prop. Stale wasn't used anymore either so I got rid of it, and the 2/3 extremely similar useEffect calls. Was getting dependency complaints from React compiler, got rid of those two (by supplying the dependencies).

### DIFF
--- a/client/src/pages/private/SiteParticipantsTable.js
+++ b/client/src/pages/private/SiteParticipantsTable.js
@@ -124,6 +124,7 @@ export default ({ id, siteId }) => {
   } = SiteDetailTabContext.useTabContext();
 
   const [orderBy, setOrderBy] = useState(columns[4]?.id || 'participantName');
+
   const handleRequestSort = (event, property) => {
     const isAsc = orderBy === property && order === 'asc';
     setOrder(isAsc ? 'desc' : 'asc');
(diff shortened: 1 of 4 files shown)


client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -328,68 +296,33 @@ export default ({ id, siteId, onArchiveParticipantAction, stale, setStale }) =>
...
+                      return (
+                        <Link
+                          component='button'
+                          variant='body2'
+                          onClick={() => participantOnClick(row.participantId)}
+                        >
+                          {row[columnId]}
+                        </Link>
+                      );
+                    } else {
```
alva-fresh, line 311: Probably no need for `else` here

client/src/pages/private/SiteParticipantsTable.js
```diff
@@ -98,138 +170,25 @@ export default ({ id, siteId, onArchiveParticipantAction, stale, setStale }) =>
...
-  const [orderBy, setOrderBy] = useState(columns[4]?.id || 'participantName');
-  const handleRequestSort = (event, property) => {
-    const isAsc = orderBy === property && order === 'asc';
-    setOrder(isAsc ? 'desc' : 'asc');
-    setOrderBy(property);
-  };
-
-  const sort = (array) => _orderBy(array, [orderBy, 'operatorName'], [order]);
+  useEffect(() => {
+    setRows(selectedTab === 'Hired Participants' ? fetchedHiredRows : fetchedWithdrawnRows);
```
fwpushan, line 181: what about the `Site Details` page?

fwpushan (commented):
All good. Please check my comments for some minor concerns.

client/src/pages/private/SiteParticipantsTable.js
taraepp: Fixed!

client/src/pages/private/SiteParticipantsTable.js
taraepp: Replaced with switch statement in last commit. Refactored a bit to use constants so I wasn't adding _more_ usage of string literals.

fwpushan (approved):
👍

PR #724, opened 2022-05-19
HCAP-1191 | TryFix | Security alerts | Fix tagging alias issue in Makefile

fwpushan:
- [x] Fix critical security alert 
- [x] Try-fix other alerts 
- [x] Fix tagging issue for test and prod

### DIFF
--- a/server/keycloak.js
+++ b/server/keycloak.js
@@ -260,6 +260,11 @@ class Keycloak {
     }
   }
 
+  getUserUrl(userId = '') {
+    if (!userId) throw new Error('keycloak: User ID is required');
+    return `${this.authUrl}/admin/realms/${this.realm}/users/${userId}`;
+  }
+
   async setUserRoles(userId, role, regions) {
     try {
       if (!Object.keys(this.roleIdMap).includes(role)) throw Error(`Invalid role: ${role}`);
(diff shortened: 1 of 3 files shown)


server/keycloak.js
```diff
@@ -260,6 +260,11 @@ class Keycloak {
     }
   }
 
+  getUserUrl(userId = '') {
+    if (!userId) throw new Error('keycloak: User ID is required');
```
alva-fresh, line 264: If `userId` defaults to `''`, doesn't it mean that this condition will always be false?

server/routes/participant-user.js
```diff
@@ -163,7 +163,7 @@ router.post(
         res.status(422).send('Already Hired or Withdrawn');
       }
     } else {
-      res.status(422).send(`No expression of interest with id: ${id}`);
+      res.status(422).send(`No expression of interest with participant id`);
```
alva-fresh, line 166: Can we rephrase this error message a bit maybe? It sounds quite ambiguous

server/keycloak.js
fwpushan: No. Defaulted means here if no arg is supplied it will take userId as an empty string and throw an error ...

server/routes/participant-user.js
fwpushan: what do you suggest?

server/keycloak.js
```diff
@@ -260,6 +260,11 @@ class Keycloak {
...
+    return `${this.authUrl}/admin/realms/${this.realm}/users/${userId}`;
```
taraepp, line 265: I think we're still performing requests from user supplied data here, but putting it in a function is good to manage this scenario consistently. Is there additional validation that we can do on the userId? Or would that just open up opportunities for SQL injection?

taraepp (approved):

server/keycloak.js
fwpushan: actual data is not user-supplied. It is from DB. How do I explain this to the scanner?

fwpushan: And this is not an SQL query .... this is KeyCloak URL which has sufficient protection

server/routes/participant-user.js
alva-fresh: There's also the same issue on line 135

alva-fresh: Maybe something like _No expression of interest found for this participant_?

PR #727, opened 2022-05-24
HCAP 1214 | ROS end date is one day less if the start date is 29th Feb(Leap Year)

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1214
- I added a function addYearToDate that checks for the date parameter being Feb 29, and adjusts the return accordingly if it is (I did not use the isLeapYear plugin because Feb 29 is always a leap year, and even if it's a leap year I need to check that it's Feb 29th)
- duplicated the function on the client
- called that function instead of dayjs add function where I was able to find it
- outlined what was expected with given inputs in unit tests

Some issues with this that I'd like to discuss!
- on client, utils/date is its own file, on server I only added one function so I put it in the existing utils file.
- in ArchiveHiredParticipantForm (client), I'm importing dayjs and addYearToDate- the date utils file _also_ exports dayjs with additional plugins as dayUtils. I'm thinking that changing dayjs to dayUtils in cases like this may be more efficient?
- I wrote unit tests! But when I tried to create a sibling directory to integration called unit, my tests didn't run. Where is this configured??
- Could not unit test client, only server (closest I got was return from function being 'undefined', otherwise import errors).


cypress/integration/dateUtils.spec.js
```diff
@@ -0,0 +1,43 @@
+const { addYearToDate } = require('../../server/utils');
```
fwpushan, line 1: This is related to cypress / e2e tests, not the right place for unit tests. Remove this from here.

fwpushan (changes requested):
Please check my comments

fwpushan (approved):

PR #728, opened 2022-05-25
HCAP-1166 | Multi-org | Reject behaviour

fwpushan:
---
- [x] Create new alert for `rejected`
- [x] Refactor `setParticipantStatus` method
- [x] Create everlasting 'reject_ack' status until prospecting for same site


taraepp (changes requested):
LGTM. Would clean up the duplicate import caught by sonarcloud where capitalizedString is imported in ParticipantTable.

taraepp (approved):
👍

PR #733, opened 2022-05-27
HCAP-1084 | Backend | Add is_current col migration | Update service methods and rest api

fwpushan:
- [x] Update backend logic to accept siteId as param to create RoS
- [x] Add migrations to add is_current flag for RoS and select latest as the most current RoS update for the participant.
- [x]  Add and update tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,9 @@ app: ## Bash into App container
 	@echo "Make: Shelling into local application container"
 	@docker-compose -f docker-compose.dev.yml exec server /bin/bash
 
+run-local-db: local-build
+	@docker-compose -f docker-compose.dev.yml up postgres
+
 # Git Tagging Aliases
 
 tag-dev:
--- a/server/migrations/1653662298756_validate-latest-ros-participant-entry.js
+++ b/server/migrations/1653662298756_validate-latest-ros-participant-entry.js
@@ -1,0 +1,19 @@
+/* eslint-disable camelcase */
+const { dbClient, collections } = require('../db');
+
+exports.shorthands = 'validate-latest-ros-participant-entry';
+
+exports.up = async () => {
+  await dbClient.db.query(`
+    WITH participant_latest_ros as (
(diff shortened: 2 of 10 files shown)


server/migrations/1653662298756_validate-latest-ros-participant-entry.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable camelcase */
+const { dbClient, collections } = require('../db');
+
+exports.shorthands = 'validate-latest-ros-participant-entry';
```
dbayly-freshworks, line 4: Just out of curiosity : What does exports.shorthands do?

Makefile
```diff
@@ -116,6 +116,9 @@ app: ## Bash into App container
 	@echo "Make: Shelling into local application container"
 	@docker-compose -f docker-compose.dev.yml exec server /bin/bash
 
+run-local-db: local-build
+	@docker-compose -f docker-compose.dev.yml up postgres
```
dbayly-freshworks, line 120: What is the reason for adding a command launch just the db?

dbayly-freshworks (approved):
Looks ok to me, just a couple questions. 
Is a completion of ROS the only place the `is_current` flag will be set to false?
Is it possible for a participant to have multiple return of service records?

server/migrations/1653662298756_validate-latest-ros-participant-entry.js
fwpushan: `exports.shorthands` is autogenerated code ...so, I added the migration name here this will provide some more readable logs.

Makefile
fwpushan: To launch a standalone DB instance without running the whole application. Sometimes standalone DB instance is handy to pre-setup DB for migration.

fwpushan:
@dbayly-freshworks 
RoS (Return of Service) is like a probation period for HCAP hire.

dbayly-freshworks (approved):
Looks good!

PR #734, opened 2022-05-28
HCAP-1226: In Progress status for MoH

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1226

Adding some more checks to render `In Progress` statuses for MoH

The issue might not be in the codebase; need to check in prod


fwpushan (approved):

taraepp (approved):
👍

PR #735, opened 2022-06-02
HCAP-1084 | RoS | Different site case

fwpushan:
- [x] Update front end

### DIFF
--- a/client/src/components/modal-forms/ReturnOfServiceForm.js
+++ b/client/src/components/modal-forms/ReturnOfServiceForm.js
@@ -1,14 +1,20 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import Grid from '@material-ui/core/Grid';
 import MuiAlert from '@material-ui/lab/Alert';
 import { Button } from '../generic';
 import { Box, Typography } from '@material-ui/core';
-import { RenderDateField, RenderRadioGroup, RenderCheckbox } from '../fields';
+import {
+  RenderDateField,
+  RenderRadioGroup,
+  RenderCheckbox,
+  RenderSelectField,
+  RenderTextField,
+} from '../fields';
 import { Field, Formik, Form as FormikForm } from 'formik';
 import { ReturnOfServiceSchema, rosPositionType, rosEmploymentType } from '../../constants';
 import { getTodayDate } from '../../utils';
-import { createReturnOfServiceStatus } from '../../services';
+import { createReturnOfServiceStatus, getAllSites } from '../../services';
 
 // Helpers
 const rosPositionTypeOptions = Object.values(rosPositionType);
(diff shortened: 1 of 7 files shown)


server/routes/employer-sites.js
```diff
@@ -87,9 +87,10 @@ router.get(
     routeRedirect({ redirect: '/api/v1/employer-sites/details', match: 'employer-sites-detail' }),
   ],
   asyncMiddleware(async (req, res) => {
-    const { hcapUserInfo: user, params } = req;
-    const { all } = params;
+    const { hcapUserInfo: user, query } = req;
+    const { all } = query;
     let result;
+    console.dir(query);
```
alva-fresh, line 93: Is this `console` necessary here?

client/src/services/return-of-service.js
```diff
@@ -30,3 +31,21 @@ export const createReturnOfServiceStatus = async ({ participantId, data }) => {
...
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${store.get('TOKEN')}`,
+      Accept: 'application/json',
+      'Content-type': 'application/json',
+    },
+  });
+
+  if (response.ok) {
+    return await response.json();
```
alva-fresh, line 47: don't think we need `await` here

client/src/components/modal-forms/ReturnOfServiceForm.js
```diff
@@ -126,6 +169,25 @@ export const ReturnOfServiceForm = ({
...
+              {values.sameSite === false && (
+                <Box mt={3} mb={2}>
+                  <Field
+                    name='site'
+                    component={RenderSelectField}
+                    label='Select Site'
+                    options={mapToOptions(sites)}
+                    onChange={(event) => handleSiteSelection(event, setFieldValue)}
+                  />
+                  <br />
```
alva-fresh, line 181: Instead of adding `<br />`; I'd rather wrapped these Fields within its own `Box` components - that way it's easier to manage spacing between them

client/src/components/modal-forms/ReturnOfServiceForm.js
```diff
@@ -126,6 +169,25 @@ export const ReturnOfServiceForm = ({
...
+                  <Field
+                    name='ha'
+                    component={RenderTextField}
+                    label='Health Authority'
+                    disabled={true}
```
alva-fresh, line 186: Do we need to display this field if it's always disabled?

client/src/components/modal-forms/ReturnOfServiceForm.js
```diff
@@ -50,6 +75,8 @@ export const ReturnOfServiceForm = ({
     employmentType: '',
     sameSite: true,
     confirm: '',
+    site: null,
+    ha: null,
```
alva-fresh, line 79: `ha` is a TextField, so I'm wondering if default should be `''`

client/src/components/modal-forms/ReturnOfServiceForm.js
fwpushan: This is as per the design

alva-fresh (approved):

PR #736, opened 2022-06-06
HCAP-1216 | Cronjob |  Infra code

fwpushan:
- [x] Create a cronjob template and deployment code
- [x] Create script entry point of cronjob 
- [x] Attach service method and report logging with script entry-point


taraepp (approved):
👍

PR #746, opened 2022-06-17
HCAP-1233 | Return of service date in future

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1233

Testing things also added:
1. A cypress command to assign an array of sites to a user (should probably always be test-ha)
2. script to add table data from arbitrary tables. Currently they're hardcoded, but this could be expanded on in the future
3. Script relies on other pieces being run first: kc-import-users, and seed-participants- technically the sites and participants _could_ be included in this script, but the data sheets will need to be regenerated so that the columns in the spreadsheet match the columns in the table.


client/src/components/modal-forms/ReturnOfServiceForm.js
```diff
@@ -127,7 +126,6 @@ export const ReturnOfServiceForm = ({
                 name='date'
                 component={RenderDateField}
                 label='Return of Service Start Date'
-                maxDate={getTodayDate()}
```
arranfw, line 130: Should maxDate be a value in line with 
> and before January 1st 2100

server/package.json
```diff
@@ -13,6 +13,7 @@
     "migrate": "node-pg-migrate",
     "feed-participants": "node scripts/feed-participants.js",
     "feed-sites": "node scripts/feed-sites.js",
+    "feed-data": "node scripts/feed-data.js",
```
arranfw, line 16: Is this generic enough to be called `feed-data`? It looks like `post_secondary_institutions` is a hard value in the script

client/src/components/modal-forms/ReturnOfServiceForm.js
taraepp: You are absolutely right!

cypress/support/commands.js
```diff
@@ -200,3 +200,39 @@ Cypress.Commands.add('callAPI', ({ api, method = 'POST', body, followRedirect =
     });
   });
 });
+
+// assigns the site_ids to the username, assuming the region is already correct
+// does not un-assign previous sites
+Cypress.Commands.add('assignSitesToHealthAuthorityUser', (username, site_ids) => {
```
arranfw, line 206: Awesome

arranfw (approved):
I didn't look too closely at the seeding script as we have plans to improve and expand it in the future, otherwise the rest of the work looks great, awesome job with `assignSitesToHealthAuthorityUser`

fw-noel (approved):

arranfw (approved):

PR #758, opened 2022-06-29
HCAP-1237/bug-fix/1256 | Fix copy issue

fwpushan:
Change copy text

### DIFF
--- a/cypress/integration/psiDetailView.spec.js
+++ b/cypress/integration/psiDetailView.spec.js
@@ -3,7 +3,7 @@ describe('Tests the PSI View', () => {
   before(() => {
     cy.kcLogin('test-moh');
     cy.visit('/psi-view');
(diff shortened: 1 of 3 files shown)


cypress/integration/psiDetailView.spec.js
```diff
@@ -4,7 +4,7 @@ describe('Tests the PSI View', () => {
     cy.kcLogin('test-moh');
     cy.visit('/psi-view');
     cy.contains('Add Post Secondary Institution').click({ force: true });
-    cy.wait(1000);
+    cy.wait(2000);
```
alva-fresh, line 7: Do we need to increase the timeout? Tests are failing either way

fwpushan: Cypress test is failing and there is no code change related to this section.

alva-fresh (approved):

PR #767, opened 2022-07-19
HCAP-1267: Health Authority value select fix

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1267

Some tweaks to allow `healthAuthority` field auto-populate based on the Site selected

### DIFF
--- a/client/src/constants/validation/schema/schema-change-ros-site.js
+++ b/client/src/constants/validation/schema/schema-change-ros-site.js
@@ -8,6 +8,6 @@ export const ChangeRosSiteSchema = yup
     startDate: yup.date().required('Start Date at a New Site is required'),
     positionType: yup.string().required('Position Type is required').oneOf(rosPositionTypeValues),
     employmentType: yup.string().optional().oneOf(rosEmploymentTypeValues),
-    site: yup.number('Invalid type for New Site').required('New Site is required'),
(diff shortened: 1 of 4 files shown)


client/src/constants/validation/schema/schema-change-ros-site.js
```diff
@@ -8,6 +8,6 @@ export const ChangeRosSiteSchema = yup
     startDate: yup.date().required('Start Date at a New Site is required'),
     positionType: yup.string().required('Position Type is required').oneOf(rosPositionTypeValues),
     employmentType: yup.string().optional().oneOf(rosEmploymentTypeValues),
-    site: yup.number('Invalid type for New Site').required('New Site is required'),
+    site: yup.number('Invalid type for New Site').required('New Site Name is required'),
```
arranfw, line 11: It seems weird that this validation is `number` but we call it `Name` or is this just a quirk of it being a dropdown?

arranfw (approved):

client/src/constants/validation/schema/schema-change-ros-site.js
alva-fresh: Yes, so the value we're keeping and validating is `siteId` = _number_; but when selecting a site from dropdown, they only see site names.
So, we're only validating `value` and not `label`

PR #779, opened 2022-08-03
HCAP-1289 Don't force clicks

arranfw:
- I don't think we need to disable waiting for actionability here, this could be a reason why inputs are missed occasionally


fw-noel (dismissed):
I see it, if it doesn't wait for the add psi to be clicked, then it could be causing the issue.
Does it lengthen the test time? Can't we just remove force from the necessary ones?

arranfw:
> I see it, if it doesn't wait for the add psi to be clicked, then it could be causing the issue. Does it lengthen the test time? Can't we just remove force from the necessary ones?

The tests seemed to run just as quickly locally, I don't know that we can say any of these forces are necessary

fw-noel (approved):

PR #780, opened 2022-08-03
fix: disable edit button for withdrawn participants

fw-noel:
Rename enableEdit to a clearer name
Disable button if logged in participant is withdrawn


taraepp (approved):
👍

arranfw (approved):

PR #782, opened 2022-08-06
HCAP-1142: MoH Edit Participant Return of Service Info

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1142

Making a draft PR as it's going to be quite a big change 😬 

❓**What's included here:** 🎨 🖌️
- Some refactoring on the `ParticipantViewDetails` page
  - breaking it down into smaller chunks basically
- Adding edit buttons to some fields for RoS
- Confirmation modal before making the changes
- Added endpoints to handle ROS field updates


arranfw (commented):
What do you think about splitting up large pieces of work like this into multiple PRs? It's probably tough to do at this point but I'd be interested in chatting about how something like this could be split up in the future to make it a bit easier to review

server/services/return-of-service.js
```diff
@@ -99,8 +109,70 @@ const getReturnOfServiceStatuses = async ({ participantId }) =>
...
+};
+
+const updateReturnOfServiceStartDate = async ({ participantId, newStartDate }) => {
+  await validateRosParticipant(participantId);
+  if (!newStartDate) {
+    throw new Error(rosError.fieldNotFound);
+  }
+
+  // validate start date
+  const res = await dbClient.db[collections.ROS_STATUS].find({
```
arranfw, line 154: How is this validating the start date? It looks like we're validating a ROS status exists

server/routes/return-of-service.js
```diff
@@ -98,29 +118,88 @@ router.get(
...
+        default:
+          throw new Error(rosError.fieldNotFound);
+      }
+
+      logger.info({
+        action: actionName,
+        performed_by: user,
+        id: response.id,
+      });
+      return res.status(201).json(response);
```
arranfw, line 193: This is an edit rather than a creation, would 200 be more appropriate?

server/routes/return-of-service.js
```diff
@@ -98,29 +118,88 @@ router.get(
...
+            newDate: newValue,
+          });
+          break;
+
+        case rosFieldUpdate.startDate:
+          response = await updateReturnOfServiceStartDate({
+            participantId: participant.id,
+            newStartDate: newValue,
+          });
+          break;
```
arranfw, line 182: I'm not convinced we need three methods for these edits, could we combine them using a pattern like

```
validateSite(req.body.site);
validateStartDate(req.body.startDate);
const edited = {
  ...old,
  date: req.body.date || old.date,
  startDate: req.body.startDate || old.startDate,
  site: req.body.site || old.site,
}
db.save(edited);
```

server/routes/return-of-service.js
```diff
@@ -98,29 +118,88 @@ router.get(
...
+
+        default:
+          throw new Error(rosError.fieldNotFound);
```
arranfw, line 185: This could be validated by checking the incoming request body, if an invalid field is found we throw an error

server/routes/return-of-service.js
```diff
@@ -2,17 +2,50 @@ const express = require('express');
...
+    logger.error({
+      action: actionType,
+      message: `Participant ${participantId} not found`,
+    });
+    return null;
+  }
+  return participant;
+};
+
+const getRosErrorMessage = (message) => {
```
arranfw, line 31: If we're getting a message, should we pass in something like `type` rather than `message`?

client/src/pages/private/participant-details/constants.js
```diff
@@ -0,0 +1,105 @@
...
+    .then((res) => {
+      setAllSites(res?.data);
+    })
+    .catch((err) => {
+      setError(`${err}`);
+    });
+};
+
+// Get Cohort name
+export const getCohortName = (cohort) => {
```
arranfw, line 102:
```suggestion
export const getCohortPSIName = (cohort) => {
```
I wonder if we want to include PSI in the name to indicate we're combining the two

client/src/pages/private/participant-details/modals/AssignCohortDialog.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Box, Grid, Typography } from '@material-ui/core';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+
+import { getCohortName } from '../constants';
+import { Dialog, Button } from '../../../../components/generic';
+
+export const AssignCohortDialog = ({ isOpen, onClose, onSubmit, participant, selectedCohort }) => {
```
arranfw, line 9:
```suggestion
export const AssignCohortSiteDialog = ({ isOpen, onClose, onSubmit, participant, selectedCohort }) => {
```
Does this make more sense?

client/src/pages/private/participant-details/constants.js
```diff
@@ -0,0 +1,105 @@
...
+      setActualParticipant(resp);
+      if (
+        resp.interested?.toLowerCase() === 'withdrawn' ||
+        resp.interested?.toLowerCase() === 'no'
+      ) {
+        setDisableAssign(true);
+        return;
+      }
+
+      psi()
```
arranfw, line 80: I would love to rename this function, looks like a `get/fetchPSI` maybe?

client/src/pages/private/participant-details/constants.js
```diff
@@ -0,0 +1,105 @@
...
+export const fetchData = ({
+  setParticipant,
+  setActualParticipant,
+  setPSIList,
+  id,
+  setError,
+  setDisableAssign,
+  setAllSites,
+}) => {
+  fetchParticipant({ id })
```
arranfw, line 68: This might be more readable with async/await, was there a reason to use promise chaining?

client/src/pages/private/participant-details/constants.js
```diff
@@ -0,0 +1,105 @@
...
+  fullName: `${inputData.firstName} ${inputData.lastName}`,
+  interested: getInterestLabel(inputData.interested),
+  ros:
+    inputData.rosStatus && Object.keys(inputData.rosStatus).length
+      ? mapRosData(inputData.rosStatus)
+      : null,
+});
+
+// Helper
+export const fetchData = ({
```
arranfw, line 59: This looks like it would belong better in a services file rather than constants

client/src/pages/private/participant-details/ParticipantDetailsView.js
```diff
@@ -0,0 +1,385 @@
...
+  const classes = useStyles();
+  // Get param
+  const { id, page, pageId } = useParams();
+  // Breadcrumb name
+  const linkName = getLinkLabel(page);
+  // Edit Button flag
+  const enableEdit = roles.some((role) => ['ministry_of_health', 'superuser'].includes(role));
+  const isMoH = roles.includes('ministry_of_health');
+
+  const rosKeyMap = {
```
arranfw, line 78: What you've done with the edit form is pretty cool, but I wonder if it's a bit too much complexity in the name of reusing the form. From the perspective of someone wanting to modify the "Edit site name form" for example, they need to understand quite a bit of logic and risk affecting other forms with their changes.

Would defining each of these forms individually be better?

client/src/pages/private/participant-details/constants.js
alva-fresh: I'm not certain about the reasoning behind using promises in this case. It could be because it's called in `useEffect` which doesn't always tolerate async/await

client/src/pages/private/participant-details/constants.js
alva-fresh: Sounds like a good idea!

client/src/pages/private/participant-details/ParticipantDetailsView.js
alva-fresh: Makes sense 👍 I think it's best to have separate forms for each field

server/routes/return-of-service.js
alva-fresh: I'm using mostly to determine which function to call, rather than for validation purposes. I wasn't sure what a `default` clause could be in this case

server/routes/return-of-service.js
alva-fresh: I don't think we can validate all fields by default. Only 1 field is being passed, so the validation for others will fail

server/services/return-of-service.js
alva-fresh: Maybe the wording is confusing: I'm looking for a participant who has `assigned-new-site` entry here. Basically someone who changed a site. If the entry is not present, then this participant never changed sites hence _Start date at a new site_ cannot be updated.

client/src/components/modal-forms/EditRosStartDate.js
```diff
@@ -0,0 +1,75 @@
+import React, { useState } from 'react';
```
fw-noel, line 1: I feel like these three components **could** be cleaned to a common component, but may not be worth the effort

client/src/services/participant.js
```diff
@@ -1,8 +1,8 @@
 import store from 'store';
 import { API_URL, pageSize, postHireStatuses } from '../constants';
 
-const getCohortName = (cohort = {}) =>
-  cohort.cohort_name && cohort.psi?.institute_name
+export const getCohortPsiName = (cohort = {}) =>
+  cohort && cohort.cohort_name && cohort.psi?.institute_name
```
fw-noel, line 5: `cohort?.cohort_name &&`
?

fw-noel (commented):
Will finish going over the server later / tomorrow morn

server/routes/return-of-service.js
```diff
@@ -2,17 +2,50 @@ const express = require('express');
...
 const router = express.Router();
 
+const validateParticipant = async (participantId, actionType) => {
+  const [participant] = await getParticipantByID({ id: participantId });
+  if (!participant) {
+    logger.error({
+      action: actionType,
+      message: `Participant ${participantId} not found`,
+    });
+    return null;
```
fw-noel, line 26: 1. Any reason we're not throwing the error from here? I'm also not sure it's a 404, vs. a 422
2. Or I guess digging deeper, does getParticipantByID not throw an error when not found? That's straight from massive js right?

server/routes/return-of-service.js
```diff
@@ -98,29 +118,88 @@ router.get(
...
+    const { email, user_id: userId, sub: localUserId } = req.user;
+    const user = userId || localUserId;
+    if (!(email && user)) {
+      return res.status(401).send('Unauthorized user');
+    }
+    const participant = await validateParticipant(participantId, actionName);
+    if (!participant) {
+      return res.status(404).send('Participant not found');
+    }
+    const { newValue, fieldType } = req.body;
```
fw-noel, line 158: I don't think I'm a fan of this pattern of passing in `fieldType`. Generally, I think a patch would be optional properties of an entity, so in this case you would do something like:
```
const { siteId, date, startDate } = req.body;
if (siteId)...
else if (date)...
else if (startDate)...
```

That said, I do agree with Arran in the comment on the three functions

server/routes/return-of-service.js
fw-noel: If you mean the validation on the functions themselves, like
```
if (!newSiteId) { throw new Error; }
```

I do think the right way to do it is a single `updateRosStatus` change. I'm not sure if Massive works by saving entities, but my thoughts are something like:
```
updateReturnOfService = async ({ participantId, newValues }) => {
_const existingRos = find participant ROS by participant id_
_if no participant, throw error_
_if existingRos has no delta with newValues, throw errror_
_any other checks, such as checking that for `assigned-new-site` if startDate has changed_
_apply new values into existingRos_
_save ROS_
}
```

Open to discussion here, but by and large, I wouldn't prefer a function for each property change

client/src/components/modal-forms/EditRosDate.js
```diff
@@ -0,0 +1,70 @@
...
+              <FormButtons
+                onClose={onClose}
+                onSubmit={async () => {
+                  setSubmitting(false);
+                  const res = await validateForm();
+                  if (Object.entries(res)?.length === 0) {
+                    openConfirmationDialog(values);
+                    return;
+                  }
+                  await submitForm();
```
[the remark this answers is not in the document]
alva-fresh: It only gets submitted if there're no errors. Basically, if errors found it'll `return` before calling `submitForm()`

client/src/components/modal-forms/EditRosStartDate.js
alva-fresh: I'll need to look into it either way. Otherwise, `SonarCloud` checks aren't going to pass 😅

client/src/services/participant.js
alva-fresh: Yep, that's probably better ☝️
I'll update it

server/routes/return-of-service.js
alva-fresh: I just realized that yesterday after positing a comment. Moving away from having multiple functions is definitely a good idea. Let me try to update that now

server/routes/return-of-service.js
alva-fresh: `getParticipantByID` doesn't throw an error by default here

alva-fresh:
@arranfw @fw-noel thank you for reviewing this and all your comments!! 🙌 
I know it's quite big, I'll avoid making such huge requests in the future. At this point, I don't think it can be broken down into smaller pieces. If I missed any of your questions, please let me know. I'm not ignoring them, it's just easy to miss something when there's 30+ comments

alva-fresh:
There's still some work remaining and I'll make it a separate request for this. I need to create a table to keep track of the changes made. The other thing I was thinking about is that we currently don't have `updated_at` column to track changes to RoS. Do you think it's a good idea to add it? Open for discussion here

client/src/components/modal-forms/form-components/ConfirmationForm.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import { Box, Typography } from '@material-ui/core';
+
+import { Dialog } from '../../generic';
+import { FormButtons } from './';
+
+export const ConfirmationForm = ({ isOpen, onSubmit, onClose }) => {
```
arranfw, line 8: This component API looks like it would be a generic confirmation component, but there are a couple issues with it

- Name is `Form` but it returns a Dialog
- Hardcoded copy prevents reuse for other features

I'd suggest renaming this component and passing in the copy as a prop so it can be reused in the future

client/src/services/return-of-service.js
```diff
@@ -61,3 +61,31 @@ export const getAllSites = async () => {
     throw new Error('Failed to fetch all sites');
   }
 };
+
+export const onRosUpdate = async (participantId, newValues) => {
```
arranfw, line 65:
```suggestion
export const updateRosStatus = async (participantId, newValues) => {
```

This naming is too coupled to the calling component, it should describe what it's doing, not what's calling it

server/services/return-of-service.js
```diff
@@ -99,8 +99,45 @@ const getReturnOfServiceStatuses = async ({ participantId }) =>
...
+    throw new Error(rosError.noSiteAttached);
+  }
+  if (newDate && !rosParticipant.data.date) {
+    throw new Error(rosError.noDate);
+  }
+  if (newStartDate && !rosParticipant.data.startDate) {
+    throw new Error(rosError.noStartDate);
+  }
+
+  dbClient.db[collections.ROS_STATUS].update(
```
arranfw, line 125: Did we talk about editing a status vs adding a new one and setting the current to `current=false`? On 
_transaction_ table I'd be wary of editing records and I'd opt for adding a new one and making it the current

If we save the editing user's ID this might even prevent the need for an audit table because we're not removing any data

arranfw (changes requested):
I'd like to chat about editing vs adding a new status, the ConfirmationForm component, and the SonarCloud warnings

server/services/return-of-service.js
arranfw: It looks like this check is gone now, did it not end up being needed?

server/routes/return-of-service.js
```diff
@@ -8,11 +8,81 @@ const { getParticipantByID } = require('../services/participants');
 const {
   makeReturnOfServiceStatus,
   getReturnOfServiceStatuses,
+  updateReturnOfServiceStatus,
 } = require('../services/return-of-service');
 const { getSiteDetailsById, getDetailsBySiteId } = require('../services/employers');
 
 const router = express.Router();
 
+const validateParticipant = async (participantId, actionType) => {
```
arranfw, line 17: These look like they belong in a services/validation folder

server/routes/return-of-service.js
```diff
@@ -8,11 +8,81 @@ const { getParticipantByID } = require('../services/participants');
...
+    case rosError.noStartDate:
+      return { label: 'Start date at a new site is not recorded', statusCode: 400 };
+    default:
+      return {
+        text: 'Internal server error: unable to create return of service status',
+        statusCode: 500,
+      };
+  }
+};
+
```
arranfw, line 85: These look like they belong in a services/validation folder

arranfw:
https://github.com/bcgov/hcap/pull/782#discussion_r944677923

Getting close, this is the last major piece I want to chat about

fw-noel:
I converted this to a draft so we can differentiate between what can be merged.
But I'm realizing now we should maybe use tags

alva-fresh:
TODO: ask Tavy if changes made by MoH should be showing up on the RoS reports

server/routes/return-of-service.js
```diff
@@ -165,44 +107,4 @@ router.get(
   })
 );
 
-// Update return of service fields
-router.patch(
```
arranfw: What was the idea behind combining the endpoints? It's fairly typical to have separate endpoints for Create and Update. I think I prefer the duplication over the branching logic in the `POST` endpoint and it's not super correct to be modifying a record in a `POST` endpoint

alva-fresh: Yeah, I also prefer `PATCH`, but I wan't sure about it in this case. I thought we're creating a new entry and not editing it, isn't that right?
Do you think we can still keep `PATCH` endpoint in this case?

arranfw: I think we want the endpoints to match the conceptual action, the fact that we're using a transaction table shouldn't affect our endpoint structure

So data-wise we're adding a record but logically we're updating a user's return of service (this is my reasoning anyway, I do see the other side) @fw-noel What do you think?

fw-noel: I need to look at the code from a higher level, but from first impressions, I think a POST seems fine, and is consistent with how things like participant statuses work

client/src/services/return-of-service.js
```diff
@@ -61,3 +61,33 @@ export const getAllSites = async () => {
...
+    throw new Error('Unable  to update the field - no changes found');
+  }
+  if (!participantId) {
+    throw new Error('Unable  to update the field - invalid data');
+  }
+  const dateTimestamp = date ? dayjs(date, 'YYYY/MM/DD').toDate() : undefined;
+  const startDateTimestamp = startDate ? dayjs(startDate, 'YYYY/MM/DD').toDate() : undefined;
+
+  return fetch(url, {
+    method: 'PATCH',
```
fw-noel, line 79: Are we back to where we started here? In this PR, are we:
- Sending a PATCH request from the frontend to update ROS status
- In the PATCH endpoint, we are creating a new ROS status for this participant and invalidating any previous ones, so as to create an audit log

Really sorry, am I misunderstanding what's happening?

fw-noel (commented):
I think I need more understanding, sorry!
As another note, this may be a good candidate for one of those Screen captures for a ticket~

fw-noel (dismissed):

server/services/return-of-service.js
```diff
@@ -1,6 +1,111 @@
...
+  if (isMohUpdate || assignNewSite) {
+    rosParticipant = await getRosParticipant(participantId);
+  }
+
+  let site_id;
+  if (isMohUpdate && data.site) {
+    site_id = await getDbSiteId(data.site);
+  }
+  if (!isMohUpdate) {
+    site_id = siteId;
```
arranfw, line 78: Is this db ID vs site id? This could use a rename or a comment to better explain what's going on here

server/services/return-of-service.js
```diff
@@ -1,6 +1,111 @@
...
+  }
+  if (updatedDate && !rosParticipant.data.date) {
+    throw new Error(rosError.noDate);
+  }
+  if (updatedStartDate && !rosParticipant.data.startDate) {
+    throw new Error(rosError.noStartDate);
+  }
+};
+
+const formatRosData = async (
```
arranfw, line 57: There are still two smells I see here

1. Flag argument in `isMohUpdate` https://martinfowler.com/bliki/FlagArgument.html
2. By the time a service method is called, validation should already have been performed on user input.

In general this is how I would design these methods to reduce conditionals and improve separation of concerns. Don't hesitate to reach out if you want to chat about this

image

arranfw (commented):
I'd like to chat about a few of these comments

server/services/return-of-service.js
fw-noel: I do agree that the validation is best performed on the controller / route function.

I'm less against the flag being present in this scenario. Seems like the alternative is two different functions or passing in the user's role, and I don't think I like any of the options more than another. Unless you had a design suggestion for that @arranfw ?

arranfw: My suggestion was to take all of the edit specific functionality and put it in an `updateRosStatus` method

arranfw (approved):
Great work :)

taraepp (approved):

PR #783, opened 2022-08-08
HCAP-1060 | Participants per page dropdown

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1060
- used the MUI TablePagination in Table component (should be reusable for other tables) 
- moved the pageSize variable from constants into ParticipantsContext


arranfw (approved):
Awesome work!

PR #790, opened 2022-08-09
HCAP-1128 cache docker

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,10 @@ local-kc-run: local-kc-build
 	@echo "Starting test local app container"
 	@docker-compose -f docker-compose.test.yml up -d
 
+local-kc-run-only:
+	@echo "Starting test local app container"
+	@docker-compose -f docker-compose.test.yml up -d 
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -81,6 +81,10 @@ local-kc-run: local-kc-build
 	@echo "Starting test local app container"
 	@docker-compose -f docker-compose.test.yml up -d
 
+local-kc-run-only:
+	@echo "Starting test local app container"
+	@docker-compose -f docker-compose.test.yml up -d
```
taraepp, line 86: This looks to me like the same as local-kc-run? What is it for?

fw-noel: Without the build

fw-noel (approved):

Makefile
arranfw: hated this name but I didn't want to change existing commands I might have preferred

build
run
build-and-run

PR #792, opened 2022-08-11
fix: Allow employers to track grad in participant view

fw-noel:
Employers should be table to track graduation in the participant details page, but not see the assign cohort tab
Add a warning if it's a user who needs to notify a health authority to assign the participant to another cohort
[Related Ticket](https://freshworks.atlassian.net/browse/HCAP-1188)


client/src/components/modal-forms/ManageGraduationForm.js
```diff
@@ -103,6 +107,17 @@ export const ManageGraduationForm = ({ initialValues, onClose, onSubmit, cohortE
                       </MuiAlert>
                     </Box>
                   )}
+                {values?.continue === 'continue_yes' &&
+                  values.status === postHireStatuses.cohortUnsuccessful &&
+                  !roles.includes('health_authority') && (
```
fw-noel, line 112: I'm not gonna lie, I kind of hate that I'm using AuthContext in this component, but I thought it better than prop drilling a flag

client/src/components/participant-details/psi-cohort.js
```diff
@@ -97,39 +93,43 @@ const PSIRouteTabs = ({
...
-            tab={tab}
-            setTab={setTab}
-            fetchData={fetchData}
-            assignAction={assignAction}
-            disabled={disabled}
-            psiList={psiList}
-          />
-        </Route>
-        <Redirect to='/' />
+        {tabDetails.trackGraduation &&
```
fw-noel, line 109: I don't love this either, I would've rather included the component in `tabDetails` or something, but I'm open to suggestions

fw-noel (commented):
A little strange to have the CustomTabs for a single tab in this case, but I thought it best to reuse so that it can look consistent.

client/src/components/participant-details/psi-cohort.js
```diff
@@ -138,11 +139,29 @@ const PSIRouteTabs = ({
...
+  const { auth } = AuthContext.useAuth();
+  const roles = useMemo(() => auth.user?.roles || [], [auth.user?.roles]);
   const baseUrl = match.url.split(tab)[0];
   const sortedList = sortPSI({ psiList, cohort: participant ? participant.cohort : {} });
+
+  // Determine the tags depending on user role
+  const assignCohort = {
+    label: 'Assign Cohort',
+    path: '/assign-cohort',
+    default: true,
```
arranfw, line 151: I'm not a huge fan of `default` here, reading this I would think `default` controls the initially selected tab, but shouldn't the first tab usually be the default?
Probably more a smell of the design of this component, if it used a composition API we could conditionally render the tabs rather then having to fiddle with the data like this

arranfw (dismissed):
Looks good, I think we could improve the API for `PSIRouteTabs` a bit to save on logic but that's my only major comment

client/src/components/participant-details/psi-cohort.js
fw-noel: Yeah I'm okay with using first tab.

arranfw (dismissed):

alva-fresh (approved):
Looks good!
Although, I think combining imports `import React, { useMemo } from 'react';` in `ManageGraduationForm` could be a very small improvement. But overall looks good 👍

arranfw (approved):

arranfw (approved):

alva-fresh (approved):

PR #793, opened 2022-08-11
HCAP-1142: MoH Edit Participant Return of Service Info (Part 2 - migration)

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1142

❓**What's included here:** 🎨 🖌️
- small bug fix with retrieving `site_id` for RoS
- creating migration to add a new table
- tracking changes made by MoH in a new table


server/services/return-of-service.js
```diff
@@ -2,6 +2,16 @@
 const { dbClient, collections } = require('../db');
 const { rosError } = require('../constants');
 
+const getRosSiteId = async (siteId) => {
```
arranfw, line 5: This naming is a bit confusing because it's get: `siteId` and we pass in `siteId`, would `getRosDbId` make more sense?

server/services/return-of-service.js
```diff
@@ -112,8 +128,12 @@ const updateReturnOfServiceStatus = async ({ participantId, newSite, newDate, ne
...
   }
 
-  if (newSite && !rosParticipant.site_id) {
-    throw new Error(rosError.noSiteAttached);
+  let newSiteId;
+  if (newSite) {
+    if (!rosParticipant.site_id) {
+      throw new Error(rosError.noSiteAttached);
+    }
+    newSiteId = await getRosSiteId(newSite);
```
arranfw, line 136: What's the need for the database ID here? It seems like an extra database call just to switch from one ID to the other, is the id in `newSite` not appropriate for logging?

server/services/return-of-service.js
```diff
@@ -122,17 +142,26 @@ const updateReturnOfServiceStatus = async ({ participantId, newSite, newDate, ne
...
-        startDate: newStartDate || rosParticipant.data.startDate,
-      },
-    }
+    updatedData
   );
+  dbClient.db[collections.MOH_EDIT_AUDIT].insert({
+    user: user || 'system',
+    data: {
+      participantId,
+      oldData: { ...rosParticipant },
```
arranfw, line 161: In other parts of the codebase where we track changes we use `from` and `to`, would it be more consistent to apply that naming here?

https://github.com/bcgov/hcap/blob/17fd7f9dcd9085c2d28c375bae1ca78600de55ab/server/services/participants.js#L136

server/services/return-of-service.js
alva-fresh: Good point!

server/services/return-of-service.js
alva-fresh: So, the `employer_sites` table has 2 ids: `id` and `siteId`; I need to link site to the `ros` table, so I need to know which id to use
`newSite` = `siteId` (I cannot attach it as FK to ros)
`newSiteId` = `id`, which I'm using as an actual siteId in this case to create relation

ScreenShot

server/db/schema.js
```diff
@@ -20,6 +20,7 @@ const collections = {
   ROS_STATUS: 'return_of_service_status',
   ADMIN_OPS_AUDIT: 'admin_operation_audit',
   SITE_PARTICIPANTS_STATUS: 'site_participants_status',
+  MOH_EDIT_AUDIT: 'moh_edit_audit',
```
arranfw, line 23: Where do we document the purpose of tables? Is it clear enough that this table is only used to track MOH edits on participant applications? If I looked at this table and it's data I feel like I could assume this might track other things MoH might be able to edit (site data for example)

Last comment I think! Sorry for the back and forth

alva-fresh:
Closing as per discussion - https://github.com/bcgov/hcap/pull/782#discussion_r944677923
creating new audit table isn't necessary

PR #795, opened 2022-08-12
HCAP-1271: 'Move to archive' bug fix

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1271

HA are now able to move withdrawn participants in `hired` state to archive

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -233,7 +233,7 @@ const ParticipantTable = () => {
     setActiveModalForm(participantEngageStatus.MULTI_SELECT_SITE);
   };
 
-  const handleEngage = async (participantId, status, additional = {}, participantInfo = null) => {
+  const handleEngage = async (participantId, status, additional = {}, _participantInfo = null) => {
     let closeModal = true;
     try {
       if (isLoadingData) {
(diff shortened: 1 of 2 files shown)


client/src/pages/private/ParticipantTable.js
```diff
@@ -233,7 +233,7 @@ const ParticipantTable = () => {
     setActiveModalForm(participantEngageStatus.MULTI_SELECT_SITE);
   };
 
-  const handleEngage = async (participantId, status, additional = {}, participantInfo = null) => {
+  const handleEngage = async (participantId, status, additional = {}, _participantInfo = null) => {
```
arranfw, line 236: Can we just remove this parameter?

arranfw (commented):
I think this makes sense, I'm mostly just not sure about which status' are set on which actions

What status does MoH archival put the user into? Based on this change I would think `hired` but that doesn't make sense

What are the statuses when the HA is archiving the user? If they're archived by MoH why are they being archived again?

client/src/pages/private/ParticipantTable.js
alva-fresh: I wonder the same, not sure at what stage it became optional. But it's being passed on for every `handleEngage`

fw-noel:
We need Tavy to confirm what behavior should look like here

PR #796, opened 2022-08-12
HCAP-1302: page width bug fix

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1302

Adding `minWidth` value to prevent pages from shrinking based on content

### DIFF
--- a/client/src/pages/private/ParticipantDetailsView.js
+++ b/client/src/pages/private/ParticipantDetailsView.js
@@ -78,6 +78,9 @@ const customStyle = makeStyles({
   rootContainer: {
     flexGrow: 1,
   },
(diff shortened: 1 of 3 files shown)


client/src/pages/private/ParticipantDetailsView.js
```diff
@@ -78,6 +78,9 @@ const customStyle = makeStyles({
   rootContainer: {
     flexGrow: 1,
   },
+  cardRoot: {
+    minWidth: '1020px',
```
arranfw, line 82: Shouldn't these be the same value?

arranfw (approved):

PR #799, opened 2022-08-17
fix: ignore ROS if participant is not in my sites

fw-noel:
https://freshworks.atlassian.net/browse/HCAP-1304

Basically, my candidates was showing a user who has already gotten ROS for a site I don't have access to and making them actionable. In talking with Tavy, we've determined that if a user gets hired for a site I don't have access to, there can be no further action on that participant.

In this PR, we're doing an inner join which means the participant no longer gets returned if they were ROS'd for a site I don't have access to. This should be okay for now. (Ideally, the participant will be returned with an empty `rosStatuses` so that they show up and still need to be acknowledged, but I was having some trouble).

### DIFF
--- a/server/services/participants-helper.js
+++ b/server/services/participants-helper.js
@@ -41,7 +41,7 @@ const flattenParticipants = (participants) => {
   return flattenedParticipantList;
 };
 
-const scrubParticipantData = (raw, joinNames) =>
+const scrubParticipantData = (raw, joinNames, sites) =>
   raw.map((participant) => {
     const statusInfos = [];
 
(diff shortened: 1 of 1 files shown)


server/services/participants-helper.js
```diff
@@ -73,13 +73,15 @@ const scrubParticipantData = (raw, joinNames) =>
       });
     }
 
+    let rosStatuses = participant.rosStatuses
+      ? participant.rosStatuses.sort((ros1, ros2) => ros2.id - ros1.id)
+      : [];
+    rosStatuses = rosStatuses.filter(rosStatus => sites.includes(rosStatus.rosSite.body.siteId));
```
arranfw, line 79: Nice, this won't affect pagination right? We're not affecting the number of records returned, just preventing ROS statuses from sites the requesting user doesn't have access to?

arranfw (approved):

server/services/participants-helper.js
fw-noel: Correct!

arranfw (approved):

PR #803, opened 2022-08-19
fix: infinite loop when database has no sites

arranfw:
This causes an infinite loop if you run the app without sites in the database


taraepp (approved):
Looks good to me!

PR #804, opened 2022-08-19
HCAP-1234 | RoS Start Date Column For MoH (+SU)

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1234

- server/services/participants-helper.js
   - rosStatuses didn't return _at all_ for MoH/SU, necessitating some logic changes
- server/migrations/1660318202044_update_participant_status_infos.js
   - updated the view used in participants-helper used for MoH/SU. This is the only usage of this view, so I changed the view to return what was missing (rosStatuses namely)
- client/src/constants/participantTableConstants.js
  - added rosStartDate to list of columns. I'm actually not 100% sure on the tabs/tabStatuses though
- client/src/pages/private/ParticipantTable.js
  - got some undefined errors after bringing in rosStatuses for MoH here, basically added null checks in a few places.

image

### DIFF
--- a/server/migrations/1660318202044_update_participant_status_infos.js
+++ b/server/migrations/1660318202044_update_participant_status_infos.js
@@ -1,0 +1,20 @@
+/* eslint-disable camelcase */
+const { dbClient, views } = require('../db');
+
+exports.shorthands = 'update-participant-status-infos';
+
+exports.up = async () => {
+  await dbClient.db.query(`CREATE OR REPLACE VIEW ${views.PARTICIPANTS_STATUS_INFOS} AS
+    SELECT p.id,
+     p.body,
+     p.search,
+     p.created_at,
+     p.updated_at,
+     jsonb_agg(ps.*) FILTER (WHERE ps.id IS NOT NULL) AS status_infos,
+     jsonb_agg(DISTINCT jsonb_build_object('id', rosstatus.id, 'data', rosstatus.data, 'status', rosstatus.status, 'site_id', rosstatus.site_id, 'rosSite', ros_site.*)) FILTER (WHERE rosstatus.id IS NOT NULL) AS ros_infos
+    FROM participants p
+      LEFT JOIN return_of_service_status rosstatus ON p.id = rosstatus.participant_id
+      LEFT JOIN participants_status ps ON p.id = ps.participant_id AND ps.current IS TRUE
+      LEFT JOIN employer_sites ros_site ON rosstatus.site_id = ros_site.id
+   GROUP BY p.id, p.body, p.search, p.created_at, p.updated_at`);
+};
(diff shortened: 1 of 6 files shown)


server/migrations/1660318202044_update_participant_status_infos.js
```diff
@@ -0,0 +1,20 @@
...
+  await dbClient.db.query(`CREATE OR REPLACE VIEW ${views.PARTICIPANTS_STATUS_INFOS} AS
+    SELECT p.id,
+     p.body,
+     p.search,
+     p.created_at,
+     p.updated_at,
+     jsonb_agg(ps.*) FILTER (WHERE ps.id IS NOT NULL) AS status_infos,
+     jsonb_agg(DISTINCT jsonb_build_object('id', rosstatus.id, 'data', rosstatus.data, 'status', rosstatus.status, 'site_id', rosstatus.site_id, 'rosSite', ros_site.*)) FILTER (WHERE rosstatus.id IS NOT NULL) AS ros_infos
+    FROM participants p
+      LEFT JOIN return_of_service_status rosstatus ON p.id = rosstatus.participant_id
```
arranfw, line 16: Should this have `AND rosstatus.current IS TRUE`

server/migrations/1660318202044_update_participant_status_infos.js
```diff
@@ -0,0 +1,20 @@
...
+
+exports.up = async () => {
+  await dbClient.db.query(`CREATE OR REPLACE VIEW ${views.PARTICIPANTS_STATUS_INFOS} AS
+    SELECT p.id,
+     p.body,
+     p.search,
+     p.created_at,
+     p.updated_at,
+     jsonb_agg(ps.*) FILTER (WHERE ps.id IS NOT NULL) AS status_infos,
+     jsonb_agg(DISTINCT jsonb_build_object('id', rosstatus.id, 'data', rosstatus.data, 'status', rosstatus.status, 'site_id', rosstatus.site_id, 'rosSite', ros_site.*)) FILTER (WHERE rosstatus.id IS NOT NULL) AS ros_infos
```
arranfw, line 14: This is intense, I'd love an explanation

fw-noel: Agreed

taraepp: This was tricky to build, but yes!
- this is the existing view `participants_status_infos` but with `ros_infos` added as a column
- I added the 2 joins to `ros_status` and `employer_sites` tables
- naturally I tried to add it like the `status_infos` column at first (doing `jsonb_agg` on the `ros_status` table, filtering out nulls)
- but there was a problem! Because it's an aggregate query, if I had 3 results from `participants_status` table and 1 from `ros_status` for a participant, I'd get 3 identical ros results
- each ros status was getting duplicated in the query for each participants status
- so I needed to do a `DISTINCT` on it to get rid of duplicates, but turns out that it doesn't know how to do a DISTINCT on json results
- that's what jsonb_build_object is for- turns it into something it can compare
- so I'm selecting distinct results from a json object that I've built

As for current...
I'm not 100% sure if we need a current filter. By the time the data makes its way to the FE, we're only looking at the ros status with the highest id. I don't know if there's any cases where the most recent ros status isn't valid but the prior status is.

fw-noel: I understand, thanks!

I think I'd still urge for the current filter just for safeguarding. You never know in the future, or if someone makes manual database changes. We may as well take advantage of `is_current` where we have it!

arranfw: ^ Agreed, for better or for worse it's a feature of these tables and if we have it we should be consistent about using it, leaving it out could cause another [HCAP-1265](https://freshworks.atlassian.net/browse/HCAP-1265)

arranfw (approved):

PR #805, opened 2022-08-20
fix: use hiring site to determine permissions for ros status

fw-noel:
http://hcapemployers.local.freshworks.club:4000/

We learned that ROS site is not what determines if an HA can see a participant or not, but more the hiring site.

This will conflict with Tara's PR


arranfw (dismissed):
Ahh this makes sense

taraepp:
After resolve merge conflict, check:
- sign in with MoH
- make sure that RoS start date column in participants table has date properly (need at least 1 ros participant)

arranfw (approved):

taraepp (approved):
Looks good to me!

arranfw (approved):

PR #807, opened 2022-08-22
fix: use employer or same site when doing employer actions

fw-noel:
This is a fix for https://freshworks.atlassian.net/browse/HCAP-1308 but serves in two ways:

When posting an employer action, there's a touch less reliance on Employer Id of a status. 1308 was caused when the `employer_id` was different from the user trying to archive. In `participant-status.js`, we see the cause:
```
    const existingCurrentStatus = currentStatusId
      ? await tx[collections.PARTICIPANTS_STATUS].findOne({ id: currentStatusId })
      : await tx[collections.PARTICIPANTS_STATUS].findOne({
          participant_id: participantId,
          current: true,
          employer_id: employerId,
        });
```

~~That is when no currentStatusId is passed in. By getting the status id first, we not pass it in here properly, but we also check that the accessing user has the right permissions for the participant.~~

We're keeping it at no status passed in, but there's still less reliance on the hiring user being the same person.

### DIFF
--- a/client/src/utils/prettifyStatus.js
+++ b/client/src/utils/prettifyStatus.js
@@ -25,7 +25,7 @@ export const getParticipantStatusData = (status, tabValue, isMoH, participantInf
   const isPendingAcknowledgement = status.includes('pending_acknowledgement');
   const isWithdrawn = status.includes('withdrawn');
   const isArchived = status.includes('archived');
-  const rosStartDate = isROS ? participantInfo.rosStatuses[0].data.date : null;
+  const rosStartDate = isROS ? participantInfo?.rosStatuses[0]?.data?.date : null;
   const finalStatus = statusInfo.data?.final_status ?? null;
 
   const toolTipText = getToolTipText({
--- a/server/routes/participant.js
+++ b/server/routes/participant.js
@@ -430,6 +430,12 @@ employerActionsRouter.post(
     if (!participant) {
       return res.status(400).json({ message: 'Could not find participant' });
     }
+    // Check site access
+    if (site && !user.sites.includes(site)) {
+      return res.status(400).json({
+        message: 'User does not have access to this site',
+      });
+    }
     const result = await setParticipantStatus(
       user.id,
       participantId,
--- a/server/routes/participant.js
+++ b/server/routes/participant.js
@@ -436,7 +436,7 @@ employerActionsRouter.post(
       status,
       { site, ...data },
       user,
-      currentStatusId || currentStatus.id
+      currentStatusId
     );
     logger.info({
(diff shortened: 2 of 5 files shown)


server/routes/participant.js
```diff
@@ -430,8 +431,10 @@ employerActionsRouter.post(
     if (!participant) {
       return res.status(400).json({ message: 'Could not find participant' });
     }
+    const currentStatus = await getParticipantCurrentStatus(participantId);
+    const currentSite = site || currentStatus?.data?.site;
```
arranfw, line 435: Should this be `currentSiteId`?

server/services/participants.js
```diff
@@ -127,6 +127,14 @@ const getParticipantByID = async (participantInfo) => {
   return participant;
 };
 
+const getParticipantCurrentStatus = async (participantId) => {
+  const participant = await dbClient.db[collections.PARTICIPANTS_STATUS].findOne({
```
arranfw, line 131: Does this take into account that participants can have more than one current status as a time? Does it matter?

arranfw (dismissed):

server/tests/participant-status.test.js
```diff
@@ -168,23 +168,23 @@ describe('Test Participant status data model and service', () => {
       city: 'Test City 1030',
     });
 
-    const emp1 = v4();
-    const emp2 = v4();
-    const emp3 = v4();
+    const emp1 = { id: v4(), sites: [site1.siteId, site2.siteId] };
```
fw-noel, line 171: I think these tests make more sense. In the original, emp1 was prospecting the same participant for two sites, which I don't believe is a scenario.

client/src/utils/prettifyStatus.js
```diff
@@ -25,7 +25,7 @@ export const getParticipantStatusData = (status, tabValue, isMoH, participantInf
   const isPendingAcknowledgement = status.includes('pending_acknowledgement');
   const isWithdrawn = status.includes('withdrawn');
   const isArchived = status.includes('archived');
-  const rosStartDate = isROS ? participantInfo.rosStatuses[0].data.date : null;
+  const rosStartDate = isROS ? participantInfo?.rosStatuses[0]?.data?.date : null;
```
fw-noel, line 28: This is unrelated to HCAP-1308, I just thought it was a minor change that I could piggyback here. Just some null checks. I couldn't reproduce Akhil's crashing issue, but this seems like a good idea regardless.

arranfw (approved):

alva-fresh (approved):
Looks good!

taraepp (approved):

PR #808, opened 2022-08-22
HCAP-1309 | Graduation date error message

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1309

- After changing the purpose of the date field, I made it so that it was "untouched" to not trigger the error message. 
- Bonus: now when going Unsuccessful-> attempt submit-> Graduated-> Unsuccessful, the error message from the first unsuccessful submit does not appear again when status changed to unsuccessful again

### DIFF
--- a/client/src/components/modal-forms/ManageGraduationForm.js
+++ b/client/src/components/modal-forms/ManageGraduationForm.js
@@ -21,12 +21,13 @@ export const ManageGraduationForm = ({ initialValues, onClose, onSubmit, cohortE
         onSubmit={onSubmit}
         validationSchema={ParticipantPostHireStatusSchema}
       >
-        {({ submitForm, values, setFieldValue }) => {
+        {({ submitForm, values, setFieldValue, setFieldTouched }) => {
           const handleStatusChange = ({ target }) => {
             const { value } = target;
             setFieldValue('status', value);
             if (value === postHireStatuses.postSecondaryEducationCompleted) {
               setFieldValue('data.date', cohortEndDate);
+              setFieldTouched('data.date', false);
               setFieldValue('continue', 'continue_yes');
             } else {
               setFieldValue('data.date', '');


fw-noel (approved):

client/src/components/modal-forms/ManageGraduationForm.js
```diff
@@ -21,12 +21,13 @@ export const ManageGraduationForm = ({ initialValues, onClose, onSubmit, cohortE
...
         validationSchema={ParticipantPostHireStatusSchema}
       >
-        {({ submitForm, values, setFieldValue }) => {
+        {({ submitForm, values, setFieldValue, setFieldTouched }) => {
           const handleStatusChange = ({ target }) => {
             const { value } = target;
             setFieldValue('status', value);
             if (value === postHireStatuses.postSecondaryEducationCompleted) {
               setFieldValue('data.date', cohortEndDate);
+              setFieldTouched('data.date', false);
```
arranfw, line 30: Wouldn't the issue be that the field is failing validation rather than if it's touched or not?

alva-fresh (approved):

client/src/components/modal-forms/ManageGraduationForm.js
arranfw: I wonder if using the shouldRevalidate option would solve this
```
              setFieldValue('data.date', cohortEndDate, true);
```
https://formik.org/docs/api/formik#setfieldvalue-field-string-value-any-shouldvalidate-boolean--void

arranfw: Do we know where that delay in the video comes from? It looks like the warning message only lasts for a second or so

arranfw (approved):

client/src/components/modal-forms/ManageGraduationForm.js
taraepp: When I was interacting with it, the error stayed up until I clicked the form. That's why I went with the "touched" route.

In my head it makes more sense that the touched value should be TRUE, not FALSE, which is the opposite of what I did. I'm thinking that due to the input's schema changing depending on the value of another input, these events fire in the following order and that's the root cause:
- change the status from Unsuccessful to Graduated
- handleStatusChange updates the graduated date
- the date field registers as touched
- the error message is displayed
- the schema changes to the graduated schema

PR #810, opened 2022-08-23
HCAP-1277 | MoH can create new phase

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1277

Included:
- migration to create global_phase table (attributes: id, name, start date, end date)
- front end modal to insert into global phase table
   - validation: all fields are required, date must be in reasonable range (1900-2099)
      - we reuse start/end date and validateDateIsReasonable error messages a lot so on FE validation these are now defined as variables (but not replaced in other places where they are used)
   - did not test that end date actually comes after start date
- backend endpoint to create the phase in db
- cypress tests (success scenario, validation failures)


client/src/services/site.js
```diff
@@ -0,0 +1,92 @@
...
+        id: row.id,
+      };
+    });
+    return rowsData;
+  } else {
+    return [];
+  }
+};
+
+export const handlePhaseCreate = async (phase) => {
```
arranfw, line 32: I think as a service method just `createPhase` is better because it's not strictly a handler in this scenario

client/src/pages/private/SiteTable.js
```diff
@@ -354,16 +192,47 @@ export default ({ sites, viewOnly }) => {
...
   };
 
   return (
     <>
-      <SiteFormsDialog
-        activeForm={activeModalForm}
-        onDialogSubmit={fetchSites}
-        onDialogClose={closeDialog}
-      />
+      {activeModalForm === 'new-site' && (
```
arranfw, line 211: I agree this still looks a bit chunky, I don't think the conditionals are necessary, and what do you think of moving the dialogs into the same component as the form `NewSiteDialog`?

Will these forms ever be rendered outside of a dialog?

client/src/pages/private/SiteTable.js
```diff
@@ -354,16 +192,47 @@ export default ({ sites, viewOnly }) => {
...
+        <Dialog title={'Create Site'} open={activeModalForm != null} onClose={closeDialog}>
+          <NewSiteForm
+            onSubmit={(values) => {
+              handleFormSubmit(handleSiteCreate, {
+                ...values,
+                siteId: parseInt(values.siteId),
```
arranfw, line 217: In the pursuit of cleaning up the markup, can the service method handle parseInt?

then we get a nice `onSubmit={(values) => handleFormSubmit(createSite(values))}`

If we curry `handleFormSubmit` we can simplify this down to

`onSubmit={handleFormSubmit(createSite)}`

client/src/pages/private/SiteTable.js
taraepp: The one thing that I don't like about it is that it is inconsistent with the rest of the codebase- we have a whole directory full of Form components that probably all go into a Dialog in every use case.

server/db/schema.js
```diff
@@ -110,6 +112,34 @@ const schema = {
...
   ],
+  phaseTables: [
+    {
+      definition: `CREATE TABLE IF NOT EXISTS ${collections.GLOBAL_PHASE} (
+        id serial primary key,
+        name varchar(255) not null,
+        start_date date not null,
+        end_date date not null,
+        created_by varchar(255) not null,
+        updated_by varchar(255) not null,
```
arranfw, line 123: updated fields should be nullable

taraepp: I can do that, but FYI when they create the phase, it uses the create user/time for the updated fields, so it never is null. I figured that creator makes sense as updater, because if I wanted to query by most recently updated that I would want to use the created date if it hadn't been edited yet.

client/src/components/modal-forms/NewPhaseDialog.js
```diff
@@ -0,0 +1,97 @@
...
+    phaseName: '',
+    startDate: '',
+    endDate: '',
+  };
+
+  const handleCreatePhase = async (phase) => {
+    const phaseJson = {
+      name: phase.phaseName,
+      start_date: phase.startDate,
+      end_date: phase.endDate,
```
arranfw, line 33: I would let the service method handle the this transformation

client/src/pages/private/SiteTable.js
```diff
@@ -284,15 +192,26 @@ export default ({ sites, viewOnly }) => {
...
+        open={activeModalForm === 'new-site'}
+        onSubmit={handleFormSubmit}
+        onClose={closeDialog}
+      />
+
+      <NewPhaseDialog
+        open={activeModalForm === 'new-phase'}
+        onSubmit={handleFormSubmit}
+        onClose={closeDialog}
       />
```
arranfw, line 215: 🤩

server/db/schema.js
```diff
@@ -110,6 +112,34 @@ const schema = {
...
+        id serial primary key,
+        phase_id integer references ${collections.GLOBAL_PHASE},
+        site_id integer references ${collections.EMPLOYER_SITES},
+        allocation integer not null,
+        start_date date,
+        end_date date,
+        created_by varchar(255) not null,
+        updated_by varchar(255) not null,
+        created_at timestamp with time zone DEFAULT now(),
+        updated_at timestamp with time zone DEFAULT now()
```
arranfw, line 139: Do we want to make these not null? (I see how it matters less with `DEFAULT` but would prevent accidental data deletion

server/services/phase.js
```diff
@@ -0,0 +1,9 @@
+const { dbClient, collections } = require('../db');
+
+const createGlobalPhase = async (phase, user) => {
+  const phaseJson = { ...phase, created_by: user.id, updated_by: user.id };
```
arranfw, line 4: Slack discussion on `updated_by` https://freshworks.slack.com/archives/C011G18PW5S/p1661456234658859

arranfw (approved):
Awesome :)

client/src/components/modal-forms/NewPhaseDialog.js
taraepp: I was on the fence with this one! The reason that I didn't was that the service is more reusable if it expects the correct attribute names (ie: if another action also creates the phase, then I don't have to use wrong attribute names to call the function). The mapping I felt belonged to the form itself.

server/db/schema.js
taraepp: definitely.

client/src/components/modal-forms/NewPhaseDialog.js
taraepp: But then I guess I put the logic in for fetchSites, but at least that one depends on the columns parameter that is also passed in to tell it how to format the data.

arranfw (dismissed):

server/routes/phase-allocation.js
```diff
@@ -0,0 +1,40 @@
...
+const logger = require('../logger.js');
+const { asyncMiddleware } = require('../error-handler.js');
+const { CreatePhaseSchema } = require('../validation');
+const { expressRequestBodyValidator } = require('../middleware');
+const { createGlobalPhase } = require('../services/phase');
+
+const router = express.Router();
+
+// Create Global Phase: POST
+router.post(
```
arranfw, line 12: I think we can merge this if we feature flag this endpoint

server/routes/phase-allocation.js
```diff
@@ -0,0 +1,40 @@
...
+      const response = await createGlobalPhase(body, user);
+      logger.info({
+        action: 'phase-allocation_post',
+        performed_by: {
+          username: user.username,
+          id: user.id,
+        },
+        globalPhaseId: response.id,
+      });
+      logger.info(response);
```
arranfw, line 31: Is it necessary to log the response here? We log `response.id` in the previous log

fw-noel (approved):
I believe nothing has changed but for the merging of the other two PRs, so I'm approving again based off that!

server/routes/phase-allocation.js
```diff
@@ -47,6 +48,10 @@ router.post(
     expressRequestBodyValidator(CreatePhaseSchema),
   ],
   asyncMiddleware(async (req, resp) => {
+    console.log('FEATURE_PHASE_ALLOCATION', FEATURE_PHASE_ALLOCATION);
```
arranfw, line 51: This is probably not meant to be here?

arranfw (approved):

PR #816, opened 2022-08-24
HCAP-1323 | RoS Start Date Column Sorting

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1323

Source of bug was that it was trying to sort on the column rosStatuses, but for Moh/SU, the query is different, it uses a view that had to have a different column name because it can only be lowercase name. First I changed it to just look at the right column in the code block that deals specifically with the sorting, but then I decided to move it up into the context/variable definition as that should also avoid future bugs and be less complex.

### DIFF
--- a/server/services/participants-helper.js
+++ b/server/services/participants-helper.js
@@ -574,7 +574,7 @@ class ParticipantsFinder {
     this.hiredGlobalJoin = 'hiredGlobalJoin';
     this.siteJoin = 'siteJoin';
     this.siteDistanceJoin = 'siteDistanceJoin';
-    this.rosStatuses = 'rosStatuses';
+    this.rosStatuses = user.isHA || user.isEmployer ? 'rosStatuses' : 'ros_infos';
   }
 
   filterRegion(regionFilter) {
--- a/server/services/participants-helper.js
+++ b/server/services/participants-helper.js
@@ -574,6 +574,7 @@ class ParticipantsFinder {
     this.hiredGlobalJoin = 'hiredGlobalJoin';
     this.siteJoin = 'siteJoin';
     this.siteDistanceJoin = 'siteDistanceJoin';
+    // MoH/SU users query a view with ros_infos column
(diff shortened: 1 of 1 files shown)


server/services/participants-helper.js
```diff
@@ -574,7 +574,7 @@ class ParticipantsFinder {
     this.hiredGlobalJoin = 'hiredGlobalJoin';
     this.siteJoin = 'siteJoin';
     this.siteDistanceJoin = 'siteDistanceJoin';
-    this.rosStatuses = 'rosStatuses';
+    this.rosStatuses = user.isHA || user.isEmployer ? 'rosStatuses' : 'ros_infos';
```
arranfw, line 577: Can we rename this field on the back end prior to sending it to the front end? Getting this inconsistency fixed as close to the source as possible would be best I think

taraepp: We can rename rosStatuses to ros_infos but we cannot rename ros_infos to rosStatuses.

arranfw: The service method that queries the view returns `ros_infos` right? Can we get it to modify that key to `rosStatuses` before sending it to the FE?

taraepp: It does change it to rosStatuses before it sends it to the FE (happens at line 75)- the sorting comes before that.

arranfw: Oh jeez sorry I thought this change was on the front end 🥴 

A comment might be all we can do, but is there anything we can do to make it more clear that we're calling it `ros_infos` because of the constraint on the view rather than because it's MoH

taraepp: I think a comment is best. Technically there's also the table variable in the constructor for PartticipantsFinder (which gets changed to the view for other user) that's really the source of the difference, but that one gets modified in a few places (when attaching a join for non-MoH) so I'm less inclined to change it (rosStatuses stays constant).

arranfw (approved):

PR #817, opened 2022-08-25
HCAP-1321: updates to EditRosSite form for MoH

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1321

Creating a shared component to edit RoS site as all fields that need to be updated are the same for HA and MoH


client/src/components/modal-forms/form-components/EditRosTemplate.js
```diff
@@ -35,30 +35,25 @@ export const EditRosTemplate = ({
...
-                  await submitForm();
-                }}
-              />
-            </Box>
-          </FormikForm>
-        )}
-      </Formik>
+        <FormButtons
+          onClose={onClose}
+          onSubmit={async () => {
```
arranfw, line 44: Just looking at this again, if the form button was `type='submit'` I imagine you'd be able to skip a lot of this extra submission/validation logic

alva-fresh: If I make button `type='submit'`, it'll trigger the submission and won't show the confirmation modal even if I set `setSubmitting(false);`

arranfw: You would have to modify the `onSubmit` logic to trigger the confirmation modal and then let the confirmation modal handle the network request

arranfw (commented):
I think a comment from earlier went missing, but I'm wondering why such  widespread changes were required? My initial interpretation of this ticket was that it was fairly simple but I could be misunderstanding what's going on

fw-noel (approved):
Looks okay

arranfw (approved):

PR #820, opened 2022-08-26
fix: using current hired status when archiving

fw-noel:
https://freshworks.atlassian.net/browse/HCAP-1308

Some consistency for Archiving.
In short, archiving from the table view in Hired Candidates passes in `site` and `currentStatusId`, both of which aren't passed in from the Participant Details view archiving. Akhil confirmed to me that for non-MoH users, Participant Details screen is only available for hired participants, so we can assume that they were hired here.

As such, we make the backend participant details endpoint return all current statuses, which in this case should really just be hired or archive. We then use those values to make it consistent with the table archive. We expect that there should only be one `hired` status, if at all. If there isn't a `hired` status, this endpoint should fail.

Why? There is a query in in the `employer-actions` endpoint for that gets the current status of the user and makes some important decisions from that. We aren't able to properly get this status unless its passed in. So at the very least, this should make archiving more consistent.

### DIFF
--- a/client/src/components/participant-details/track-graduation.js
+++ b/client/src/components/participant-details/track-graduation.js
@@ -13,8 +13,16 @@ import { postHireStatuses } from '../../constants';
 import { AuthContext } from '../../providers';
 import { useToast } from '../../hooks';
 import { formatCohortDate } from '../../utils';
+import { participantStatus } from '../../constants';
 // Helper function to call archive participant service
-const handleArchive = async (participantId, openToast, dispatchFunction, additional = {}) => {
+const handleArchive = async (
+  participantId,
+  openToast,
+  dispatchFunction,
+  additional = {},
+  siteId = null,
+  currentStatusId = null
+) => {
   const response = await fetch(`${API_URL}/api/v1/employer-actions`, {
     method: 'POST',
     headers: {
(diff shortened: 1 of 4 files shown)


client/src/components/participant-details/track-graduation.js
```diff
@@ -179,7 +193,17 @@ export const TrackGraduation = (props) => {
                 }}
                 onSubmit={async (values) => {
                   setShowArchiveModal(false);
-                  await handleArchive(props?.participant?.id, openToast, dispatchFunction, values);
+                  const hiredStatus = props?.participant?.latestStatuses?.find(
+                    (status) => status.status === participantStatus.HIRED
```
fw-noel, line 197: We can assume that of all current statuses returned, there is only one hired for a specific site, at most.

server/routes/participant.js
```diff
@@ -453,7 +458,15 @@ employerActionsRouter.post(
       participant_id: participantId,
       status,
     });
-    return res.status(201).json({ data: result });
+    let returnStatus = 201;
+    if (
```
fw-noel, line 462: While I don't love this, it seems the best way to properly return an error.

I've seen other areas of the backend return `success: true` from the service function. The problem with that is that it would require a fair bit more refactoring, since `setParticipantStatus` is called by a bunch of other places.

server/services/participant-status.js
```diff
@@ -102,7 +102,7 @@ const setParticipantStatus = async (
       : await tx[collections.PARTICIPANTS_STATUS].findOne({
           participant_id: participantId,
           current: true,
-          or: [{ employer_id: employerId }, { 'data.site IN': user.sites || [] }],
+          employer_id: employerId,
```
fw-noel, line 105: Reverting this to the way it was before. If we properly pass in the status id, we shouldn't need to check this clause for user sites. I'm worried I introduced some badness.

arranfw (approved):

alva-fresh (approved):
Looks good!

PR #819, opened 2022-08-26
HCAP: dependabot alerts

alva-fresh:
https://github.com/bcgov/hcap/security/dependabot

`dicer` discussion: https://freshworks.slack.com/archives/C011G18PW5S/p1661469661319249

PR #821, opened 2022-08-29
fix: update package lock

arranfw:
Looks like package-lock got out of step somehow, this should fix it


arranfw:
This should be resolved by #819

PR #824, opened 2022-08-31
HCAP-1286 | Site Details Allocation tab

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1286
Also included: https://freshworks.atlassian.net/browse/HCAP-1298

Should be added before closing tickets:
- testing (there is no automated testing at all)
- checking at endpoint if the site is assigned to the user requesting information
- some endpoints may be made obsolete by this change and should be marked as deprecated
- look for instances where sites are shown in a list with allocation- references/queries will need to be updated

Result:
image


Makefile
```diff
@@ -14,7 +14,7 @@ export TOOLS_NAMESPACE=$(OS_NAMESPACE_PREFIX)-tools
 export DATABASE_URL=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@localhost:${POSTGRES_PORT}/${POSTGRES_DB}
 
 # Feature flags
-export FEATURE_PHASE_ALLOCATION=true
+export FEATURE_PHASE_ALLOCATION=false
```
arranfw, line 17: Have we have a discussion on turning this off on dev environments? We'll need to make sure we flip the configmaps if that's the case

But I'm not convinced we want to turn this off right away

server/migrations/phase-allocation-migrations/1661284647061_add-phase-tables.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-restricted-syntax, no-await-in-loop */
-const { dbClient, schema } = require('../db');
+const { dbClient, schema } = require('../../db');
```
arranfw, line 2: This should be in the other PR but if these are merged together I won't complain

server/services/phase.js
```diff
@@ -1,9 +1,105 @@
+/* eslint-disable camelcase */
+// disabling camelcase check so that we can manipulate snake_case attributes without changing structure
+const dayjs = require('dayjs');
+const isBetween = require('dayjs/plugin/isBetween');
 const { dbClient, collections } = require('../db');
+const { getSiteByID } = require('./employers');
+
+dayjs.extend(isBetween);
```
arranfw, line 8: This already happened in `client/src/utils/date.js`, does it need to happen here too?

arranfw (approved):
Looks good, I just want to verify sub-tasks have been created for the TODOs

server/services/phase.js
taraepp: Yep, this is server-side. This duplication is one of the reasons why I'd like client and server to be able to share code.

client/src/services/feature-flag.js
```diff
@@ -4,11 +4,13 @@ export const flagKeys = {
...
 
-export const FeatureFlag = ({ featureKey, children }) => {
+export const featureFlag = (featureKey) => {
   // Values come from the API as booleans, no need to check for === 'true'
-  const featureValue = store.get(featureKey);
+  return store.get(featureKey);
+};
 
-  if (featureValue) {
+export const FeatureFlag = ({ featureKey, children }) => {
```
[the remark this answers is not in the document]
taraepp: I like this name change. 👍

client/src/services/site.js
```diff
@@ -1,6 +1,19 @@
...
 
+export const fetchSitePhases = async (siteId) => {
+  const response = await fetch(`${API_URL}/api/v1/phase-allocation/${siteId}`, {
+    headers: { Authorization: `Bearer ${store.get('TOKEN')}` },
+    method: 'GET',
+  });
+  let phases = [];
+  if (response.ok) {
+    phases = await response.json();
+    phases = phases.data;
```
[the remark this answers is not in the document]
taraepp: Good point, I tidied it up a bit. I'd say it's still flawed- if it either fails to fetch the site's participants (fetchSiteRows is the next function in this file) or fails to fetch phases, in both cases it returns [] and doesn't toast or otherwise communicate to the user.

server/routes/phase-allocation.js
```diff
@@ -4,10 +4,39 @@ const logger = require('../logger.js');
...
+  [
+    keycloak.allowRolesMiddleware('health_authority', 'ministry_of_health', 'employer'),
+    keycloak.getUserInfoMiddleware(),
+  ],
+  asyncMiddleware(async (req, res) => {
+    const { hcapUserInfo: user } = req;
+    const siteId = parseInt(req.params.id, 10);
+    let result;
+    // TODO: check that user has access to siteId
+    if (siteId) {
```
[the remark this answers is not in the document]
taraepp: Pretty much just bad requests, which are probably impossible to send through the FE. Leaving it without _any_ error checking on the params feels wrong to me but I've fixed it up so that at least it makes sense (previously it ran a map function on an undefined in the case of !siteId which is no bueno)

server/services/phase.js
arranfw: AhhI didn't notice they were client/server 👍

fw-noel (approved):

PR #827, opened 2022-09-01
HCAP-1314 Consolidate getSites methods

arranfw:
Notes: 
- Many places in the app refer to `user.sites` to list sites
- `GET /user` retrieved all sites then filtered by site ID, this wasn't sorted

Attn:
- In order to properly define test cases in isolation I needed to close and open the db between test cases which adds >10s to this test suite. This is because of some strange behaviour from Massive when attempting to clear the employer_sites table.
- When switching to use `saveSites` I noticed a difference in the validation schemas for the various site editing/creation endpoints which led me to create `batchSiteObject`. I think this isn't great and I'd love feedback on fixing this. Should we make large scale changes to validation/controllers/client to address these inconsistencies?

Changes:
- I consolidated the two getSites methods and handled the access for different roles in that function
- Limit fields on getSites to all that's currently needed on FE

### DIFF
--- a/server/services/employers.js
+++ b/server/services/employers.js
@@ -56,7 +56,7 @@ const updateSite = async (id, site) => {
  * @returns list of sites which the user has access to
  */
 
-const getSites = async (user) => {
+const getSitesForUser = async (user) => {
   const criteria = {};
 
   if (user.isHA) {
@@ -129,7 +129,7 @@ module.exports = {
   saveSingleSite,
   saveSites,
   updateSite,
-  getSites,
+  getSitesForUser,
   getSiteByID,
   getSiteDetailsById,
 };
--- a/server/services/employers.js
+++ b/server/services/employers.js
@@ -50,7 +50,29 @@ const updateSite = async (id, site) => {
   return dbClient.db[collections.EMPLOYER_SITES].updateDoc({ id }, changes);
 };
 
-const getSites = async () => dbClient.db[collections.EMPLOYER_SITES].findDoc({});
+/**
+ * Get all accessible sites for a user
+ * @param {*} user user with roles and sites to filter
+ * @returns list of sites which the user has access to
+ */
+
+const getSites = async (user) => {
+  const criteria = {};
+
+  if (user.isHA) {
+    criteria.siteId = user.sites.map((site) => `${site}`); // cast to string
+    criteria.healthAuthority = user.regions;
+  }
+
+  if (user.isEmployer) {
+    criteria.siteId = user.sites.map((site) => `${site}`); // cast to string
+  }
+
+  return dbClient.db[collections.EMPLOYER_SITES].findDoc(criteria, {
+    order: [{ field: `siteName`, direction: 'asc' }],
+    fields: ['siteId', 'siteName', 'operatorName', 'healthAuthority', 'postalCode', 'allocation'],
+  });
+};
 
 const getSiteDetailsById = async (id) => {
   const site = await dbClient.db[collections.EMPLOYER_SITES].findDoc({ id });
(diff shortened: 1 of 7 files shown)


server/services/employers.js
```diff
@@ -50,7 +50,29 @@ const updateSite = async (id, site) => {
...
 };
 
-const getSites = async () => dbClient.db[collections.EMPLOYER_SITES].findDoc({});
+/**
+ * Get all accessible sites for a user
+ * @param {*} user user with roles and sites to filter
+ * @returns list of sites which the user has access to
+ */
+
+const getSites = async (user) => {
```
fw-noel, line 59:
```suggestion
const getSitesForUser = async (user) => {
```

server/services/employers.js
```diff
@@ -50,7 +50,29 @@ const updateSite = async (id, site) => {
...
+  const criteria = {};
+
+  if (user.isHA) {
+    criteria.siteId = user.sites.map((site) => `${site}`); // cast to string
```
fw-noel, line 63: Why is the casting necessary?

server/tests/employer.site.e2e.test.js
```diff
@@ -78,16 +78,24 @@ describe('api-e2e tests for /employer-sites route', () => {
     expect(res.status).toEqual(200);
   });
 
-  it('should get sites', async () => {
+  it.only('should get sites', async () => {
```
fw-noel, line 81: temporary artifact?

server/services/employers.js
arranfw: I looked into this a bit further and it looks like some MassiveJS quirkiness, I double checked the database and both of these values are stored as numbers but passing in an array of numbers to `criteria.siteId` results in this error from Massive

```
error: operator does not exist: text = integer
```

I tried [casting ](https://massivejs.org/docs/criteria-objects#casting) `siteId` to INT in the `fields` option which gets it to return a number but still fails when trying to query with a number 🙃 

Without casting: 
image

image


With casting:

image

image

server/tests/employer.site.e2e.test.js
arranfw: https://github.com/bcgov/hcap/pull/827/commits/93a48a3efdb9f3b272f18b7bdd04a143572fccfa

server/services/employers.js
arranfw: https://github.com/bcgov/hcap/pull/827/commits/aac66977b1910486f3f670ae584deba271aafa01

fw-noel (approved):

server/tests/employer.site.e2e.test.js
```diff
@@ -29,19 +29,54 @@ const siteObject = ({ id, name }) => ({
...
+/**
+ * This method is required because different endpoints for creating/editing
+ * sites have different names for similar fields (ie: siteContactPhone vs siteContactPhoneNumber)
+ *
+ * This method wraps siteObject() and renames those fields to pass validation
+ *
+ * @param {*} object containing an id and name used to create a test site
+ * @returns an object containing the given id and name and the rest of the required fields for batch saving
+ */
+const batchSiteObject = ({ id, name }) => {
```
taraepp, line 41: Do you think it would be possible to rename the attributes in the different places so that they match? Or are there constraints (other than time, resources) preventing us from doing so?

If it's possible, I'm in favour of putting a ticket in the backlog to clean that up.

taraepp (approved):

server/tests/employer.site.e2e.test.js
arranfw: It looks like a decent sized chunk of understanding/work to do, I agree it would be good to add something to the backlog

- siteContactPhone: 16 results
- siteContactPhoneNumber: 10 results
- operatorPhone: 34 results

PR #831, opened 2022-09-07
feat: create a cron job for withdrawing expired participants

fw-noel:
We're creating a cron job to clean up expired participants, which we are defining as:
- Participants that were created >6 months ago but never engaged (>500 participants in prod, seemingly)
- Participants who are still marked as "interested", but were rejected for one reason or another > 6 months ago, with no other status since (>1200 participants in prod, seemingly)

We're bound to have missed a case or two given the complicated nature of statuses, but this hopefully gets a good start. Need to confirm with Amanda / Tania if this is in the ballpark.

I may include tests if there's time, but I wanted to put up the draft so that people can start poking holes at the logic.

Also, I'm pretty sure the cron jobs don't get created automatically at the moment.


server/services/clean-up.js
```diff
@@ -57,9 +57,130 @@ const cleanStaleInProgressParticipant = async () => {
...
+          data_obj = JSONB_SET('{}'::JSONB, '{cleanupDate}', to_jsonb(NOW()));
+        ELSE
+          data_obj = JSONB_SET(status_rec.data::JSONB, '{cleanupDate}', to_jsonb(NOW()));
+        END IF;
+        UPDATE participants_status SET data = data_obj, current = false WHERE id = status_rec.id;
+      END LOOP;
+      -- Withdraws participant and sets an expiration date
+      body_obj = JSONB_SET(participant_rec.body::JSONB, '{expiredDate}', to_jsonb(NOW()));
+      body_obj = body_obj || '{"interested" : "withdrawn"}';
+      UPDATE participants SET body = body_obj, updated_at = NOW() WHERE id = participant_rec.id;
```
arranfw, line 169: this looks like a bit of code duplication, there might not be a better way to do it perf-wise but couldn't we create a more generic function that invalidates statuses for a given list of users?

server/services/clean-up.js
```diff
@@ -57,9 +57,130 @@ const cleanStaleInProgressParticipant = async () => {
...
+      FROM participants_status ps
+      WHERE p.id = ps.participant_id
+    );
+  `;
+  await dbClient.db.query(pushUnengagedParticipants);
+
+  const pushRejectedParticipants = `
+    INSERT INTO expired_participants_table
+    SELECT
+      p.id, max(ps.created_at) AS last_updated, TRUE as perviously_engaged
```
arranfw, line 93:
```suggestion
      p.id, max(ps.created_at) AS last_updated, TRUE as previously_engaged`
```

server/scripts/expire-participants.js
```diff
@@ -0,0 +1,34 @@
...
+    console.log(`Connected to database`);
+    await createOpenParticipantsTable();
+    const report = await getExpiredParticipants();
+    if (report.length > 0) {
+      console.log(`Withdrawing ${report.length} open, expired participants`);
+      console.table(report);
+    } else {
+      console.log('No expired open participants found');
+    }
+    await expiredParticipantsStatusInvalidation();
```
arranfw, line 25: Do we still want to run this if `report.length === 0`?

server/services/clean-up.js
```diff
@@ -57,9 +57,130 @@ const cleanStaleInProgressParticipant = async () => {
   return results;
 };
 
+/**
+ * @description - Creates a table of all participants that should currently be Open - Never engaged, or current status = rejected / reject_ack
+ * @returns {Promise<Array>}
+ */
+const createOpenParticipantsTable = async () => {
```
arranfw, line 64: is `openParticipantsTable` the right term for this?

server/scripts/expire-participants.js
arranfw: I don't love having two cleanup jobs named with synonyms, how is `expired` different from `stale`?

Can we come up with naming that differentiates the two?

maybe `createStaleOpenParticipantsTable`/`cleanStaleOpenParticipants` or something?

arranfw (commented):
Could we go over the SQL loop together?

server/services/clean-up.js
fw-noel: Nope, outdated when I renamed / reworked everything, thanks

server/scripts/expire-participants.js
fw-noel: Yeah, expired vs. stale, hard to distinct between the two even though there is a distinction.

I like cleanStaleOpen vs. cleanStaleInProgress (the existing job)

server/services/clean-up.js
fw-noel: What would that function look like? A piece of sql that we call in here? Or a separate query that we call after all this?

arranfw: I was thinking a loop that took in an array of IDs and then invalidated all participant statuses for those IDs

Also I didn't read this super closely but is it necessary to be a loop? 

why not
```
update participants_status set current = false where participant_id = foo
```

fw-noel: Because in many cases, we're adding to the json data of a participant status to add the clean up date. I don't think that can be done in a general query for multiple users.

server/services/clean-up.js
```diff
@@ -57,9 +57,129 @@ const cleanStaleInProgressParticipant = async () => {
...
+    )
+    GROUP BY p.id;
+  `;
+  await dbClient.db.query(pushRejectedParticipants);
+};
+
+/**
+ * Gets all participants from temporary table who haven't been updated in 6 months
+ * Relies on createStaleOpenParticipantsTable
+ * @returns {void}
```
taraepp, line 121: nitpick: doesn't look like this returns void

taraepp (approved):
Looks good to me! Minor nitpick on a comment.
The "code smells" that SC picked up don't look legitimate to me.

arranfw (approved):

alva-fresh (approved):

arranfw (approved):

server/services/clean-up.js
```diff
@@ -101,24 +103,48 @@ const createStaleOpenParticipantsTable = async () => {
...
+        ps.created_at > (NOW() - interval '6 month')
+        OR ps.current = true
+      )
       GROUP BY ps.participant_id
     )
     GROUP BY p.id;
   `;
   await dbClient.db.query(pushRejectedParticipants);
+
+  // Anyone who was affected by the stale, in progress script will have only CURRENT=FALSE statuses
```
arranfw, line 122: Why do they need to be addressed by this script? Just to set `interested = 'withdrawn'`?

arranfw (dismissed):

server/services/clean-up.js
```diff
@@ -108,10 +109,10 @@ const createStaleOpenParticipantsTable = async () => {
       SELECT DISTINCT ON (ps.participant_id)
       ps.participant_id
       FROM participants_status ps
-      WHERE ps.status NOT IN ('rejected', 'reject_ack')
-      AND (
-        ps.created_at > (NOW() - interval '6 month')
-        OR ps.current = true
+      WHERE ps.created_at > (NOW() - interval '3 month')
```
arranfw, line 112: accidental change from 6mos?

server/services/clean-up.js
```diff
@@ -169,7 +170,8 @@ const dropStaleOpenParticipantsTable = async () => {
  * @returns {void}
  */
 const invalidateStaleOpenParticipants = async () => {
-  const updateStatement = `DO $$
+  const updateStatement = `START TRANSACTION;
```
arranfw, line 173: This is what I was thinking about for transactions 

https://massivejs.org/docs/tasks-and-transactions

So perhaps something like

```
db.withTransaction(async tx => {
  await tx.query(updateStatement);
});

```

arranfw (changes requested):

server/services/clean-up.js
fw-noel: Oh, okay

arranfw (approved):

PR #834, opened 2022-09-07
HCAP-376 Show Site Hires

arranfw:
https://freshworks.atlassian.net/browse/HCAP-376

- Rewrote the sites query because `fields` causes a lot of unexpected issues within MassiveJS
- Add "Hires" column to sites table


taraepp (approved):
Approved!

For the sake of posterity, I would like to note that the 'fields' parameter in MassiveJS is incompatible with joined queries, as it automatically adds 'id' to the SELECT list, which will always be an ambiguous field that causes the query to fail.

PR #838, opened 2022-09-09
Hcap 1337 region sites

fw-noel:
Add a query on the endpoint to be able to get regions, and separate the region specific query.

I'm sorry that I couldn't avoid a flag, but I am trying to limit it to an extent!

https://freshworks.atlassian.net/browse/HCAP-1337

### DIFF
--- a/server/routes/employer-sites.js
+++ b/server/routes/employer-sites.js
@@ -8,6 +8,7 @@ const {
   saveSingleSite,
   updateSite,
   getSitesForUser,
+  getSitesForRegion,
   getAllSites,
   getSiteByID,
 } = require('../services/employers');
@@ -111,9 +112,15 @@ router.get(
     routeRedirect({ redirect: '/api/v1/employer-sites/details', match: 'employer-sites-detail' }),
   ],
   asyncMiddleware(async (req, res) => {
+    const { regions } = req.query;
     const { hcapUserInfo: user } = req;
 
-    const result = await getSitesForUser(user);
+    let result;
+    if (regions && user.isHA && user.regions.length > 0) {
+      result = await getSitesForRegion(user.regions);
+    } else {
+      result = await getSitesForUser(user);
+    }
 
     logger.info({
       action: 'employer-sites_get',
--- a/server/routes/employer-sites.js
+++ b/server/routes/employer-sites.js
@@ -103,6 +103,32 @@ router.get(
   })
 );
 
+router.get(
+  '/region',
+  [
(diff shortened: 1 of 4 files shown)


server/routes/employer-sites.js
```diff
@@ -87,8 +89,38 @@ router.get(
...
+      },
+      sites_accessed: result.map((site) => site.siteId),
+    });
+    return res.json({ data: result });
+  })
+);
+
+// Read: Get All User Sites
+router.get(
+  '/user',
```
arranfw, line 108: Why not sites/region?

I guess that wouldn't let us use multiple sites but I don't think that's a real use case is it?

arranfw (dismissed):
Left a suggestion for route structure but not a blocker

alva-fresh (approved):

server/routes/employer-sites.js
fw-noel: If we do, on the frontend we'll have to differentiate the user on the Sites component. As in

if user.isHA, request sites/region
else if user.isMoH, request sites

I'm leaning towards not in this case. Don't love the query flag, but there's a little annoyance no matter which way we try it

arranfw: Ah fair enough 👍

server/routes/employer-sites.js
```diff
@@ -111,9 +112,15 @@ router.get(
     routeRedirect({ redirect: '/api/v1/employer-sites/details', match: 'employer-sites-detail' }),
   ],
   asyncMiddleware(async (req, res) => {
+    const { regions } = req.query;
     const { hcapUserInfo: user } = req;
 
-    const result = await getSitesForUser(user);
+    let result;
+    if (regions && user.isHA && user.regions.length > 0) {
```
arranfw, line 119: After looking at this again, I think I'd prefer having this logic on the FE, then we have a more understandable

```
if (HA)
  GET (/sites/region)
else 
  GET (/sites)
```
and we can more easily handle this authorization in middleware

Still just a suggestion and not a blocker

arranfw (approved):

server/routes/employer-sites.js
taraepp: Yeah I'd like it as just a separate endpoint and keep all the logic of which endpoint to hit on the FE.

fw-noel: Fair enough

taraepp (approved):
👍

arranfw (approved):

PR #839, opened 2022-09-09
fix: remove base url

fw-noel:
https://freshworks.atlassian.net/browse/HCAP-1343

TBH, I'm not sure why this started failing, but it did. No where else seems to really use that baseURL pattern so this is more consistent anyway.

### DIFF
--- a/client/src/services/psi-cohort.js
+++ b/client/src/services/psi-cohort.js
@@ -2,13 +2,11 @@ import store from 'store';
 import { API_URL } from '../constants';
 import { formatCohortDate } from '../utils';
 
-const baseURL = `${API_URL}/api/v1`;
-
 const getCohortAvailAbleSize = (cohort) =>
   cohort.cohort_size - (cohort.participantsCohorts?.length || 0);
 
 export const getPsi = async () => {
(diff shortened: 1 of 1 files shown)


client/src/services/psi-cohort.js
```diff
@@ -2,13 +2,11 @@ import store from 'store';
...
 import { formatCohortDate } from '../utils';
 
-const baseURL = `${API_URL}/api/v1`;
-
 const getCohortAvailAbleSize = (cohort) =>
   cohort.cohort_size - (cohort.participantsCohorts?.length || 0);
 
 export const getPsi = async () => {
-  const response = await fetch(`${baseURL}/psi/with-cohorts`, {
+  const response = await fetch(`${API_URL}/api/v1/psi/with-cohorts`, {
```
alva-fresh, line 9: It's still failing for me even after updating the base url for some reason
only I get a different error
Screen Shot 2022-09-09 at 4 51 01 PM

taraepp (approved):
Fix worked for me, but we should find out why it didn't for Alva.

arranfw (approved):
@alva-fresh This fix looks like it should work for the bug in the ticket, your error is hopefully not related here, can you take some time to investigate where that error is coming from?

PR #841, opened 2022-09-12
HCAP-1317 Feature flag docs

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1317

- Adds documentation on how to set feature flags
- Remove feature flags from makefile, treat `.env` as the source of truth for local development


taraepp (approved):
👍

fw-noel (approved):
What did you do to get this to build properly? I thought that was why we added the env in the makefile in the first place

arranfw:
> What did you do to get this to build properly? I thought that was why we added the env in the makefile in the first place

I was wrong in my understanding, the makefile was only for local development and nothing is injected in the pipeline, they're just pulled from configmaps. So rather than define them in .env and the makefile we can just just .env for local

PR #843, opened 2022-09-12
HCAP-1317 feature flag adjustments

arranfw:
Here are a couple of changes I forgot about to make to make things more consistent


taraepp (approved):
👍

PR #844, opened 2022-09-12
HCAP-1345 Add subquery to skip counting duplicate participants

arranfw:
I missed filtering duplicate participants, I moved from a join to a subquery to properly count hires. The subquery is modeled after https://github.com/bcgov/hcap/blob/dev/server/services/employers.js#L130

https://freshworks.atlassian.net/browse/HCAP-1345

### DIFF
--- a/server/services/employers.js
+++ b/server/services/employers.js
@@ -75,16 +75,24 @@ const getSitesWithCriteria = async (additionalCriteria, additionalCriteriaParams
         employer_sites.body -> 'healthAuthority' as "healthAuthority",
         employer_sites.body -> 'postalCode' as "postalCode",
         employer_sites.body -> 'allocation' as "allocation",
-        count(ps.id) :: INT as "hireCount"
+        (
+          SELECT
+              count(*)
+          FROM
+              participants_status ps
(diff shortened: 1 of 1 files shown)


taraepp (approved):
👍

server/services/employers.js
```diff
@@ -75,16 +75,24 @@ const getSitesWithCriteria = async (additionalCriteria, additionalCriteriaParams
...
         employer_sites.body -> 'postalCode' as "postalCode",
         employer_sites.body -> 'allocation' as "allocation",
-        count(ps.id) :: INT as "hireCount"
+        (
+          SELECT
+              count(*)
+          FROM
+              participants_status ps
+              LEFT outer JOIN participants_status ps_da on ps.participant_id = ps_da.participant_id
+              AND ps_da.status = 'archived'
```
fw-noel, line 84: Sorry, why archived?

arranfw: With the self join this is a non-straightforward way to get hired statuses where people aren't archived with the following line  `and ps_da.data ->> 'type' is null`

This a based off the other count for hired participants https://github.com/bcgov/hcap/blob/dev/server/services/employers.js#L130

fw-noel (approved):

PR #845, opened 2022-09-12
HCAP-1341 Remove additional unneeded sorting

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1341

I guess lodash and postgres treat spaces differently in their sorting algorithms

- removed extra sort on FE for sitenames

### DIFF
--- a/client/src/components/modal-forms/EditRosSiteForm.js
+++ b/client/src/components/modal-forms/EditRosSiteForm.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import _orderBy from 'lodash/orderBy';
 
 import Alert from '@material-ui/lab/Alert';
 import { Box } from '@material-ui/core';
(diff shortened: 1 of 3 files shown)


alva-fresh (commented):
We also use it in `SelectProspectingSiteForm.js` and `src/pages/private/UserView.js`. Do you think it's a good idea to remove it in there as well?

arranfw:
> 

Great call! thank you!

taraepp:
I think this will deal with _some_ of the sorting behaviour, but postgres's sort is also case-sensitive so I'm thinking there may still be some issues. In the raw query you can do lower()/upper() on the order by field but not sure if massive will cooperate.
image

arranfw:
> I think this will deal with _some_ of the sorting behaviour, but postgres's sort is also case-sensitive so I'm thinking there may still be some issues. In the raw query you can do lower()/upper() on the order by field but not sure if massive will cooperate. !

It should at least get the sorting to be consistent between the two components which was my understanding of what this ticket was describing

alva-fresh (dismissed):

taraepp (dismissed):
👍 
I think with 1347 merged you should be able to remove _orderBy entirely from UserView.

fw-noel (approved):

PR #846, opened 2022-09-12
HCAP-1347 | Sort order on sites table

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1347
- issue caused by lodash _orderBy function in FE
- made a util function to deal with it
- I replaced the function in ONE place with a new one (ie: to deal with the bug)
- but a search for _orderBy shows it being used in 8 different files, probably causing the same issue there

### DIFF
--- a/client/src/pages/private/SiteTable.js
+++ b/client/src/pages/private/SiteTable.js
@@ -20,7 +20,7 @@ import {
 } from '../../constants';
 import { TableFilter } from '../../components/generic/TableFilter';
 import { useToast } from '../../hooks';
-import { handleReportDownloadResult } from '../../utils';
+import { handleReportDownloadResult, orderByCaseInsensitive } from '../../utils';
 import { AuthContext } from '../../providers';
 import { FeatureFlaggedComponent, flagKeys } from '../../services';
 import { fetchRegionSiteRows, fetchSiteRows } from '../../services/site';
@@ -168,7 +168,13 @@ export default ({ sites, viewOnly }) => {
     );
   }, [roles]);
 
-  const sort = (array) => _orderBy(array, [orderBy, 'operatorName'], [order]);
+  const sort = (array) => {
+    return _orderBy(
+      array,
+      [(item) => orderByCaseInsensitive(item, orderBy), 'operatorName'],
+      [order]
+    );
+  };
 
   useEffect(() => {
     if (sites) {
--- a/client/src/pages/private/SiteTable.js
+++ b/client/src/pages/private/SiteTable.js
@@ -19,7 +19,7 @@ import {
 } from '../../constants';
 import { TableFilter } from '../../components/generic/TableFilter';
 import { useToast } from '../../hooks';
-import { handleReportDownloadResult, sortRows } from '../../utils';
+import { handleReportDownloadResult, sortObjects } from '../../utils';
 import { AuthContext } from '../../providers';
 import { FeatureFlaggedComponent, flagKeys } from '../../services';
 import { fetchRegionSiteRows, fetchSiteRows } from '../../services/site';
@@ -167,9 +167,7 @@ export default ({ sites, viewOnly }) => {
     );
   }, [roles]);
 
-  const sort = (array) => {
-    return sortRows(array, orderBy, order);
-  };
+  const sort = (array) => sortObjects(array, orderBy, order);
 
   useEffect(() => {
     if (sites) {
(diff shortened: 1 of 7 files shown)


alva-fresh (dismissed):

client/src/pages/private/SiteTable.js
```diff
@@ -168,7 +168,13 @@ export default ({ sites, viewOnly }) => {
     );
   }, [roles]);
 
-  const sort = (array) => _orderBy(array, [orderBy, 'operatorName'], [order]);
+  const sort = (array) => {
```
arranfw, line 171: Do we need this sort at all? I would think we should be handling sorting on the back end in most cases, sites are surely sorted from the API now right?

I guess this component isn't doing API pagination or anything so it might require a bit of a rewrite to properly pass in pagination values from the back end. If we can't fit it into this PR that would be a great enhancement in the future

arranfw: In the case that we can't do server side sortation, I vote we get rid of lodash's orderBy and write our own that will function a bit more like we want, this solution is mostly fine but it won't sort uppercase and lowercase letters perfectly, ie:

With this method:
`a, A, a => a, A, a`

When it should be  
`a, A, a => a, a, A`

It's a pretty low priority change so don't consider this a blocker but I'd recommend writing a function that makes use of `localeCompare`, something like

```
/**
 * Sorts a list of objects based on the values of a given key
 * 
 * @param {*} items list of objects to sort
 * @param {*} key object key to sort by
 * @param {*} direction asc or desc, will not sort if the value is anything else
 * @returns sorted list of objects based on given key and direction
 */
const orderBy = (items, key, direction) => {
  return items.sort((itema, itemb) => {
    switch (direction) {
      case 'asc':
        return itema[key].localeCompare(itemb[key], 'en', { numeric: true });
      case 'desc':
        return itemb[key].localeCompare(itema[key], 'en', { numeric: true });
      default:
        return 0;
    }
  });
};
```

client/src/utils/gen-util.js
```diff
@@ -4,3 +4,15 @@ export const keyedString = (str, keyValues) =>
...
+
+/**
+ * for use with _orderBy lodash function
+ * _orderBy sorts in a case sensitive way like: [1, Z, a, b, c]
+ * so check if field is not numeric, lowercase it. toString for safety.
+ * @param {{}} item:
+ * @param {string} orderBy: field to order by
+ * @returns []
+ */
+export const orderByCaseInsensitive = (item, orderBy) => {
```
arranfw, line 16: This function looks like it's conditionally lowercasing property values rather than ordering things, trying to come up with an accurate name for this function is a bit difficult which makes me feel like we're a bit off-pattern here

client/src/pages/private/SiteTable.js
arranfw: Nit: I don't love this function name

taraepp: We do need the sort function, it's what's called on the FE when the user sorts by clicking on a column (participants table gets sorted on BE but sites table and others get sorted on FE). 
I will get rid of the lodash function. It seems silly to use a library function when it doesn't even do what we need.
Pasting your function worked so I'm not going to reinvent the wheel, but I'm renaming to avoid conflicts with the 'orderBy' variable name that we use everywhere with it.

arranfw (approved):
Awesome!

client/src/utils/gen-util.js
```diff
@@ -4,3 +4,39 @@ export const keyedString = (str, keyValues) =>
...
+
+/**
+ * Sorts a list of objects based on the values of a given key
+ *
+ * @param {[{}]} items list of objects to sort
+ * @param {string} key object key to sort by
+ * @param {string} direction asc or desc, will not sort if the value is anything else
+ * @returns sorted list of objects based on given key and direction
+ */
+export const sortRows = (items, key, direction) => {
```
arranfw, line 28: Total non-blocker, what do you think about `sortObjects` to decouple the naming from tables?

arranw (dismissed):

client/src/utils/gen-util.js
taraepp: I knew there would be a better suggestion for a name! I think I'll also default sort to 'asc'.

arranfw (approved):

alva-fresh (approved):

PR #847, opened 2022-09-13
HCAP-1349: cohort available seats fix

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1349

This request is for adding an extra join to select only those participants who have a status assigned to them

### DIFF
--- a/client/src/pages/private/CohortTable.js
+++ b/client/src/pages/private/CohortTable.js
@@ -60,6 +60,8 @@ export default ({ cohorts, editCohortAction }) => {
       case 'end_date':
         return formatCohortDate(row[columnId]);
       case 'remaining_seats':
+        console.log(row.participantStatusJoin);
+        console.log(row.participants);
         return row['cohort_size'] - row['participants'].length;
       default:
         return row[columnId];
--- a/client/src/pages/private/CohortTable.js
+++ b/client/src/pages/private/CohortTable.js
@@ -60,8 +60,6 @@ export default ({ cohorts, editCohortAction }) => {
       case 'end_date':
         return formatCohortDate(row[columnId]);
       case 'remaining_seats':
-        console.log(row.participantStatusJoin);
-        console.log(row.participants);
         return row['cohort_size'] - row['participants'].length;
       default:
         return row[columnId];
--- a/server/services/cohorts.js
+++ b/server/services/cohorts.js
@@ -62,6 +62,7 @@ const getCohortParticipants = async (cohortId) =>
     .find({
       'cohortParticipantsJoin.cohort_id': cohortId,
       'participantStatusJoin.current': true,
+      'participantStatusJoin.status': ['hired', 'archived'],
     });
 
 // Get all Cohorts associated with a specific PSI
@@ -85,7 +86,12 @@ const getPSICohorts = async (psiID) =>
     })
     .find({
       psi_id: psiID,
+      // Additional condition to filter out the participants who don't have status assigned to them
       'participantStatusJoin.current !=': false,
+      or: [
+        { 'participantStatusJoin.status': ['hired', 'archived'] },
+        { 'participantStatusJoin.status': null },
+      ],
     });
 
 const mapDataToCohort = (cohort) => {
--- a/server/services/cohorts.js
+++ b/server/services/cohorts.js
@@ -76,10 +76,20 @@ const getPSICohorts = async (psiID) =>
           cohort_id: 'id',
         },
       },
+      participantStatusJoin: {
+        type: 'LEFT OUTER',
+        relation: collections.PARTICIPANTS_STATUS,
+        on: {
+          participant_id: 'participants.participant_id',
+        },
+      },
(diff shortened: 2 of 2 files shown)


server/services/cohorts.js
```diff
@@ -78,6 +79,7 @@ const getPSICohorts = async (psiID) =>
     })
     .find({
       psi_id: psiID,
+      or: [{ 'participants.is_current': true }, { 'participants.is_current': null }],
```
arranfw, line 82: Is this a different bug that you're also fixing? This code hasn't been change recently has it?

alva-fresh: I'd say it's a different bug in a way. It selects participants who are no longer assigned to this cohort (`is_current` = false), hence the number of available participants is wrong.
This ticket says that the number of available participants is different in cohort table/cohort details. I'm trying to figure out why and I thought this could be one of the reasons

alva-fresh: Nevermind, Tavy says it's not how it should be - https://teams.microsoft.com/l/message/19:IUaAmNEwTGmcqt3CjlNCrc_j7pTI3rB5QRpVrTFxNHQ1@thread.tacv2/1663111279482?tenantId=5b973f99-77df-4beb-b27d-aa0c70b8482c&groupId=3f22a178-b2af-4d1a-99cd-bc1781c11210&parentMessageId=1663111279482&teamName=HCAP&channelName=General&createdTime=1663111279482&allowXTenantAccess=false
I'll update the query 👍

client/src/pages/private/CohortTable.js
```diff
@@ -60,6 +60,8 @@ export default ({ cohorts, editCohortAction }) => {
       case 'end_date':
         return formatCohortDate(row[columnId]);
       case 'remaining_seats':
+        console.log(row.participantStatusJoin);
```
fw-noel, line 63: Logs!

alva-fresh: Yes, testing them now - I'll remove them shortly

server/services/cohorts.js
```diff
@@ -75,9 +75,17 @@ const getPSICohorts = async (psiID) =>
...
+        type: 'LEFT OUTER',
+        relation: collections.PARTICIPANTS_STATUS,
+        on: {
+          participant_id: 'participants.participant_id',
+        },
+      },
     })
     .find({
       psi_id: psiID,
+      'participantStatusJoin.current !=': false,
```
arranfw, line 88: Why the double negative here instead of just `true`?

arranfw: Is there more criteria we can go off of? I'm wondering if we should be expecting certain statuses for cohort participants. What status should cohort participants be? Just checking for statuses that are current doesn't seem straight forward

alva-fresh: When setting `'participantStatusJoin.current': true,` it'll select all cohorts that have participants with statuses set to `true` and omit all the cohorts that don't have any participants at all. So, instead of doing `or: ['participantStatusJoin.current': true, 'participantStatusJoin.current': null]` I'm checking that it's not `false`.
Also, we're not looking for any particular statuses, just checking that participants have some statuses assigned to them. Alternatively, we can check is the status is `hired` or `archived`

arranfw: I guess there aren't any cases where they wouldn't be hired or archived at this point right? This is definitely a tough piece to understand

Which participants are we trying to omit by adding this condition?

arranfw: But how would a participant assigned to a cohort not have any participant statuses?

alva-fresh: I suspect it might be some legacy data in dev and test

arranfw: I hesitate to change application code based on data that we don't expect to exist, @fw-noel how do you feel about this? As an alternative we could not fix this here but remove the problematic records in dev/test

At the very least I'd advocate for adding a comment that this is an additional guard to protect against invalid records

alva-fresh: Adding a comment makes total sense to me 👍
What about check for `hired` and `archived` participants? I feel like it might be a good idea to add it, since we're already using it in the reports

server/services/cohorts.js
```diff
@@ -75,9 +76,22 @@ const getPSICohorts = async (psiID) =>
...
+        type: 'LEFT OUTER',
+        relation: collections.PARTICIPANTS_STATUS,
+        on: {
+          participant_id: 'participants.participant_id',
+        },
+      },
     })
     .find({
       psi_id: psiID,
+      // Additional condition to filter out the participants who don't have status assigned to them
```
arranfw, line 89: This comment reiterates what the code does but not why, we should aim to describe the intent of the code with comments

So acknowledging this as a safeguard against potential bad data and even referring to the ticket/GH discussion could offer more insight to a future reader

arranfw (approved):

fw-noel (approved):

PR #848, opened 2022-09-13
HCAP-1352 | Nulls breaking sort

taraepp:
Ticket: https://freshworks.atlassian.net/browse/HCAP-1352
Chose to treat nulls like empty string, as that should be consistent for user.

### DIFF
--- a/client/src/utils/gen-util.js
+++ b/client/src/utils/gen-util.js
@@ -18,6 +18,7 @@ export const sortObjects = (items, key, direction = 'asc') => {
     return [];
   }
   return items.sort((itema, itemb) => {
+    // coalesce null fields to empty string
     const a = itema[key] ?? '';
     const b = itemb[key] ?? '';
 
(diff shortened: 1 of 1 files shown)


client/src/utils/gen-util.js
```diff
@@ -18,11 +18,14 @@ export const sortObjects = (items, key, direction = 'asc') => {
     return [];
   }
   return items.sort((itema, itemb) => {
+    const a = itema[key] ?? '';
```
arranfw, line 21:
```suggestion
    // guard against null values
    const a = itema[key] ?? '';
```

arranfw (dismissed):
A comment may help here, but isn't strictly necessary

client/src/utils/gen-util.js
taraepp: I had one actually and removed it before I committed. I'll add one back

arranfw: Hmm, this comment is describing what the code is doing rather than explaining what we want the code to do
Typically we want the latter over the former

arranfw (approved):

PR #849, opened 2022-09-14
HCAP-1355 Calculate remaining seats on the server

arranfw:
- Calculating this field on the server allows us to treat it like any other field on the FE which fixes our sorting problem

https://freshworks.atlassian.net/browse/HCAP-1355


alva-fresh (approved):
Looks good to me!

PR #851, opened 2022-09-15
HCAP-1349: modification to cohort participants query

alva-fresh:
https://freshworks.atlassian.net/browse/HCAP-1349

In test db the query was selecting multiple entries for the same participant possibly due to `JOIN`. After setting `SELECT COUNT(DISTINCT "participants"."id")"` the number of available seats became correct for both queries.
This request is for filtering out duplicates which causing incorrect count. Yet I don't like this solution, so happy to hear thoughts on this 

Including 2 possible solutions here as separate commits based on further findings: https://teams.microsoft.com/l/message/19:IUaAmNEwTGmcqt3CjlNCrc_j7pTI3rB5QRpVrTFxNHQ1@thread.tacv2/1663206188323?tenantId=5b973f99-77df-4beb-b27d-aa0c70b8482c&groupId=3f22a178-b2af-4d1a-99cd-bc1781c11210&parentMessageId=1663204501914&teamName=HCAP&channelName=General&createdTime=1663206188323&allowXTenantAccess=false

Screen Shot 2022-09-14 at 7 10 45 PM

### DIFF
--- a/client/src/pages/private/CohortDetails.js
+++ b/client/src/pages/private/CohortDetails.js
@@ -69,7 +69,16 @@ export default ({ match }) => {
     }
   };
 
-  const getParticipantGraduationStatus = (participantStatuses) => {
+  const getParticipantGraduationStatus = (participantStatuses, participantId, participants) => {
+    // check if any of the participants were re-assigned to the same cohort
+    if (
+      participants.filter((participant) => participant.participant_id === participantId).length > 1
+    ) {
+      // extra validation for the edge case
+      if (!participantStatuses.find((postHireStatus) => postHireStatus.is_current === true)) {
+        return getPostHireStatusLabel(participantStatuses[participantStatuses.length - 1]);
+      }
+    }
     if (!participantStatuses || participantStatuses.length === 0) return 'Not recorded';
     const graduationStatus = participantStatuses.find(
       (postHireStatus) => postHireStatus.is_current === true
@@ -146,21 +155,25 @@ export default ({ match }) => {
                   renderCell={(columnId, row) => {
                     switch (columnId) {
                       case 'firstName':
-                        return row.body[columnId];
+                        return row.participantsJoin?.[0].body[columnId];
                       case 'lastName':
                         return (
                           <Link
                             component='button'
                             variant='body2'
-                            onClick={() => handleOpenParticipantDetails(row.id)}
+                            onClick={() => handleOpenParticipantDetails(row.participant_id)}
                           >
-                            {row.body[columnId]}
+                            {row.participantsJoin?.[0].body[columnId]}
                           </Link>
                         );
                       case 'siteName':
-                        return row.siteJoin?.body[columnId];
+                        return row.siteJoin?.body?.[columnId];
                       case 'graduationStatus':
-                        return getParticipantGraduationStatus(row.postHireJoin);
+                        return getParticipantGraduationStatus(
+                          row.postHireJoin,
+                          row.participant_id,
+                          rows
+                        );
                       default:
                         return row[columnId];
                     }
--- a/server/services/cohorts.js
+++ b/server/services/cohorts.js
@@ -29,27 +29,27 @@ const getCohort = async (id) =>
   });
 
 const getCohortParticipants = async (cohortId) =>
-  dbClient.db[collections.PARTICIPANTS]
+  dbClient.db[collections.COHORT_PARTICIPANTS]
     .join({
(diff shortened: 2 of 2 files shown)


arranfw:
Does this mean the earlier query was fine and the larger number was incorrect?

alva-fresh:
@arranfw you're totally right - the precious query was fine, although it did fix some inconsistencies in dev. Now this issue is not happening in dev anymore, but still present in test

server/services/cohorts.js
```diff
@@ -97,10 +97,21 @@ const getPSICohorts = async (psiID) => {
...
 
   // calculate remaining cohort seats
-  psiCohorts = psiCohorts.map((cohort) => ({
-    ...cohort,
-    remaining_seats: cohort.cohort_size - cohort.participants.length,
-  }));
+  psiCohorts = psiCohorts.map((cohort) => {
+    const uniqueCohortParticipants = [];
+    // filter out duplicates
+    cohort.participants.forEach((participant) => {
```
arranfw, line 103: It looks like we should be able to alter the query to only join on current participants to fix this issue rather than an additional loop in JS

image

arranfw: Something like

```
.find({
  psi_id: psiID,
  // Additional condition to filter out the participants who don't have status assigned to them
  // This is used as a safeguard against legacy data
  // ref: https://github.com/bcgov/hcap/pull/847
  'participantStatusJoin.current !=': false,
+ 'participants.current': true,
  or: [
    { 'participantStatusJoin.status': ['hired', 'archived'] },
    { 'participantStatusJoin.status': null },
  ],
});
```

arranfw (dismissed):

server/services/cohorts.js
alva-fresh: Not exactly, this condition was present [before](https://github.com/bcgov/hcap/pull/847/commits/9e30b5b8cf07a6538c0aa4e14282fa21a74bd868#diff-eed45eb42a087a4e84846d8d5b06d6ce602ba0d8f8d26e6ae6a6550333c1ea95R64), but I had to remove it because the requirement is to display all participants including those who are no longer assigned to this cohort

client/src/pages/private/CohortDetails.js
```diff
@@ -69,7 +69,16 @@ export default ({ match }) => {
...
-  const getParticipantGraduationStatus = (participantStatuses) => {
+  const getParticipantGraduationStatus = (participantStatuses, participantId, participants) => {
+    // check if any of the participants were re-assigned to the same cohort
+    if (
+      participants.filter((participant) => participant.participant_id === participantId).length > 1
+    ) {
+      // extra validation for the edge case
+      if (!participantStatuses.find((postHireStatus) => postHireStatus.is_current === true)) {
+        return getPostHireStatusLabel(participantStatuses[participantStatuses.length - 1]);
+      }
```
taraepp, line 80: I'm finding the logic here a bit difficult to follow.
For the conditions, would prefer to see something like:
```
const numDuplicateParticipants = participants.filter...
if (numDuplicateParticipants > 0) {....
const currentPostHireStatuses = ...
if (!currentPostHireStatuses)...
```
where we have a meaningful variable names assigned so it's easier to read the business logic from the code.

server/services/cohorts.js
```diff
@@ -60,7 +60,7 @@ const getCohortParticipants = async (cohortId) =>
       },
     })
     .find({
-      'cohortParticipantsJoin.cohort_id': cohortId,
+      cohort_id: cohortId,
       'participantStatusJoin.current': true,
       'participantStatusJoin.status': ['hired', 'archived'],
```
taraepp, line 65: I'm not totally sure if this is the problem, but you may be able to rework the query to take advantage of the count function, which you can use to count unique ids (but is incompatible with .find()). The conditions in the `find` can usually be moved into the joins.

arranfw (commented):
I forgot to press submit!! sorry

alva-fresh:
closing this pull request as per discussion: https://teams.microsoft.com/l/message/19:IUaAmNEwTGmcqt3CjlNCrc_j7pTI3rB5QRpVrTFxNHQ1@thread.tacv2/1663281487060?tenantId=5b973f99-77df-4beb-b27d-aa0c70b8482c&groupId=3f22a178-b2af-4d1a-99cd-bc1781c11210&parentMessageId=1663194560543&teamName=HCAP&channelName=General&createdTime=1663281487060&allowXTenantAccess=false

PR #853, opened 2022-09-15
HCAP-1360 Error on unused vars

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1360

So we can catch these earlier and not rely on sonarcloud


taraepp (approved):
👍

PR #858, opened 2022-09-19
HCAP-1363 Separate joins for hired and archived status

arranfw:
- Oversight from the last fix was that joining on `hired` would omit `archived` participants, we run into issues when we combine archived and hired participants in the same join so I split them up into two joins
  - When combining hired and archived into `participantStatusJoin` we can't guarantee `siteId` will be populated because some archive statuses don't have `siteId`
  - By splitting them up, we ensure we get all hired and archived participants, and can also be much more confident `siteJoin` will populate by referring to the hired status

https://freshworks.atlassian.net/browse/HCAP-1363

### DIFF
--- a/server/services/cohorts.js
+++ b/server/services/cohorts.js
@@ -28,8 +28,8 @@ const getCohort = async (id) =>
     id,
   });
 
-const getCohortParticipants = async (cohortId) =>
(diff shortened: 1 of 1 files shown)


taraepp (dismissed):
👍

alva-fresh (approved):

PR #859, opened 2022-09-19
HCAP-1364 Cohort participant count

arranfw:
Only counting current participants should stop us counting duplicate assignees to a single cohort

https://freshworks.atlassian.net/browse/HCAP-1364


taraepp (approved):
👍

PR #860, opened 2022-09-19
HCAP-1364 Dedupe cohort participants

arranfw:
https://freshworks.atlassian.net/browse/HCAP-1364

I forgot about reassigning, this is similar to a part of a fix @alva-fresh made before, but calculated on the server to reduce FE changes


alva-fresh (approved):

taraepp (approved):
👍

PR #868, opened 2023-01-26
HCAP-1379: Update vulnerable packages that don't have breaking changes

rhattie-ey:
[HCAP-1379](https://eydscanada.atlassian.net/browse/HCAP-1379)

This PR updates a number of dependencies project-wide. No breaking changes should be introduced. These updates address:

* Every vulnerable server dependency detected by `npm audit`
* Every vulnerable client dependency not required by the current version of `react-scripts`
* `moment` in the project-wide scope


vanics (dismissed):
It looks like that for the client dependencies only the package-lock.json was updated. Did we force update some of the vulnerabilities or it was more underlying dependencies? Anyway... I don't see a problem with it. It was more for my understanding.

rhattie-ey:
> It looks like that for the client dependencies only the package-lock.json was updated. Did we force update some of the vulnerabilities or it was more underlying dependencies?

It looks like a preinstall script was partially responsible for the mess in lock, which I just addressed (see latest commit), but also yes, on the client all I touched was sub-dependencies.

The big culprit on the client is `react-scripts` - the problem is, updating that means updating Webpack and making some other fixes to the build process. I propose we save that for another ticket and PR.

server/package.json
```diff
@@ -28,11 +28,11 @@
   },
   "dependencies": {
     "axios": "0.21.4",
-    "body-parser": "1.19.0",
+    "body-parser": "^1.20.1",
```
fw-noel, line 31: I recommend pinning these to a specific version so that we're not having to deal with any potential incompatible updates

rhattie-ey: True. I'll fix that when I get a moment.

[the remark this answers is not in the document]
vanics: Ops, my bad... I left my peer review open and didn't notice Noel commented on the same thing afterwards. Anyway. Good work!

fw-noel (approved):

vanics (approved):

PR #873, opened 2023-02-07
HCAP-1284 - Updated Site table with Allocations/Phases

mmumby:
Screen Shot 2023-02-06 at 3 17 35 PM

- Updated Sites table on UI to include a custom column component for Allocations, with phase start and end date
- BE resource updated to pull allocation from site_phase_allocation, and startDate, and endDate from the current phase
- Current phase is determined if today's date is between startDate and endDate. 
- The table already has FE filtering that orders the allocations N/A -> 0 -> numbers and can sort ascending/descending. 
- If there is no allocation associated with the site/phase 'N/A' will display.

- Removed 'allocations' field input from sites create and edit forms (removed all related validations for allocations on the employer_sites schema) (Allocations used to be below HA)
Screen Shot 2023-02-07 at 12 45 31 PM


client/src/pages/private/SiteTable.js
```diff
@@ -24,15 +25,37 @@ import { FeatureFlaggedComponent, flagKeys } from '../../services';
 import { fetchRegionSiteRows, fetchSiteRows } from '../../services/site';
 import { useTableStyles } from '../../components/tables/DataTable';
 
+const renderAllocations = (row) => {
+  const formattedDate = (date) => dayjs(row[date]).format('MMM D YYYY');
+  return (
+    <div>
```
fw-noel, line 31: Can we use empty < Box >es instead of empty divs, just for consistency

client/src/pages/private/SiteTable.js
```diff
@@ -24,15 +25,37 @@ import { FeatureFlaggedComponent, flagKeys } from '../../services';
...
+      {row['allocation'] || row['startDate'] ? (
+        <div>
+          <div style={{ fontWeight: '700' }}>{row['allocation']}</div>
+          <FeatureFlaggedComponent featureKey={flagKeys.FEATURE_PHASE_ALLOCATION}>
+            <div style={{ color: '#272833', fontWeight: '400' }}>
```
fw-noel, line 36: Can we get a `makeStyles` in this page instead of inline?

mmumby: I have this function outside the react class in order for the `columns` to reference it, and I cannot use `makeStyles` outside of the component, do you want me to move the styling to a regular object?

client/src/pages/private/SiteTable.js
mmumby: done!

client/src/pages/private/SiteTable.js
mmumby: created a regular style object vs inline. Let me know if this works for this use case!

client/src/pages/private/SiteTable.js
```diff
@@ -24,15 +25,47 @@ import { FeatureFlaggedComponent, flagKeys } from '../../services';
...
+  allocation: {
+    fontWeight: 700,
+  },
+  dates: {
+    color: '#272833',
+    fontWeight: 400,
+  },
+};
+
+const renderAllocations = (row) => {
```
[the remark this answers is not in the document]
mmumby: Good call - I am also of two minds here. When I know the component can be reused, I will always make a new component, but wasn't sure if it was necessary for this one since it seemed very specific to this table. However, the components are all pretty small, so I'll keep it that way!

fw-noel (approved):

client/src/pages/private/SiteTable.js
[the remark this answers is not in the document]
rhattie-ey: Sounds good.

rhattie-ey (approved):

PR #875, opened 2023-02-07
hcap 1384/hcap 1385 - phase form validations

mmumby:
- Add Placeholder text to input field (going off design in ticket for text)
- Add Max character validation to form (I used 255 as the db limit, but can decrease, can update error message to anything) 
Screen Shot 2023-02-07 at 2 22 51 PM
Screen Shot 2023-02-07 at 2 23 08 PM

- Added validation to ensure endDate is not before or on the same day as startDate (Can update error message)
Screen Shot 2023-02-07 at 2 43 55 PM

### DIFF
--- a/client/src/components/modal-forms/PhaseDialog.js
+++ b/client/src/components/modal-forms/PhaseDialog.js
@@ -63,7 +63,12 @@ export const PhaseDialog = ({ onSubmit, onClose, open, content, isNew }) => {
         {({ submitForm }) => (
           <FormikForm>
             <Box>
-              <Field name='phaseName' component={RenderTextField} label='* Phase name' />
+              <Field
+                name='phaseName'
+                component={RenderTextField}
+                label='* Phase name'
+                placeholder='Type in here'
+              />
               <Box
                 display='flex'
                 justifyContent='space-between'
--- a/client/src/components/modal-forms/PhaseDialog.js
+++ b/client/src/components/modal-forms/PhaseDialog.js
@@ -67,7 +67,7 @@ export const PhaseDialog = ({ onSubmit, onClose, open, content, isNew }) => {
                 name='phaseName'
                 component={RenderTextField}
                 label='* Phase name'
-                placeholder='Type in here'
+                placeholder='Enter a phase name'
               />
               <Box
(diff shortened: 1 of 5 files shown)


client/src/constants/validation/schema/schema-create-phase.js
```diff
@@ -2,14 +2,22 @@ import * as yup from 'yup';
...
+      if (startDate) {
+        const nextDay = new Date(startDate.getTime() + 86400000);
+        return schema.min(nextDay, 'End date must be at least 1 day after Start date');
+      } else {
+        return schema;
+      }
+    })
     .required(errorMessage)
     .typeError(errorMessage)
     .test('is-reasonable', errorDateIsReasonable, validateDateIsReasonable),
```
mmumby, line 23: `.min(yup.ref('startDate'), "End date can't be before start date")` could have been used instead of the when statement, but it didn't prevent the same day from being selected.

fw-noel (commented):
@mmumby I would love to see backend validation in this PR as well, `server/validators/phase.js`

mmumby:
@fw-noel I added the BE validations - note the BE is using the dates as string vs the FE which is using date, so the validations are different but still function the same.

client/src/components/modal-forms/PhaseDialog.js
```diff
@@ -63,7 +63,12 @@ export const PhaseDialog = ({ onSubmit, onClose, open, content, isNew }) => {
         {({ submitForm }) => (
           <FormikForm>
             <Box>
-              <Field name='phaseName' component={RenderTextField} label='* Phase name' />
+              <Field
+                name='phaseName'
+                component={RenderTextField}
+                label='* Phase name'
+                placeholder='Type in here'
```
rhattie-ey, line 70: Nitpick: could be more specifically relevant to the field's purpose. Something like "Enter a phase name", perhaps?

mmumby: Good suggestion, updated!

fw-noel (approved):

rhattie-ey (approved):
Looks good to me! Approved.

PR #878, opened 2023-02-10
Update to FK reference on site_phase_allocation to employer_site

mmumby:
The `getAllSitePhases` was referencing the incorrect FK id between allocations and sites, updated to use the `id` column, not the form input column `siteId`

id = 8
Updated UI
Screen Shot 2023-02-10 at 4 49 51 PM
site_phase_allocation in the db matching site_id = 8 
Screen Shot 2023-02-10 at 4 50 33 PM

### DIFF
--- a/server/services/phase.js
+++ b/server/services/phase.js
@@ -82,7 +82,7 @@ const getAllSitePhases = async (siteId) => {
       phase.id, 
       spa.id
     `,
-    [site.siteId]
+    [site.id]
   );
 
   // Transform data format and return it


rhattie-ey (approved):
Looks good to me. At some point we should really unify all these IDs to prevent these sorts of things...

mmumby:
@rhattie-ey Agreed - It's always confusing when the business case Id for the table and the PK for the record are basically the same name. I like Guids as the PK for this reason

fw-noel (approved):

PR #879, opened 2023-02-14
Feature/hcap 1287 - Set/Edit allocations

mmumby:
# work completed:
## DB
- dropped `start_date` and `end_date` from `site_phase_allocation`

## FE
- added form to set/edit Allocations
- Populated start/end date on form with phase dates - not editable.
- Added Validation (Must be a positive number, required)

## BE
- added POST and PATCH
- added BE validation 
- returned `allocation_id` to the phase payload to allow editing on FE
- removed references to allocation `start_date` and `end_date`

## Testing
- Cypress tests for happy path, and UI validations
- Server and E2E tests testing happy path, unauthorized, data validation for POST and PATCH

## UI
Screen Shot 2023-02-14 at 12 32 16 PM
Screen Shot 2023-02-14 at 12 32 56 PM

### DIFF
--- a/client/src/services/allocations.js
+++ b/client/src/services/allocations.js
@@ -1,0 +1,28 @@
+import store from 'store';
+import { API_URL } from '../constants';
+
+export const createPhaseAllocation = async (allocationJson) => {
+  const response = await fetch(`${API_URL}/api/v1/allocation`, {
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${store.get('TOKEN')}`,
+      Accept: 'application/json',
+      'Content-type': 'application/json',
+    },
+    body: JSON.stringify(allocationJson),
+  });
+  return response;
+};
+
+export const updatePhaseAllocation = async (allocationId, allocationJson) => {
+  const response = await fetch(`${API_URL}/api/v1/allocation/${allocationId}`, {
+    method: 'PATCH',
+    headers: {
+      Authorization: `Bearer ${store.get('TOKEN')}`,
+      Accept: 'application/json',
+      'Content-type': 'application/json',
+    },
+    body: JSON.stringify(allocationJson),
+  });
+  return response;
+};
--- a/client/src/services/allocations.js
+++ b/client/src/services/allocations.js
@@ -16,7 +16,7 @@ export const createPhaseAllocation = async (allocationJson) => {
 
 export const updatePhaseAllocation = async (allocationId, allocationJson) => {
   const response = await fetch(`${API_URL}/api/v1/allocation/${allocationId}`, {
-    method: 'PUT',
+    method: 'PATCH',
     headers: {
       Authorization: `Bearer ${store.get('TOKEN')}`,
       Accept: 'application/json',
--- a/client/src/services/allocations.js
+++ b/client/src/services/allocations.js
@@ -1,7 +1,7 @@
 import store from 'store';
 import { API_URL } from '../constants';
 
-export const createPhaseAllocation = async (allocationJson) => {
+export const createPhaseAllocation = async (payload) => {
   const response = await fetch(`${API_URL}/api/v1/allocation`, {
     method: 'POST',
     headers: {
@@ -9,12 +9,12 @@ export const createPhaseAllocation = async (allocationJson) => {
       Accept: 'application/json',
       'Content-type': 'application/json',
     },
-    body: JSON.stringify(allocationJson),
+    body: JSON.stringify(payload),
   });
   return response;
 };
 
-export const updatePhaseAllocation = async (allocationId, allocationJson) => {
+export const updatePhaseAllocation = async (allocationId, payload) => {
   const response = await fetch(`${API_URL}/api/v1/allocation/${allocationId}`, {
     method: 'PATCH',
     headers: {
@@ -22,7 +22,7 @@ export const updatePhaseAllocation = async (allocationId, allocationJson) => {
       Accept: 'application/json',
       'Content-type': 'application/json',
     },
-    body: JSON.stringify(allocationJson),
+    body: JSON.stringify(payload),
   });
   return response;
 };
--- a/client/src/services/allocations.js
+++ b/client/src/services/allocations.js
@@ -1,7 +1,7 @@
 import store from 'store';
 import { API_URL } from '../constants';
 
-export const createPhaseAllocation = async (payload) => {
+export const createAllocation = async (payload) => {
   const response = await fetch(`${API_URL}/api/v1/allocation`, {
     method: 'POST',
     headers: {
@@ -14,7 +14,7 @@ export const createPhaseAllocation = async (payload) => {
   return response;
 };
 
-export const updatePhaseAllocation = async (allocationId, payload) => {
+export const updateAllocation = async (allocationId, payload) => {
   const response = await fetch(`${API_URL}/api/v1/allocation/${allocationId}`, {
     method: 'PATCH',
     headers: {
--- a/client/src/services/allocations.js
+++ b/client/src/services/allocations.js
@@ -16,7 +16,7 @@ export const createPhaseAllocation = async (allocationJson) => {
 
 export const updatePhaseAllocation = async (allocationId, allocationJson) => {
   const response = await fetch(`${API_URL}/api/v1/allocation/${allocationId}`, {
-    method: 'PATCH',
+    method: 'PUT',
     headers: {
       Authorization: `Bearer ${store.get('TOKEN')}`,
       Accept: 'application/json',
--- a/server/migrations/1676328102966_drop-allocation-dates.js
+++ b/server/migrations/1676328102966_drop-allocation-dates.js
@@ -1,0 +1,27 @@
+/* eslint-disable camelcase */
+const { collections } = require('../db');
+
+exports.shorthands = 'drop-allocation-dates';
+
+exports.up = (pgm) => {
+  pgm.dropColumns(collections.SITE_PHASE_ALLOCATION, ['start_date', 'end_date'], {
+    ifExists: true,
+  });
+};
+
+exports.down = (pgm) => {
+  pgm.addColumns(
+    collections.SITE_PHASE_ALLOCATION,
+    {
+      start_date: {
+        type: 'date',
+        notNull: false,
+      },
+      end_date: {
+        type: 'date',
+        notNull: false,
+      },
+    },
+    { ifNotExists: true }
+  );
+};
--- a/server/routes/allocation.js
+++ b/server/routes/allocation.js
@@ -4,7 +4,11 @@ const logger = require('../logger.js');
 const { asyncMiddleware } = require('../error-handler.js');
 const { CreateAllocationSchema } = require('../validation');
 const { expressRequestBodyValidator } = require('../middleware');
-const { createPhaseAllocation, updatePhaseAllocation } = require('../services/allocations');
+const {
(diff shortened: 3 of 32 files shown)


server/routes/allocation.js
```diff
@@ -0,0 +1,85 @@
...
+  ],
+  asyncMiddleware(async (req, resp) => {
+    if (!FEATURE_PHASE_ALLOCATION) {
+      return resp.status(501).send('Phase allocation feature not active');
+    }
+    try {
+      const { body, hcapUserInfo: user } = req;
+      const response = await updatePhaseAllocation(req.params.id, body, user);
+      logger.info({
+        action: 'phase-allocation_post',
```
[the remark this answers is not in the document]
mmumby: I was following the pattern for creating global phases (which I guess isn't fully implemented yet) but it was using a patch. I can make it a put!

server/services/allocations.js
```diff
@@ -0,0 +1,42 @@
...
+ * @property {number} site_id       FK for employer_site
+ */
+
+const getPhaseAllocation = async (siteId, phaseId) => {
+  const allocation = await dbClient.db[collections.SITE_PHASE_ALLOCATION].findOne(
+    {
+      site_id: siteId,
+      phase_id: phaseId,
+    },
+    { limit: 1 }
```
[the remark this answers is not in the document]
mmumby: no - forgot to remove it after changing it from `find` to `findOne`, I'll update it!

server/routes/allocation.js
```diff
@@ -0,0 +1,85 @@
...
+  })
+);
+
+// Update phase allocation: PATCH
+router.patch(
+  '/:id',
+  [
+    keycloak.allowRolesMiddleware('ministry_of_health'),
+    keycloak.getUserInfoMiddleware(),
+    expressRequestBodyValidator(CreateAllocationSchema),
```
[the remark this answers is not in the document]
mmumby: phase_id/site_id are included in the BE validations (not the FE though) 

```
phase_id: yup.number().required('Allocation must be related to a phase'),
site_id: yup.number().required('Allocation must be related to a site'),
```

server/routes/allocation.js
```diff
@@ -0,0 +1,85 @@
...
+          username: user.username,
+          id: user.id,
+        },
+        allocationId: response.id,
+      });
+      logger.info(response);
+      return resp.status(201).json(response);
+    } catch (err) {
+      logger.error(err);
+      return resp.status(400).send(`${err}`);
```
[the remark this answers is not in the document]
mmumby: Updated to be a specific message,  but appears to show "Bad Request" in the toast message regardless of what is sent. I will have to investigate further.

server/routes/allocation.js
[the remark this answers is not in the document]
mmumby: Just realized this is not what you were referring to... updated the sting to match the request... but should it be a PATCH or PUT?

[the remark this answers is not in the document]
fw-noel: Patch is good! I just meant the logger info :)

server/routes/allocation.js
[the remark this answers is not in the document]
fw-noel: Yeah it's less what it shows right now but more what it could potentially show to the user, whether its in an error message or just the network tab. If there's a possibility that it returns something like
"Error on UPDATE table_name SET" etc. etc., not wonderful.

Updating to a specific message resolves this comment :)

server/routes/allocation.js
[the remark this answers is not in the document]
mmumby: switched it back to PATCH!

server/routes/allocation.js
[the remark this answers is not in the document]
fw-noel: Sorry yes, I was wondering if using the same Schema for the EDIT endpoint works okay since phase_id and site_id is expected, but I don't think we want to / need to be passing those in?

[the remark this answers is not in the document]
mmumby: ohh good point, I was passing those in but it's not necessary. I'll update the patch/validation for that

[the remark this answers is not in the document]
mmumby: updated!

server/migrations/1676328102966_drop-allocation-dates.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable camelcase */
+const { collections } = require('../db');
+
+exports.shorthands = 'drop-allocation-dates';
+
+exports.up = (pgm) => {
+  pgm.dropColumns(collections.SITE_PHASE_ALLOCATION, ['start_date', 'end_date'], {
```
rhattie-ey, line 7: I appreciate this migration, makes things tidier.

server/services/allocations.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable camelcase */
+// disabling camelcase check so that we can manipulate snake_case attributes without changing structure
+const { dbClient, collections } = require('../db');
+
+/**
+ *
```
rhattie-ey, line 6: I like that there's a type being defined, but I don't see any references being made to it yet. Is that going to be done in another commit? I'm also wondering if there's a clearer, more unique type name than `allocation` to give it, but that's a nitpick.

server/routes/allocation.js
```diff
@@ -0,0 +1,85 @@
+const express = require('express');
```
rhattie-ey, line 1: Having files called both `allocation` and `phase-allocation` is confusing, as are the endpoints of the same name. I'd propose coming up with a better distinction between the two - perhaps `phase-allocation` is the actual allocation stuff (as you are doing here), and the global stuff becomes `global-phases` or something? I'm open to ideas on how to rename these.

client/src/services/allocations.js
```diff
@@ -0,0 +1,28 @@
+import store from 'store';
+import { API_URL } from '../constants';
+
+export const createPhaseAllocation = async (allocationJson) => {
```
rhattie-ey, line 4: I know it's a convention established in other files, but the fact that a parameter called `allocationJson` isn't actually JSON bothers me. I'd suggest renaming it to something like `allocationData`, to better hint at its type.

rhattie-ey (commented):
Looks good so far! No major complaints from me.

server/tests/util/keycloak.js
```diff
@@ -21,6 +22,11 @@ const healthAuthority = {
   password: process.env.KC_TEST_HA_PWD || 'password',
 };
 
+const ministryOfHealth = {
+  username: 'test-moh',
+  password: 'password',
```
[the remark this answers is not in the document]
rhattie-ey: Resolved in latest batch of commits.

server/services/phase.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
mmumby: I meant to remove this -  I think I wanted to push to GitHub to test a theory/bypass linting and forgot to remove it. Thanks for catching it, removed it now!

server/routes/allocation.js
mmumby: This is exactly what I wanted feedback on - it was getting confusing for me too. I originally started naming everything 'PhaseAllocations' to match the db table/FK relationship (without using SitePhaseAllocation because that seemed too long). But eventually, I simplified everything to just 'allocation' as there appeared to be some overlap with global phases (Which has a few different names in the codebase -ie  phase/phases, sitePhases, globalPhases)

I can rename a few things - I kind of think allocations and phases can stand alone without any extra context in their names. 
ie - createAllocation, updateAllocation, getPhase, createPhase, updatePhase. Thoughts?

client/src/services/allocations.js
mmumby: renamed to payload!

server/services/allocations.js
mmumby: I think perhaps I am confused, I was using this as a doc block to define what is being returned from `getPhaseAllocation`, What am I missing?

server/services/phase.js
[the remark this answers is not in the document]
rhattie-ey: Just a heads up, you also left the same directives in `allocations.service.test` and `allocation.e2e.test`, as of the current commit.

server/routes/allocation.js
fw-noel: I think `phases` vs. `allocations` would be the way to go. I wouldn't think global phase, because there's a chance it won't be fully "global" later on.

server/services/allocations.js
rhattie-ey: `@typedef` is a JSDoc tag which *defines* a type. It does not mark any given symbols as that type. For that, you need `@type`. Your JSDoc block is equivalent to the following TypeScript:

```typescript
type allocation {
  id: number;
  phase_id: string;
  allocation: number;
  site_id: number;
}
```

In other words, you're making a new *type* at the file scope, which happens to share a name with a local *variable*, but has nothing to do with it. We can check this by hovering over the `allocation` variable:
image

This can be fixed by tagging the variable with `@type {typename}`, like so:
image

And now, VSCode and other IDEs can provide information as you edit, and TS can check typing properly:
image
image

client/src/services/allocations.js
rhattie-ey: Much better!

server/routes/allocation.js
mmumby: If this is the agreed path forward, I can rename all the allocation stuff in this PR, and create a separate PR/ticket to rename phases.

server/services/allocations.js
mmumby: ah - okay thank you for the detailed explanation, I think I have it working the correct way now.

server/services/phase.js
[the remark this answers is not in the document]
mmumby: removed!

server/routes/allocation.js
rhattie-ey: That works for me!

server/services/allocations.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+const getAllocation = async (siteId, phaseId) => {
+  /**
+   * @typedef {Object} allocationResponse  type for DB response to the `getAllocation` query
+   * @property {number} id                 PK ID of the allocation
+   * @property {number} phase_id           FK for phase
+   * @property {number} allocation         Number of allocations available
+   * @property {number} site_id            FK for employer_site
+   *
+   *
```
rhattie-ey, line 18: Nitpick: these extra lines can be removed.

rhattie-ey (dismissed):
Looks good to me!

server/validators/allocation.js
```diff
@@ -0,0 +1,36 @@
...
+const { validateBlankOrPositiveInteger } = require('./helpers');
+
+const CreateAllocationSchema = yup
+  .object()
+  .noUnknown('Unknown field in entry')
+  .shape({
+    allocation: yup
+      .number()
+      .nullable()
+      .max(99, 'Must be between 0-99')
```
fw-noel, line 11: Ah shoot sorry I meant to say, no max limits!
Please remove the related test too!

mmumby: removed!

server/services/allocations.js
mmumby: removed!

fw-noel (approved):

rhattie-ey (approved):
Looks good, approved!

vanics (approved):

PR #887, opened 2023-02-17
testing: new make command for testing in clean db

fw-noel:
### Note
You will need a new variable in your local .env file, `NODE_ENV=local`

It's still not 100% ideal, but I do think it helps to have that separation between automated test db and dev db. This will ensure that you have a clean slate to work with everytime you run the local testing setup command


docs/development.md
```diff
@@ -49,23 +49,21 @@ Go to `http://keycloak.local.freshworks.club:8080/auth` console and add users/as
...
-  - `make local-server-workspace`
-  - `npm test`
-- or by running
-  - `make local-server-tests`
-
-### To run e2e tests:
-- Make sure containers are running
-  - `make local-kc-run`
+- Whether or not you're currently running the app, run the test setup. This will ensure your server is pointed to the `db_test` database, clearing and repopulating it.
+  - `make local-test-setup`
```
mmumby, line 55: 2 things I noticed while testing: 
1) the makefile command is defined as: `local-testing-setup` 
2) `make kc-import-users` needs to be run before 'make local-testing-setup' or there appears to be an issue regarding the participant's table and the setup fails. 

It might also be helpful to mention in the comment to set up local keycloak before running any of these commands, as that might not be obvious to new devs.

mmumby:
Seems to work well, just one comment regarding the issues I had while getting setup. Happy to approve once that's updated. 

Can you also add `NODE_ENV=local` to the .env.example file.

server/scripts/clear-data.js
```diff
@@ -0,0 +1,19 @@
...
+  if (require.main === module) {
+    try {
+      await dbClient.connect();
+      await dbClient.runRawQuery(
+        `truncate ${[Object.values(databaseCollections).join(',')]} RESTART IDENTITY;`
+      );
+      await dbClient.disconnect();
+      process.exit(0);
+    } catch (err) {
+      console.log('Failed to truncate tables', err);
```
rhattie-ey, line 15: You'll want `/* eslint-disable no-console */` at the top of this file to allow use of `console`. Otherwise, ESLint will throw warnings, and Huskey might even prevent commits from working.

server/db/schema.js
```diff
@@ -2,12 +2,11 @@
  * Define db schema rules including collection names and indexes
  */
 
-const collections = {
+const databaseCollections = {
```
rhattie-ey, line 5: Might be a good idea to add comments here explaining what each of these are - it's not immediately apparent from the names.

server/scripts/clear-data.js
vanics: Also, shouldn't this be a console.error?

rhattie-ey: Agreed, or at least warn.

rhattie-ey (commented):
Looks pretty much ready to merge!

server/scripts/clear-data.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable no-console */
+require('dotenv').config({ path: '../.env' });
+const { dbClient } = require('../db');
+const { databaseCollections } = require('../db/schema');
```
rhattie-ey, line 4: `Require`s should be replaced with imports, and calls to `dotenv` should be replaced with `import './load-env';` at the top of the file.

fw-noel: Oh yes, my bad, didn't handle the merge

server/scripts/clear-data.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-console */
+import './load-env';
+require('dotenv').config({ path: '../.env' });
```
rhattie-ey, line 3: Why is this still here?

fw-noel: I don't know!!! I swear I removed it

mmumby (approved):

rhattie-ey (approved):

PR #892, opened 2023-02-23
HCAP-1424: Add types to server validators directory

rhattie-ey:
Closes:
* [HCAP-1424](https://eydscanada.atlassian.net/browse/HCAP-1424) (converting validation to TS)
* [HCAP-1425](https://eydscanada.atlassian.net/browse/HCAP-1425) (updating `yup` to 1.0.0)

This PR:
* Updates `yup` to 1.0.0 on the server
* Updates validation code to work with latest version of `yup`
* Converts all validation code to TS

Not much actual typing is going on here - the only in-scope typing besides helpers is exports, and most of that is taken care of by updating `yup`. Helpers cannot expect a specific type, in most cases, since their job is to help verify data structure. At some point more idiomatic type checking could be added to them, though.


server/validators/phase.ts
```diff
@@ -1,16 +1,16 @@
...
   start_date: yup
     .string()
     .required('Start date is required')
     .test('is-date', 'Not a valid date', validISODateString),
   end_date: yup
     .string()
-    .when('start_date', (startDate, schema) => {
+    // See https://github.com/jquense/yup/issues/1901 for why this is needed
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .when('start_date', (startDate: any, schema) => {
```
fw-noel, line 13: Does this need to be `any`?

fw-noel (approved):

server/validators/phase.ts
rhattie-ey: Yes. There seems to be a bug in `yup`'s typing - forcefully making this `any` prevents TS from complaining about stuff that isn't actually an issue. Hopefully once `yup` gets patched, we can pull all of this out.

mmumby (approved):
Looks good, nice work!

PR #895, opened 2023-02-24
HCAP-1401: Add types to server services directory

rhattie-ey:
This PR:
* Converts all server services files to TypeScript
* Adds stronger typing across much of the services directory
* Converts a couple constant objects into enums
* Refactors the following files into multiple smaller ones:
    * `participants-helper`
    * `participants`
    * `reporting`
    * `participant-status`
        * `setParticipantStatus` was further refactored into multiple smaller functions for readability
* Fixes an ESLint bug with enums

This PR does *not*: 
* Achieve 100% strong typing across `services`. It's a huge directory, and certain parts of it are very time consuming to untangle. Some parts will just need to go un-annotated for now. Everyone is encouraged to continue tightening up types when possible.
* Fully refactor `services`. Again, huge directory, and that's out of scope for this ticket. This PR does work in a few refactors, though.


server/constants/validation-constants.ts
```diff
@@ -74,12 +74,14 @@ export const participantStatuses = [
...
 
-export const postHireStatuses = {
-  orientationCompleted: 'orientation_completed',
-  postSecondaryEducationUnderway: 'post_secondary_education_underway',
-  postSecondaryEducationCompleted: 'post_secondary_education_completed',
-  cohortUnsuccessful: 'cohort_unsuccessful',
-};
+// Not sure why eslint thinks this is a shadow - it isn't.
+// eslint-disable-next-line no-shadow
+export enum postHireStatuses {
```
fw-noel, line 79: Can you explain why const -> enums makes sense for these changes?

rhattie-ey: Two reasons:
1. It's more idiomatic. The entire point of an enum is to store a set of named constants, and that's exactly what we're doing here.
2. Stronger type checking. When using enums, we can say "this MUST be one of these values", and TS will do its best to enforce that. Each member is also treated as a unique type of sorts, which can be handy. The main benefit here is enforcing strict values, though - code that takes an enum mandates that invalid values doesn't reach it.

fw-noel: Good reasons :)

vanics: Love this!

vanics:
I love your attention to details and your dilligence improving the code base. Well done! It looks great.
The change was considerable, I skimmed here but will open locally to explore more but so far what I've seem is really god.

server/services/return-of-service.ts
```diff
@@ -4,7 +4,28 @@ import { rosError } from '../constants';
 import { getParticipantHiredStatuses } from './participant-status';
 import logger from '../logger';
 
-const getRosParticipantStatus = async (participantId) => {
+interface RosParticipant {
```
fw-noel, line 7: Hmm my concern with at least this naming is that RosParticipant might consist of more than these properties, even if this is all we need for validation. Will be checking on this

server/services/reporting/util.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Validation layer to see if the user has access to requested health region
+ * @param user data of a user who accesses the endpoints
+ * @param regionId health region
+ * @return true if the user has access
+ */
+export const checkUserRegion = (user, regionId: string): boolean =>
+  user && user.regions?.includes(regionId);
+
+export const mapRosEntries = (rosEntries: RosEntry[]) =>
```
fw-noel, line 36: This (and the interface) feels out of place in a utils file. Feeling it more in a ROS specific file or your existing milestones one

fw-noel (commented):
Am just over halfway done, tomorrow I really want to review the participant status split and run the branch locally

server/services/participant-status/set-status.ts
```diff
@@ -0,0 +1,177 @@
...
+      employerId,
+      data,
+      site,
+      isRosComplete,
+      participant,
+      existingCurrentStatus,
+      currentStatus,
+      currentStatusId,
+    });
+    if ('result' in transitionResult) return transitionResult.result;
```
fw-noel, line 128: This should probably be explained with a comment

server/services/participants/participants-multi.ts
```diff
@@ -0,0 +1,288 @@
+import readXlsxFile from 'node-xlsx';
```
fw-noel, line 1: I don't think I love the `participant-single` and `participants-multi` distinction. I think separating out the batch update stuff may be the way to go, and `getParticipants` can be lumped in with the stuff in `-single`.

fw-noel (commented):
Still running into the validation issue when running it locally. The refactor itself looks okay, will be wanting to thoroughly test it

server/services/user-validation.ts
```diff
@@ -1,7 +1,14 @@
 import { getParticipantByID } from './participants';
 import logger from '../logger';
 
-const validateParticipant = async (participantId, actionType) => {
+interface User {
```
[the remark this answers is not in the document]
rhattie-ey: I'd say keep it here for now until we can figure out how many other files use this same structure of user object. Perhaps I should rename it to something like `ValidationUser` for specificity's sake though?

server/services/sync-user.ts
```diff
@@ -2,11 +2,13 @@ import { dbClient, collections } from '../db';
 import keycloak from '../keycloak';
 import logger from '../logger';
 
-export const syncUser = async ({ log }) => {
+export const syncUser = async ({ log }: { log: boolean }) => {
```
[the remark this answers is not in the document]
rhattie-ey: That's not a return type, that's an argument type. Note the parenthesis. The return type is currently `Promise<any[]>`, since the return type of `updateDoc` on `USERS` is not defined.

server/constants/participant-status.ts
```diff
@@ -1,17 +1,17 @@
...
-  PENDING_ACKNOWLEDGEMENT: 'pending_acknowledgement',
-  ALREADY_HIRED: 'already_hired',
-  INVALID_STATUS_TRANSITION: 'invalid_status_transition',
-  INVALID_ARCHIVE: 'invalid_archive',
-  INVALID_STATUS: 'invalid_status',
-  UNAVAILABLE: 'unavailable',
-  ROS: 'ros',
-  REJECT_ACKNOWLEDGEMENT: 'reject_ack',
-};
+export enum participantStatus {
```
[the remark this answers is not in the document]
rhattie-ey: Fixed! That was a relic from before it was an enum.

server/services/participant-status/set-status.ts
rhattie-ey: Added a couple comments at this section.

server/services/participants/participants-multi.ts
rhattie-ey: Good idea, changed as suggested.

server/services/sync-user.ts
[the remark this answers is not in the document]
fw-noel: Definitely just misread that lol, thanks

server/services/allocations.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable camelcase */
+import { HcapUserInfo } from '../keycloak';
```
mmumby, line 2: Is it worth disabling the camelcase rule across the entire server? We seem to be disabling this rule in several files.

server/services/participant-status/get-statuses.ts
```diff
@@ -0,0 +1,13 @@
+import { dbClient, collections } from '../../db';
+
+export const getParticipantHiredStatuses = async (participantId: number) => {
+  const statuses = await dbClient.db[collections.PARTICIPANTS_STATUS].find({
+    participant_id: participantId,
+    status: 'hired',
+    current: true,
```
mmumby, line 7: Should this be updated to use the ParticipantStatus const? There are a few places where the status is typed, is that an intentional decision, or were they missed? (If its intentional I could be misunderstanding the purpose of an enums const)

server/services/user-validation.ts
```diff
@@ -1,7 +1,14 @@
...
+  email?: string;
+  // eslint-disable-next-line camelcase
+  user_id: string;
+  sub: string;
```
mmumby, line 8: In this situation - user_id is referencing the `user.body.userInfo.id` not `user.id`, therefore it's a string not a number?

mmumby: This isn't an issue with your pr, I'm just noticing the inconsistencies between IDs being numbers vs strings (Which makes sense when we're referencing the body of the data and not the PK)

mmumby (commented):
Nice work, Raelynn. Few comments, but seems to be a great improvement overall.

server/services/allocations.ts
rhattie-ey: I think I prefer how it is at the moment. If anything, I'd argue that we should be working to refactor out as much camelcase as possible. One potential solution there would be a services-wide utility function that converts snake case to camel case (and an inverse function), which we'd then use on all DB operations to ensure all JS project-wide is seeing nothing but camels. A discussion for another time, though.

server/services/participant-status/get-statuses.ts
rhattie-ey: That's just a result of how many literals were scattered around the codebase, but good catch for a refactor opportunity. I went through and swapped a lot of that out.

server/services/user-validation.ts
rhattie-ey: Yeah, IDs are... messy. Not a fan of how they're handled right now. Potential refactor opportunity, but comes with a potential need for DB work.

mmumby: yeah a large refactor indeed.

mmumby (dismissed):

server/services/cohorts.ts
```diff
@@ -4,7 +4,7 @@ import utc from 'dayjs/plugin/utc';
 import { dbClient, collections } from '../db';
 import { validate, CreateCohortSchema } from '../validation';
 import { getPostHireStatusesForParticipant } from './post-hire-flow';
-import { postHireStatuses } from '../constants';
+import { postHireStatuses, ParticipantStatus as ps } from '../constants';
```
fw-noel, line 7: Is this just to avoid confusing things with the Massive queries?

fw-noel (approved):

server/services/cohorts.ts
rhattie-ey: It's just to make sure we're using our constants as much as possible, instead of leaving the string equivalent of magic numbers everywhere.

PR #896, opened 2023-02-24
HCAP-1422 - Phase/Allocations do not get populated after Site is edited

mmumby:
## Work completed:
- Initially, when the details page mounted - The parent component `SiteViewDetails`, and the child 'SiteViewDetaisTabs' were each fetching the site details. The child was handling all logic for participants/phases, but the parent was handling logic for editing sites. (Thus why phases/allocations were not being re-fetched after successful edit)
Screen Shot 2023-02-24 at 9 42 41 AM
Screen Shot 2023-02-24 at 9 46 11 AM

- Reduced network calls on initial load to 3 - getSite, getParticipants, getPhases, and moved that logic to the parent. 
- Moved API calls to service files. (There is inconsistency in how we handle the response, but I did not address those inconsistencies in this PR)
- Fixed bug - if you edit a site, the phase data is still visible.
- Manually tested al actions to ensure they are still working as intended. 

## Additional work
- removed an additional button click on cypress test, it was already being clicked in `setAllocationForm()`, also updated spelling mistake on fn name.


client/src/pages/private/SiteViewDetails.js
```diff
@@ -50,15 +55,24 @@ export default ({ match }) => {
...
+      const participants = await fetchSiteParticipants(columnIDs, site.siteId);
+      setIsLoading(false);
+      return setSite({
+        ...site,
+        hiredParticipants: participants.hiredParticipants,
+        withdrawnParticipants: participants.withdrawnParticipants,
+        phases,
+      });
+    } else {
+      return setSite({ hiredParticipants: [], withdrawnParticipants: [], phases: [] });
```
vanics, line 75: Is there not a need to report any error? A lil toast? :)

mmumby: good call - I've updated it to use the same toast/error message as the PATCH.

client/src/services/site.js
```diff
@@ -59,3 +66,65 @@ export const createSite = async (siteJson) => {
...
+
+export const fetchSiteParticipants = async (columnIDs, siteId) => {
+  const response = await fetch(`${API_URL}/api/v1/employer-sites/${siteId}/participants`, {
+    headers: { Authorization: `Bearer ${store.get('TOKEN')}` },
+    method: 'GET',
+  });
+
+  if (response.ok) {
+    const { hired, withdrawn } = await response.json();
+    const hiredParticipants = mapSiteParticipantsDataToRow(hired, columnIDs);
```
fw-noel, line 91: I'd suggest separating out this logic (separating hired and withdrawn) from the API request itself. Single responsibility principle.

Keep fetchSiteParitcipants to just grabbing the data eand returning it.

fw-noel (commented):
This seems to work, running it locally! But I think I would like to see my suggested change implemented or debated :)

client/src/services/site.js
rhattie-ey: I think that's reasonable, although I don't have any major issues with the way it is right now either.

rhattie-ey:
Looks good so far. I'll give it a proper review on Monday once I have a chance to try it out myself.

client/src/pages/private/SiteViewDetails.js
```diff
@@ -50,15 +55,28 @@ export default ({ match }) => {
...
+        phases,
+      });
+    } else {
+      openToast({
+        status: ToastStatus.Error,
+        message: response.error || response.statusText || 'Server error',
+      });
+      return setSite({ hiredParticipants: [], withdrawnParticipants: [], phases: [] });
     }
   }, [id, setSite]);
```
rhattie-ey, line 81: Pretty sure `openToast` should be in this array.

rhattie-ey (commented):
Looks good to me overall!

client/src/services/site.js
mmumby: Updated - Moved the mapping logic to the component. However, regarding inconsistencies, we are mapping the response for `fetchRegionSiteRows` and `fetchSiteRows` in the service file in a similar way before returning to the component.

client/src/pages/private/SiteViewDetails.js
mmumby: updated! Thanks for pointing that out

fw-noel (approved):

client/src/pages/private/SiteViewDetails.js
```diff
@@ -64,25 +97,29 @@ export default ({ match }) => {
...
+      setIsLoading(false);
+      return setSite({});
     }
-  }, [id, setSite]);
+  }, [id, setSite, openToast, setIsLoading]);
 
   useEffect(() => {
     fetchDetails();
-  }, [id, fetchDetails]);
+  }, []);
```
rhattie-ey, line 122: Another missing dependency here.

client/src/pages/private/SiteViewDetails.js
rhattie-ey: Great! It might be worth ensuring you have eslint integration enabled in your editor - it has caught details like this many times for me.

client/src/pages/private/SiteViewDetails.js
mmumby: I also added a check `if (site) fetchData()`, but due to setting the state multiple times in `fetchData`, the useEffect was being called multiple times, which I don't want. Removing the dependency resolved that - open to other solutions though!

rhattie-ey: The fact that it does that feels a bit iffy to me, structurally - I'll see if I can find a fix.

rhattie-ey: I have a potential solution, although it's more verbose.

1. Add a `hasFetched` state variable
2. In `fetchDetails`, add a `setHasFetched` call to show that we've tried to fetch at least once
3. Change the contents of lines 122-124 to:
```js
useEffect(() => {
    if (!hasFetched) fetchDetails();
  });
  ```

It's more verbose, but also makes it more obvious that we only want to try this once.

mmumby: Updated: Thanks for the suggestion, I did try it with the "isLoading" flag, but fetchData was updating it multiple times, and for some reason, I didn't think to create an additional flag just for this use case.

rhattie-ey: Looks good now!

rhattie-ey (approved):
Looks good! Approved.

fw-noel (approved):

PR #897, opened 2023-02-27
HCAP-1294 - Phase date range validation

mmumby:
WIP


client/src/components/modal-forms/PhaseDialog.js
```diff
@@ -15,18 +18,44 @@ const useStyles = makeStyles(() => ({
...
+  },
+  listItem: {
+    fontSize: '14px',
+    paddingTop: '0',
+    paddingBottom: '0',
+  },
 }));
 
-export const PhaseDialog = ({ onSubmit, onClose, open, content, isNew }) => {
+export const PhaseDialog = ({ onSubmit, onClose, open, content, isNew = false }) => {
```
fw-noel, line 32: I would've thought `isNew` would be defaulted to true, but I don't think it matters too much

client/src/components/modal-forms/PhaseDialog.js
```diff
@@ -15,18 +18,44 @@ const useStyles = makeStyles(() => ({
...
   const { openToast } = useToast();
   const classes = useStyles();
+  const [phases, setPhases] = useState([]);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    if (loading) {
```
fw-noel, line 39: I don't think the name `loading` is very clear here. Just sounding it out, it sounds like "if we're currently loading, then fetch data".

In DataTables, `isPendingRequests` is used I think. May want to follow

cypress/integration/phase.spec.js
```diff
@@ -84,15 +83,26 @@ describe('Phase functionality', () => {
...
+
+    cy.contains('p.Mui-error', 'Conflict with 1 or more phases');
+  });
+
   it('MoH can edit the start date and end date of a phase', () => {
     navigateToEditForm();
 
-    cy.get('[name=Startdate]').clear().type(`{ctrl+v}2024/01/05`);
-    cy.get('[name=Enddate]').clear().type(`{ctrl+v}2027/01/04`);
+    cy.get('[name=Startdate]').clear().type(`{ctrl+v}1990/06/06`);
```
fw-noel, line 104: Just realized, you should probably add a comment on why the `paste` is needed

client/src/components/modal-forms/PhaseDialog.js
mmumby: `PhaseDialog` is being referenced in two spots, currently, it is easier to add additional props/flags to the new reference as its being rendered in the JSX whereas the edit reference is being passed as an object to the new DataTable component, so would require some updates to DataTable. For that reason, I've defaulted the flag to false (It was reading as undefined prior)

client/src/components/modal-forms/PhaseDialog.js
mmumby: updated!

cypress/integration/phase.spec.js
mmumby: Added comments in all tests using this logic, also added a comment to another section where we're mocking a downArrow click.

cypress/integration/phase.spec.js
```diff
@@ -34,8 +35,8 @@ describe('Phase functionality', () => {
     navigateToForm();
     const formValues = {
       phaseName: 'Test phase',
-      startDate: '2021/03/30',
-      endDate: '2022/03/31',
+      startDate: '1990/01/01',
+      endDate: '1992/01/01',
```
rhattie-ey, line 39: What's the reason for this change?

server/tests/phase.e2e.test.js
```diff
@@ -26,21 +26,60 @@ describe('api e2e tests for /phase', () => {
...
+      .post(`/api/v1/phase`)
+      .send({
+        name: 'Test Phase name overlap Post',
+        start_date: '2009/12/31',
+        end_date: '2010/12/31',
+      })
+      .set(header);
+    expect(res.status).toEqual(400);
+  });
+
```
rhattie-ey, line 74: Nice work with these tests, should be pretty comprehensive.

rhattie-ey (commented):
Looks good! Once the other comments are addressed, I'd be ready to approve this.

cypress/integration/phase.spec.js
mmumby: As a fresh start, I went through and added all new dates to the phase tests to ensure there were no overlaps. It likely could have stayed as it was though.

rhattie-ey (dismissed):

rhattie-ey:
We should merge this after #895 to make resolution simpler.

mmumby:
@rhattie-ey I'm good with that plan!

fw-noel (approved):

rhattie-ey (approved):

PR #899, opened 2023-03-01
HCAP-1432 Apple Silicon computers development

vanics:
Add documentation and scripts to support development on a Apple Silicon computers


rhattie-ey (approved):
Looks good to me! I'm tempted to see how this runs on ARM Linux too, once it's merged - could be worth renaming the dockerfile and commands from `m` to `arm` if that's the case.

vanics:
@rhattie-ey, I like that! That's a good point and technically and it should be supported by linux running on a arm64 architecture as well, so I renamed the files.

rhattie-ey (approved):

PR #900, opened 2023-03-01
fix: status sites relate to siteID not site.id

fw-noel:
Realized while looking through prod data that participant statuses refer to a site's `siteId` rather than a site's `id`. You can see examples of this in the `employerSpecificJoin` and other such joins.

This should fix the phases request


mmumby (approved):
good catch!

rhattie-ey (approved):
It's good to see this fixed!

PR #903, opened 2023-03-02
HCAP-1416: Make server scripts TS compatible

rhattie-ey:
Really small PR. Renames server scripts to *.ts, and fixes a couple minor type issues. Does **not** ensure thorough type safety of scripts.


server/scripts/feed-data.ts
```diff
@@ -26,17 +26,18 @@ import { collections } from '../db/schema';
     // all responses to inserts, including duplicates and errors
     const response = [];
     // for debugging: the last record attempted to process before error
-    /**
-     * @type {any}
-     */
-    let lastInsert = '';
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let lastInsert: any = '';
```
vanics, line 31: This is odd, I assume based off the explicit exception above that this needs to be any.

vanics:
Are you sure you don't want to reduce the complexity of some of those functions? A complexity of 27 seems high. 

I am just joking... You didn't come up with that logic.

I quite like SonarCloud code complexity assessment, though. It's nice the way it quantify things. That said we really want to spend our time refactoring other places. :)

image

server/scripts/feed-data.ts
rhattie-ey: Yes, because it could be a string, or an arbitrary object of unknown value. We don't care about its properties, since it's just for printing anyways, so `any` it is.

mmumby (approved):
Tested the scripts, all work as they should :) nice work, once again!

PR #904, opened 2023-03-02
HCAP-1435: Clean up duplicate script code (participants-stats)

rhattie-ey:
Must rebase and merge after #903.

This PR fixes the presence of 4 redundant scripts, unifying them all into one stats script. This is just a quick and simple refactor I was considering adding to #903, but decided to split into a different PR instead.


mmumby (approved):
Looks good!

PR #906, opened 2023-03-03
Bulk Allocations - HCAP-1282/1283/1297

mmumby:
# Work completed
- Ability to add allocations in bulk
- Added multi-select to the site View table if MOH and 'Set Allocation' button if MOH
- Added bulk allocation form
- added bulk allocation POST API
- Added query params to GetAllPhases to conditionally include and array of allocations
- Added bulkAllcation service that checks for the existence of an allocation and either updates or creates a new one.
- Added logic for UI `acknowledgement` when overriding existing allocations 

## Follow-up work
- Test suites (e2e, server, cypress)

### UI
Screen Shot 2023-03-03 at 1 08 20 PM
Screen Shot 2023-03-03 at 1 08 33 PM
- The alert has a scroll if over 100px
Screen Shot 2023-03-03 at 1 08 59 PM

### Video of feature/errors/validations

https://user-images.githubusercontent.com/25125247/222812362-1f7ecddb-47af-465d-b9cc-2f5fa7fc6e5f.mov


client/src/components/modal-forms/BulkAllocationForm.js
```diff
@@ -0,0 +1,164 @@
...
+    allocation: '',
+    phase_id: '',
+    existingAllocations: false,
+    acknowledgement: false,
+  };
+
+  const handleSetAllocation = (phaseId, setValue) => {
+    // get all allocations associated to the selected phase
+    // filter the allocations and return all that are associated to the sites selected.
+    const allocations = phases
```
fw-noel, line 52: This looks like it could be potentially optimized further. Lots of array functions going on that makes it a little harder to read and I wonder if it could be shortened. Don't know how yet, but I'm wondering if maybe some sort of dictionary of allocations that we could pull from.

Feel free to disagree though!

client/src/components/modal-forms/BulkAllocationForm.js
```diff
@@ -0,0 +1,164 @@
...
+      .find(({ id }) => id === phaseId)
+      ?.allocations.filter(({ site_id }) => siteIds.includes(site_id))
+      .map(({ allocation, phase_id, site_id }) => ({
+        allocation,
+        phase_id,
+        site_id,
+        siteName: sites.find((site) => site.id === site_id).siteName,
+      }));
+
+    setExistingAllocations(allocations);
```
fw-noel, line 62: Not sure I understand why this needs to be in state AND in the form values? Is it in values because it's being submitted as a payload on submit?

client/src/pages/private/SetBulkAllocation.js
```diff
@@ -0,0 +1,57 @@
...
+
+  const openBulkAllocationModal = () => {
+    setActiveModalForm('bulk-allocation');
+  };
+
+  const fetchData = useCallback(async () => {
+    let phases = await fetchPhases('?includeAllocations=true');
+    setPhases(phases);
+    setIsPendingRequests(false);
+  }, [fetchPhases, setIsPendingRequests, setPhases]);
```
fw-noel, line 31: Github Actions has a format check here

client/src/services/phases.js
```diff
@@ -26,8 +26,9 @@ const mapPhasesResponse = async (data, columns) => {
   return rowsData;
 };
 
-export const fetchPhases = async () => {
-  const response = await fetch(`${API_URL}/api/v1/phase/`, {
+export const fetchPhases = async (queryString = null) => {
+  const url = queryString ? `${API_URL}/api/v1/phase${queryString}` : `${API_URL}/api/v1/phase/`;
```
fw-noel, line 30: Necessary? Can't we just remove the trailing forward slash and add ${queryString} at the end? Or would that go `api/v1/phasenull`?

server/services/phase.ts
```diff
@@ -111,11 +111,22 @@ export const getAllSitePhases = async (siteId: number): Promise<sitePhase[]> =>
...
   // NOTE: this will not allow for full access to the phase list once it hits that 100k limit!
   // If there is any chance of this hitting that number, or if performance starts to suffer,
   // pagination support should be added.
-  const phases = await dbClient.db[collections.GLOBAL_PHASE].find({}, { limit: 100000 });
+  const phases = await dbClient.db[collections.GLOBAL_PHASE]
+    .join({
+      allocations: {
+        relation: collections.SITE_PHASE_ALLOCATION,
+        type: 'LEFT OUTER',
+        omit: !query?.includeAllocations,
```
fw-noel, line 123: Neat

server/validators/allocation.ts
```diff
@@ -32,3 +32,27 @@ export const UpdateAllocationSchema = yup
...
+      .required('Allocations are required')
+      .lessThan(100000, 'Must be less than 100,000')
+      .test(
+        'validate-blank-or-number',
+        'Must be a positive number',
+        validateBlankOrPositiveInteger
+      ),
+    phase_id: yup.number().required('Allocations must be related to a phase'),
+    siteIds: yup.array().of(yup.number()).required('Allocations must be related to a site'),
+    existingAllocations: yup.boolean().required(),
```
fw-noel, line 51: I think we can keep these (existingAllocations and acknowledgement) as a strictly frontend validation, I don't think it makes a lot of sense in a post request

server/services/allocations.ts
```diff
@@ -57,3 +57,25 @@ export const updateAllocation = async (allocationId: number, allocation, user: H
   const res = await dbClient.db[collections.SITE_PHASE_ALLOCATION].update(allocationId, data);
   return res;
 };
+
+export const createBulkAllocation = async (payload: any, user: HcapUserInfo) => {
+  const updateResults = [];
+  await Promise.all(
+    payload.siteIds.map(async (id) => {
+      const allocation = await getAllocation(id, payload.phase_id);
```
fw-noel, line 65: This feels like a lot of requests as well. Can we write up a service for getting allocation for multiple sites in a phase? Or maybe grabbing all and filtering based on siteId?

server/routes/phase.js
```diff
@@ -1,4 +1,5 @@
 import express from 'express';
+import * as qs from 'querystring';
```
fw-noel, line 2: Github Actions says this isn't used?

fw-noel (commented):
Looks alright! I had a few nitpicks and a few optimization pieces.

client/src/services/phases.js
mmumby: `api/v1/phasenull` is exactly what it does! I was also hoping to simplify here

client/src/components/modal-forms/BulkAllocationForm.js
mmumby: The `valuesexistingAllocations` is just a boolean, There didn't seem to be a need to have the entire array in the payload for the yup validation. the state variable `existingAllocations` is being used to display the site/allocations in the alert message.

client/src/services/phases.js
mmumby: alternatively, I could default it to an empty string `queryString = ' ' ` and remove the const url

server/routes/phase.js
mmumby: removed! not sure why my linting didn't pick this up on the commit hook

client/src/pages/private/SetBulkAllocation.js
mmumby: removed!

server/validators/allocation.ts
mmumby: good call, removed

client/src/services/phases.js
fw-noel: Nah, this is fine!

client/src/components/modal-forms/BulkAllocationForm.js
mmumby: It can be shortened. It does require two array.filters/finds. I have split it into two consts to make it easier to read: 
```
 // get all allocations associated to the selected phase
    const phaseAllocations = phases.find(({ id }) => id === phaseId)?.allocations;
    // filter the allocations and return all that are associated to the sites selected.
    const phaseSiteAllocations = phaseAllocations.filter(({ site_id }) =>
      siteIds.includes(site_id)
    );
```

I was then mapping to format the object as I needed the site name to be included in the object, (Initially, I was passing this array to the yup validator/BE so I intentionally made the object smaller)  But that isn't necessary anymore, and I can read the siteName a different way. (created an object hash)

server/services/allocations.ts
mmumby: good call - I have updated it to make 1 query to `getAllocationsForSites`, I then check to ensure the site being referenced exists in the array or existing allocations `allocationFound`, if it is, I update allocation, if not I create allocation. Let me know if this solution is more aligned with what you were thinking.

fw-noel (approved):
Approved! The only other thing is that it might be worth combining the POST set allocation and POST bulk allocation endpoints now, so we won't have to worry about it later

client/src/utils/date.js
```diff
@@ -10,6 +10,7 @@ dayjs.extend(isBetween);
 export const dayUtils = dayjs;
 
 export const dateToString = (dateObj) => dayjs(dateObj).format('YYYY/MM/DD');
+export const dateToTextString = (dateObj) => dayjs.utc(dateObj).format('MMM DD, YYYY');
```
rhattie-ey, line 13: Not a fan of these names. All strings are text strings... the difference is the formatting. The names of these functions should reflect that.

server/routes/allocation.js
```diff
@@ -78,4 +87,35 @@ router.patch(
   })
 );
 
+// Create bulk allocation: POST
```
rhattie-ey, line 90: At some point we might wanna roll this into the main allocation route, kind of like what we discussed with the other upcoming bulk work. This is fine for now though.

server/services/phase.ts
```diff
@@ -111,11 +111,22 @@ export const getAllSitePhases = async (siteId: number): Promise<sitePhase[]> =>
   }));
 };
 
-export const getAllPhases = async () => {
+export const getAllPhases = async (query = null) => {
```
rhattie-ey, line 114: Document what this new query parameter does with JSDoc, and add a type. Neither is immediately clear at the moment, and ideally we should be striving for as much strictly defined typing as possible.

The name is also a potential source of confusion - I'd expect a parameter called `query` to be something like a SQL query in this region of the codebase, but here it looks like, presumably, query parameters from a request? I haven't had a chance to check in-IDE or look in greater depth, but that's what it looks like from here.

Considering all you're doing with this property is reading a bool from it, I'd highly suggest just replacing `query` with an `includeAllocations` parameter, and adjusting your call to `getAllPhases` to pass it that specific property of the query object.

rhattie-ey (commented):
Looks good, although there is some potential room for cleanup.

server/routes/allocation.js
mmumby: That was mentioned in the dev channel - I did move it OOS at this point due to the logic of post vs patch whereas the example we discussed only involved post. If we merge this feature into the existing POST, some updates/decisions will need to be made regarding the existing UI. (ie if for some reason they try to post a new allocation and one exists, do we allow/override and inform the user, or prevent that from occurring as it currently is implemented). 
If there is time after testing is completed I can look into merging the two.

client/src/utils/date.js
mmumby: Updated to differentiate between long vs short form. If you have a better name suggestion let me know!

server/services/phase.ts
mmumby: Also - I created a `phase` type as it was a different response than `sitePhase`, which exposed the inconsistencies of the response in regards to casing (snake vs camel). I'm making it OOS for this ticket as the API was already implemented and there is still lots of testing to complete before EOS, but it's a Good Candidate for a tech debt ticket to update the casing, and decide on preferred formatting for APIs moving forward. (again, if there is time after testing I can update)

client/src/utils/date.js
rhattie-ey: Looks good enough to me!

server/services/phase.ts
```diff
@@ -111,16 +125,22 @@ export const getAllSitePhases = async (siteId: number): Promise<sitePhase[]> =>
   }));
 };
 
-export const getAllPhases = async (query = null) => {
+/**
+ * Gets all phases with or without associated allocations
+ *
+ * @param includeAllocations  condition to include associated allocation records
+ * @returns phase[]   An array of phases with formatted
```
rhattie-ey, line 132: Typo in description - with formatted what, exactly?

server/services/phase.ts
rhattie-ey: Good point in regards to casing inconsistency, let's have a discussion about how we want to handle that later, and make a ticket for it. I like the new changes you made here.

server/services/phase.ts
mmumby: fixed! the sun was shining bright on my screen this morning, tough to see the comments.

fw-noel (approved):

rhattie-ey (approved):

PR #912, opened 2023-03-08
HCAP-1428 - Allocation not showing up on site detail view

mmumby:
## Work completes
- updated `getSiteByID` to include a query to get the current phase allocation record. (This should be the same allocation visible on the site view table) 
 
Screen Shot 2023-03-08 at 3 58 01 PM

### DIFF
--- a/server/services/employers.ts
+++ b/server/services/employers.ts
@@ -1,4 +1,3 @@
-import dayjs from 'dayjs';
 import { dbClient, collections } from '../db';
 import { validate, EmployerSiteBatchSchema } from '../validation';
 import { userRegionQuery } from './user';
--- a/server/services/employers.ts
+++ b/server/services/employers.ts
@@ -3,6 +3,7 @@ import { validate, EmployerSiteBatchSchema } from '../validation';
 import { userRegionQuery } from './user';
 import type { HcapUserInfo } from '../keycloak';
 import { formatDateSansTimezone } from '../utils';
+import { Allocation } from '../services/allocations';
 
 export interface EmployerSite {
   id: number; // Internal ID for site
@@ -167,7 +168,7 @@ export const getSiteByID = async (id: number): Promise<EmployerSite> => {
     throw new Error(`No site found with id ${id}`);
   }
 
-  const currentAllocation = await dbClient.runRawQuery(
+  const allocationResponse: Allocation[] = await dbClient.runRawQuery(
     `
     SELECT spa.allocation
       FROM site_phase_allocation spa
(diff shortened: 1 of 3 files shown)


server/services/employers.ts
```diff
@@ -168,6 +167,19 @@ export const getSiteByID = async (id: number): Promise<EmployerSite> => {
     throw new Error(`No site found with id ${id}`);
   }
 
+  const currentAllocation = await dbClient.runRawQuery(
```
rhattie-ey, line 170: Make returned shape from this query clearer with a type annotation.
```suggestion
  const allocationResponse: [{ allocation: number }?] = await dbClient.runRawQuery(
```

server/services/employers.ts
```diff
@@ -168,6 +167,19 @@ export const getSiteByID = async (id: number): Promise<EmployerSite> => {
...
+    SELECT spa.allocation
+      FROM site_phase_allocation spa
+        JOIN (
+          SELECT id FROM phase
+          WHERE CURRENT_DATE BETWEEN start_date AND end_date
+          LIMIT 1
+        ) p ON p.id = spa.phase_id
+      WHERE spa.site_id = $1;
+    `,
+    [site[0].id]
```
rhattie-ey, line 181: Verify that `allocationResponse` constains data. If the database doesn't find any valid phase allocations (i.e. there's no allocations for this site at the present time), it will **not** give you a row in the result, causing a server error if you try to read that row. This check should be done right next to the query for readability.

Note: here I am using NaN, but if 0 is preferred, feel free to switch that out.

```suggestion
  const currentAllocation = allocationResponse.length ? allocationResponse[0].allocation : NaN;
```

server/services/employers.ts
```diff
@@ -198,7 +210,12 @@ export const getSiteByID = async (id: number): Promise<EmployerSite> => {
...
-  site[0].nonHcapHires = nonHcapHires;
-  return site[0];
+
+  const siteResponse = {
+    ...site[0],
+    hcapHires,
+    nonHcapHires,
+    allocation: currentAllocation[0].allocation,
+  };
+  return siteResponse;
```
rhattie-ey, line 220: No need to make an object and then return it. Simpler to just return directly.
```suggestion
  return {
    ...site[0],
    hcapHires,
    nonHcapHires,
    allocation: currentAllocation,
  };
```

rhattie-ey (changes requested):
Found a critical bug that causes a server error if the viewed site does not have allocations.

server/services/employers.ts
mmumby: reverted back to the original return!

fw-noel (approved):

rhattie-ey (approved):
Looks good! Nice to see the additional tests here, too.

PR #908, opened 2023-03-08
HCAP-1396: Create a seeding script for other tables

rhattie-ey:
Original ticket: [HCAP-1396](https://eydscanada.atlassian.net/browse/HCAP-1396)

### `feed-data` rewrite

Overhauled the entire script, rewriting it from the ground up.

Features:
* Encompasses all tables needed for testing
* Simpler
* More robust error handling
  * Specific logic to handle and report foreign key errors
* More human-friendly console output
* Fully compliant with strict TS
* Generalized enough to work for any table in the database
* Importable for use in wrapper scripts
* Has a `--skip-missing-tables` flag to allow population of incomplete databases (such as for server tests)

### Test data generation

[Jupyter notebook used for generation](https://gist.github.com/rhattie-ey/1efbec9b828f076ed603e050e284a611)

Created new test data CSVs for the following tables:

* cohort_participants
* cohorts
* participant_post_hire_status
* participants
* participants_status
* phases
* post_secondary_institutions
* return_of_service_status
* site_phase_allocation

### Misc

* Updated docs
* Removed the now unneeded `copy-sample-data` command
* Updated scripts, makefiles, and actions to use new scripts
* Deleted old test data from server/script/csv and server/test-data, including all XLSX files other than those in server/tests/mock
* Flagged code that is now deprecated due to this PR
* Updated Node.js to LTS release 18.15.0
* Updated tests to work with new data
* Updated GitHub actions to use only feed-data for seeding

## Next steps
* [Improve realism of participant statuses (HCAP-1464)](https://eydscanada.atlassian.net/browse/HCAP-1464)
* [Convert migration XLSX assets to CSV (HCAP-1465)](https://eydscanada.atlassian.net/browse/HCAP-1465)

PR #913, opened 2023-03-10
Feature - HCAP-1430

mmumby:
## Work completed
- Enhanced the `HireForm` to `fetchSitePhases` when a site is selected in the dropdown, 
- The currentPhase is determined by the hiredDate selected
- An alert is displayed with the number of allocations/remaining hires for that phase is shown to the user
Screen Shot 2023-03-10 at 8 02 41 AM
Screen Shot 2023-03-10 at 8 20 51 AM


cypress/integration/participantStatus.spec.js
```diff
@@ -9,7 +9,7 @@ describe('Participants status suite', () => {
     cy.contains('button', 'View My Candidates').click();
   };
 
-  const interviewParticipantForm = (siteId) => {
+  const interviewParticipantForm = () => {
     // let it select today as the date
```
mmumby, line 13: `siteId` wasn't being used in this function.

rhattie-ey:
@mmumby this can now be resumed, since #908 is merged.

client/src/components/modal-forms/HireForm.js
```diff
@@ -17,66 +20,115 @@ const hireInitialValues = {
...
+        const currentPhase =
+          values.hiredDate && values.site
+            ? phases?.filter(
+                (phase) =>
+                  Date.parse(phase.startDate) <= Date.parse(values.hiredDate) &&
+                  Date.parse(values.hiredDate) <= Date.parse(phase.endDate)
+              )[0]
+            : null;
+
+        return (
```
rhattie-ey, line 44: For the sake of clarity I would prefer to see some of this logic brought out into a function prior to `return`. Not a huge fan of defining a function that returns JSX *inside the return JSX of a function*.

cypress/integration/participantStatus.spec.js
```diff
@@ -9,21 +29,26 @@ describe('Participants status suite', () => {
...
     cy.contains('button', 'Submit').click();
   };
 
   const hireParticipantForm = (siteId) => {
     // both dates
-    cy.get('input[name=DateHired]').clear().type('2022/01/14');
-    cy.get('input[name=StartDate]').clear().type('2022/01/15');
+    // Ensure dates selected are within the phase range of Sacred Macaw
+    cy.get('input[name=DateHired]').clear().type('2023/01/14');
+    cy.get('input[name=StartDate]').clear().type('2023/01/15');
```
rhattie-ey, line 43: Not the biggest fan of hardcoding the dates, even if that's what the existing test did. Perhaps this could be extracted from the DB or test data CSV beforehand, in case test phases change down the line?

cypress/integration/participantStatus.spec.js
```diff
@@ -9,21 +29,26 @@ describe('Participants status suite', () => {
...
     // select site
     cy.get('#mui-component-select-site').click();
     cy.get(`li[data-value='${siteId}']`).click();
     // acknowledge participant accepted offer in writing
     cy.get('input[name="acknowledge"]').click();
+    //  expect alert with allocations/remainingHires to exist
+    cy.wait(1000);
```
rhattie-ey, line 50: Why the one second wait? I'd like to see at least a comment explaining why - although ideally I'd like to think there's a better solution here, especially considering how long our tests already take. Anyone with more Cypress experience (@fw-noel?) want to chime in here?

cypress/integration/participantStatus.spec.js
```diff
@@ -9,21 +29,26 @@ describe('Participants status suite', () => {
...
+    cy.get('.MuiAlert-message').contains(`This site has 100 allocations assigned`);
```
rhattie-ey, line 51: Any reason for this to be a template?

```suggestion
    cy.get('.MuiAlert-message').contains('This site has 100 allocations assigned');
```

rhattie-ey (commented):
Looks good, although I have a couple questions.

cypress/integration/participantStatus.spec.js
```diff
@@ -140,7 +165,7 @@ describe('Participants status suite', () => {
   };
 
   it('Flow test: available --> engage by HA --> hired --> archived & acknowledged', () => {
-    const participantId = 4;
+    const participantId = 5;
```
fw-noel, line 168: This is due to the new test data?

cypress/integration/participantStatus.spec.js
fw-noel: I am curious if its necessary. Its probably to wait for the alert, but are our other tests needing it?

cypress/integration/participantStatus.spec.js
mmumby: No, I changed it during local testing as that participant was already used in my local env. I'll revert

cypress/integration/participantStatus.spec.js
mmumby: Unintentional result of a copy/paste situation. I'll fix it!

cypress/integration/participantStatus.spec.js
fw-noel: It seems to be used in a few other tests, so with consistency in mind, I'm okay with this. Wonder if they can be shortened to a half second or something

client/src/components/modal-forms/HireForm.js
```diff
@@ -1,8 +1,12 @@
-import React from 'react';
+/* eslint-disable */
```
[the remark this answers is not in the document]
mmumby: just removed it!

cypress/integration/participantStatus.spec.js
mmumby: So - the test was passing locally with a `wait(1000)`, but not in the pipeline. I increased the wait to `wait(2000)` and the tests are now passing in the pipeline. The wait is necessary to allow the component to complete the async functions before the alert exists on the UI.

fw-noel (approved):

vanics (approved):
It looks good!

cypress/integration/participantStatus.spec.js
mmumby: I've updated to use the csv data, select the phase I want, and im adding a few weeks to the phase.startDate to ensure the hire occurs within that phase.

rhattie-ey (approved):

PR #914, opened 2023-03-10
HCAP-1461 - Site name too long

mmumby:
## work completed
- Added max 255 characters for siteName on all yup validations
- Added tests to check failure if siteName is too long.
- Used addEllipsisMask on the link and modal header for siteName if over 50 characters (For the link - the full site name is displayed directly under, for the modal header if the siteName was too long the modal wouldn't scroll thus preventing editing. The full name is still visible in the form)

# clean up
- Deleted the `truncateString` function as a function already existed that did the same thing. `addEllipsisMask`


fw-noel (dismissed):

rhattie-ey:
`addEllipsisMask` could use a better name, but that's OOS for this PR.

rhattie-ey (approved):
Looks good!

rhattie-ey:
Just noticed the dupe import in `SiteViewDetails.js` flagged by Sonar, once that's fixed consider this approved by me.

rhattie-ey (approved):

PR #918, opened 2023-03-13
HCAP-1402 Make server routes directory TS

rhattie-ey:
Renames all files in routes/ to .ts, and fixes a minor typing error in `set-status`.


mmumby (approved):
nice update!

PR #922, opened 2023-03-15
fix: increase pending users limit

fw-noel:
Sets the pending users to have a max like most other user requests. (Likely not a long term solution).
Sets brief representation as well, as it's not necessary to get more info than that


fw-noel:
screen

This is what it takes to displace the last column.

mmumby (dismissed):
🎉

mmumby (approved):

PR #924, opened 2023-03-16
HCAP-1449 - Remove ability for employer to bulk engage participants

mmumby:
## Work completed

Note - The FE isn't using the `employer-actions/bulk-engage` API, it is looping through participants and hitting `employer-actions`, so I did not make any changes to the permissions of that API as employers can still engage participants individually. (The FE is currently making 3 API calls per participant in the bulk engage list)
- removed employer permission from bulk-engage API (although it's deprecated/only being used by tests) 
- updated test suites to include failure if unauthenticated, 
- added check for HA on UI to see multi-select and bulk engage button

HA view:
Screen Shot 2023-03-16 at 10 56 24 AM

Employer view:
Screen Shot 2023-03-16 at 10 57 26 AM

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -202,6 +202,7 @@ const ParticipantTable = () => {
   const isSuperUser = roles.includes('superuser');
   const isAdmin = isMoH || isSuperUser;
   const isEmployer = roles.includes('health_authority') || roles.includes('employer');
+  const isHA = roles.includes('health_authority');
 
   const fetchParticipants = async () => {
     if (!columns) return;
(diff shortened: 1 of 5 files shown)


client/src/pages/private/ParticipantTable.js
```diff
@@ -202,6 +202,7 @@ const ParticipantTable = () => {
   const isSuperUser = roles.includes('superuser');
   const isAdmin = isMoH || isSuperUser;
   const isEmployer = roles.includes('health_authority') || roles.includes('employer');
+  const isHA = roles.includes('health_authority');
```
vanics, line 205: You could also have done something like:

```
const isHA = roles.includes('health_authority');
const isEmployer = isHA || roles.includes('employer');
```

vanics (dismissed):

fw-noel (approved):

vanics (approved):

PR #926, opened 2023-03-16
HCAP-1443 MoH/HA See site that hired participant

rhattie-ey:
Work completed:

* Added site and status to details page

### DIFF
--- a/client/src/pages/private/ParticipantDetailsView.js
+++ b/client/src/pages/private/ParticipantDetailsView.js
@@ -265,16 +265,19 @@ export default () => {
             </Box>
             <Typography variant='h2'>Participant Details</Typography>
             <Grid container spacing={2} className={classes.gridSection}>
-              {Object.keys(keyLabelMap).map((key) => (
-                <Grid key={key} item xs={12} sm={6} xl={3}>
-                  <Typography variant='body1'>
-                    <b>{keyLabelMap[key]}</b>
-                  </Typography>
-                  <Typography test-id={'participantDetailsView' + key} variant='body1'>
-                    {participant[key]}
-                  </Typography>
-                </Grid>
-              ))}
+              {Object.keys(keyLabelMap).map(
+                (key) =>
+                  participant[key] && (
+                    <Grid key={key} item xs={12} sm={6} xl={3}>
+                      <Typography variant='body1'>
+                        <b>{keyLabelMap[key]}</b>
+                      </Typography>
+                      <Typography test-id={'participantDetailsView' + key} variant='body1'>
+                        {participant[key]}
+                      </Typography>
+                    </Grid>
+                  )
+              )}
             </Grid>
 
             {/* Participant RoS Info */}
--- a/client/src/services/participant-details.js
+++ b/client/src/services/participant-details.js
@@ -25,6 +25,17 @@ const getInterestLabel = (isInterested) => {
   }
 };
 
+/**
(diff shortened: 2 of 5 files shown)


client/src/pages/private/ParticipantDetailsView.js
```diff
@@ -265,16 +265,19 @@ export default () => {
...
-                  </Typography>
-                  <Typography test-id={'participantDetailsView' + key} variant='body1'>
-                    {participant[key]}
-                  </Typography>
-                </Grid>
-              ))}
+              {Object.keys(keyLabelMap).map(
+                (key) =>
+                  participant[key] && (
+                    <Grid key={key} item xs={12} sm={6} xl={3}>
```
mmumby, line 271: Do we still want to view the label regardless if the value is null/undefined?

rhattie-ey: Potentially, although I feel like it's cleaner to just hide it. @fw-noel, thoughts?

client/src/services/participant.js
```diff
@@ -44,12 +45,21 @@ export const fetchParticipant = async ({ id }) => {
     const { participant } = await resp.json();
     const cohort = await fetchParticipantCohort({ id });
     const postHireStatus = await fetchParticipantPostHireStatus({ id });
+    const status = participant.latestStatuses.length
+      ? participant.latestStatuses[0].status
+      : 'available';
+    const site =
+      status === 'hired'
+        ? await (await fetchSite(participant.latestStatuses[0].siteId)).json()
+        : null;
```
mmumby, line 54: Is 'available' a new status? (I was looking at the server const ParticipantStatus for reference)

client/src/services/participant-details.js
```diff
@@ -25,6 +25,17 @@ const getInterestLabel = (isInterested) => {
...
+ * @param {string} status Status string from DB
+ * @returns Title case formatted status
+ */
+const getStatusLabel = (status) =>
+  status
+    .split('_')
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
+
 export const getParticipantPageLabel = (pageName) => {
```
mmumby, line 39: Could you move this to a util file? Seems like it could be used elsewhere in the future.

client/src/services/participant.js
rhattie-ey: It's what the table labels participants with no status, so I tried to keep the same convention going.

client/src/services/participant-details.js
rhattie-ey: Potentially, yeah.

rhattie-ey: (I assume you mean `getStatusLabel` and not `getParticipantPageLabel`.)

mmumby: Yes I did!

client/src/pages/private/ParticipantDetailsView.js
rhattie-ey: Alternative idea: put the site name right in the status string? So it would say something like "Available" or "Hired - Sleepy Cat Hospital" for status.

fw-noel (approved):

client/src/pages/private/ParticipantDetailsView.js
fw-noel: Hiding it I think is the way to go

PR #927, opened 2023-03-16
HCAP-1415 - Bulk Graduate

mmumby:
## Work completed
- Updated existing API - `/post-hire-status` to accept an array of `participantIds` instead of a single ID. 
- Updated the FE  cohort-details page to include 'Bullk Graduate' button, multi select on the cohort page, participant table - visible for HA
- If the form is being rendered by "bulk graduate' disable the 'unsuccessful flow' and only allow graduation to be set.
- If a participant already has a graduation status, disallow the user from opening the form, show Alert above the table (See screenshot)

UI: 
Screen Shot 2023-03-17 at 2 30 20 PM
Screen Shot 2023-03-20 at 2 21 35 PM
Screen Shot 2023-03-20 at 4 53 47 PM

modal/form if setting the status on an individual
Screen Shot 2023-03-17 at 2 31 40 PM


server/services/cohorts.ts
```diff
@@ -93,9 +93,9 @@ export const filterCohortParticipantsForUser = (cohortParticipants, user) => {
   if (user.isHA) {
     // Remove participants hired outside of HA's region
     // participant.siteJoin is joined based on hired status' siteId
-    return cohortParticipants.filter((participant) =>
-      user.regions.includes(participant.siteJoin.body.healthAuthority)
-    );
+    return cohortParticipants.filter((participant) => {
+      return user.regions.includes(participant.siteJoin?.body?.healthAuthority);
+    });
```
mmumby, line 98: This was causing a 500 error locally if logged in as HA visiting `cohorts/:id`

client/src/components/fields/RenderRadioGroup.js
```diff
@@ -43,21 +43,24 @@ export const RenderRadioGroup = ({
...
-              control={<Radio color={option.color || 'primary'} />}
-            />
-            {hiddenCheckbox?.fields.includes(field.value) &&
-              option.value === field.value &&
-              hiddenCheckbox.node}
-          </Fragment>
-        ))}
+        {options.map(
+          (option) =>
+            option && (
```
rhattie-ey, line 48: What is the purpose of this check? Why might `options` contain a falsy value?

client/src/components/modal-forms/ManageGraduationForm.js
```diff
@@ -4,13 +4,31 @@ import React, { useMemo } from 'react';
...
 import { Button } from '../../components/generic/Button';
+import { makeStyles } from '@material-ui/core/styles';
 import { AuthContext } from '../../providers';
 import { ParticipantPostHireStatusSchema } from '../../constants/validation';
 import { postHireStatuses } from '../../constants';
 
-export const ManageGraduationForm = ({ initialValues, onClose, onSubmit, cohortEndDate }) => {
+const useStyles = makeStyles(() => ({
+  list: {
+    maxHeight: '100px',
```
rhattie-ey, line 14: This is a container with text inside, correct? Generally I'd prefer to see em or rem used here if that's the case. It's a nitpick though.

server/routes/post-hire-status.ts
```diff
@@ -26,23 +26,38 @@ router.post(
     const user = userId || localUserId;
     // Validate the request body
     await validate(ParticipantPostHireStatusSchema, body);
-    const { participantId } = body;
+    const { participantIds } = body;
     // Check participant exists
-    const [participant] = await getParticipantByID(participantId);
-    if (!participant) {
+    const notParticipants = [];
```
rhattie-ey, line 31: Both of these arrays should ideally have typing, but especially notParticipants.

server/routes/post-hire-status.ts
```diff
@@ -26,23 +26,38 @@ router.post(
...
+    );
+
+    if (notParticipants.length) {
       logger.error({
         action: 'post-hire-status_post',
-        message: `Participant does not exist with id ${participantId}`,
+        message: `Participant(s) do not exist with id ${JSON.stringify(notParticipants)}`,
       });
+
       return res.status(422).send('Participant does not exist. Please check participant ID');
```
rhattie-ey, line 45: Could potentially use a template here like in the logger error, to make it clearer what went wrong from the client end (may help in Cypress debugging, for example).

server/routes/post-hire-status.ts
```diff
@@ -26,23 +26,38 @@ router.post(
...
     // Check participant exists
-    const [participant] = await getParticipantByID(participantId);
-    if (!participant) {
+    const notParticipants = [];
+    await Promise.all(
+      participantIds.map(async (id) => {
+        const [participant] = await getParticipantByID(id);
+        if (!participant) notParticipants.push(id);
+      })
+    );
```
rhattie-ey, line 37: This can be made cleaner and more purely functional with a filter instead of `push`.

```suggestion
    const notParticipants = (
      await Promise.all(participantIds.map(async (id) => ({id, valid: await getParticipantByID(id)).length > 0)})
    ).filter((participant) => participant.valid);
```

server/routes/post-hire-status.ts
```diff
@@ -26,23 +26,38 @@ router.post(
...
     // Get Cohort
-    const cohorts = await getAssignCohort({ participantId });
-    if (cohorts.length === 0) {
+    const noCohorts = [];
+    await Promise.all(
+      participantIds.map(async (id) => {
+        const cohorts = await getAssignCohort({ participantId: id });
+        if (cohorts.length === 0) noCohorts.push(id);
+      })
+    );
```
rhattie-ey, line 55: As with participants, this can be simplified with a filter. See my other suggestion for an implementation example.

server/services/cohorts.ts
```diff
@@ -94,7 +94,7 @@ export const filterCohortParticipantsForUser = (cohortParticipants, user) => {
     // Remove participants hired outside of HA's region
     // participant.siteJoin is joined based on hired status' siteId
     return cohortParticipants.filter((participant) =>
-      user.regions.includes(participant.siteJoin.body.healthAuthority)
+      user.regions.includes(participant.siteJoin?.body?.healthAuthority)
```
rhattie-ey, line 97: Some of this optional chaining could be avoided by adding typing to `cohortParticipants`, but this also works.

client/src/components/fields/RenderRadioGroup.js
mmumby: In ManageGraduationForm.js I conditionally rendered the options depending on if bulk graduation was occurring (We only want to allow the successful flow) 
```
 options={[
                    {
                      value: postHireStatuses.postSecondaryEducationCompleted,
                      label: 'Graduated',
                    },
                    !isBulkGraduate && {
                      value: postHireStatuses.cohortUnsuccessful,
                      label: 'Unsuccessful cohort',
                    },
                  ]}
```
This returned options as [object, false], which was displaying the radio button without anything else.

server/routes/post-hire-status.ts
mmumby: The push was partly due to the error message getting logged, I wanted to inform which id's didn't have participants
`Participant(s) do not exist with id ${JSON.stringify(notParticipants)}`

vanics (dismissed):
It looks good!

@rhattie-ey provided a few nice suggestions that you might want to consider, but overall is a sound implementation.

client/src/components/modal-forms/ManageGraduationForm.js
vanics: That's an interesting suggestion. I personally tend not to use `em` and `rem` as much but when talking about accessibility they make way more sense. 

As a side note for reference, @mmumby if you consider to follow up on this nitpick consider that:
EM is relative to the parent element’s font size, so if you wish to scale the element’s size based on its parent’s size, use EM.

REM is relative to the root (HTML) font size, so if you wish to scale the element’s size based on the root size, no matter what the parent size is, use REM. If you’ve used EM and are finding sizing issues due to lots of nested elements, REM will probably be the better choice.

My recommendation would be to be consistent with other parts of the application, so whatever we use the most in a specific context ideally would be the way to go.

mmumby: Great discussion - I removed the Alert from this form, and no longer needed this class! I will keep this in mind for sizing in the future though.

server/routes/post-hire-status.ts
mmumby: This example doesn't work as anticipated - We cannot chain a filter onto a Promise.all(), we would require a second variable to filter/map to get the correct format for the alert. 

```
const notParticipants: {
      id: string;
      valid: boolean;
    }[] = await Promise.all(
      participantIds.map(async (id) => ({ id, valid: (await getParticipantByID(id)).length > 0 }))
    );

    const filteredParticipants: {
      id: string;
      valid: boolean;
    }[] = notParticipants.filter((obj) => !obj.valid);

    if (filteredParticipants.length) {
      logger.error({
        action: 'post-hire-status_post',
        message: `Participant(s) do not exist with id **${JSON.stringify(
          filteredParticipants.map(({ id }) => id)
        )}**`,
      });

      return res
        .status(422)
        .send(
          `Participant(s) do not exist with id **${JSON.stringify(
            filteredParticipants.map(({ id }) => id)
          )}.** Please check participant ID`
        );
    }
    ```
    
 Additionally, I can handle the .map in the `filteredParticipants` const vs in JSON.stringify - Although I think this refactor as a whole does not simplify the previous implementation.

File: client/src/components/fields/RenderRadioGroup.js (hunk 2 above)
Comment (rhattie-ey) [reply client/src/components/fields/RenderRadioGroup.js:48]: In that case, I'd say you're better off filtering out the potentially false element there (`.filter(Boolean)` on the array as you define it) instead of making this logic expect arbitrary falsy values mixed in with actual options.

File: server/routes/post-hire-status.ts (hunk 4 above)
Comment (rhattie-ey) [reply server/routes/post-hire-status.ts]: Ideally `participantIds` should also have typing, but at least giving `notParticipants` a type is a big improvement.

File: client/src/components/fields/RenderRadioGroup.js (hunk 2 above)
Comment (mmumby) [reply client/src/components/fields/RenderRadioGroup.js]: Updated!

File: server/routes/post-hire-status.ts (hunk 4 above)
Comment (mmumby) [reply server/routes/post-hire-status.ts]: Updated!

File: server/routes/post-hire-status.ts (hunk 6 above)
Comment (mmumby) [reply server/routes/post-hire-status.ts]: Discussed Via teams chat. Updated.

File: server/routes/post-hire-status.ts [hunk 9]
```diff
@@ -10,57 +10,108 @@ import keycloak from '../keycloak';
...
+  body: {
+    participantIds: number[];
+    status: postHireStatuses;
+    data: {
+      graduationDate?: string;
+      unsuccessfulCohortDate?: string;
+    };
+  };
+  user: {
+    user_id: string;
```
rhattie-ey, line 27: Add an ESLint directive here to prevent errors, potentially alongside a JSDoc comment at the top of the type to explain what this is used for (and by extension why it would need a snake case property name).

server/routes/post-hire-status.ts
rhattie-ey: Looks good!

rhattie-ey (commented):
One minor change to prevent ESLint errors, and this is otherwise approved by me. Great work!

server/routes/post-hire-status.ts
mmumby: Updated!

rhattie-ey (approved):
Looks good, although I'd suggest addressing the comments before merging. I'll re-approve if needed.

server/routes/post-hire-status.ts
```diff
@@ -10,57 +11,109 @@ import keycloak from '../keycloak';
...
-    const [participant] = await getParticipantByID(participantId);
-    if (!participant) {
+    const { participantIds } = body;
+
+    const notParticipants: {
+      id: number;
+      valid: boolean;
+    }[] = (
+      await Promise.all(
+        participantIds.map(async (id) => ({ id, valid: (await getParticipantByID(id)).length > 0 }))
```
fw-noel, line 53: Bit interesting here. I think it works fine, though would've been nice to have a service to just get multiple participants by ids.

fw-noel (approved):

server/routes/post-hire-status.ts
```diff
@@ -10,57 +11,109 @@ import keycloak from '../keycloak';
...
+    status: postHireStatuses; //  String representing the status
+    data: {
+      graduationDate?: string; // Date set If status = successful
+      unsuccessfulCohortDate?: string; // Date set If status = unsuccessful
+    };
+  };
+  /** User setting the status */
+  user: {
+    user_id: string; // User id from keycloak
+    sub: string; // Additional user id from keycloak
```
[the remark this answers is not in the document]
mmumby: Updated in the follow up testing PR!

PR #928, opened 2023-03-17
fix: Add preferred location to participant when hired

fw-noel:
HCAP-1467 - Add Preferred Location to Hired Participant
If a participant gets hired outside of their preferred regions, other users who should be able to see them under "Hired Sites" cannot because of the preferred region. This work changes that so that on hire for these scenarios, they are visible to the HA.

We will need a separate script to handle 182 participants for which this is an issue. However, this should do for future user.


server/tests/util/integrationTestData.ts
```diff
@@ -74,14 +74,16 @@ interface TestSiteData {
   siteId?;
   city?: string;
   siteName?: string;
+  healthAuthority?: string;
   [key: string]: string | number | boolean | undefined;
 }
 
 export const makeTestSite = async (
-  { siteId, city, siteName, ...rest }: TestSiteData = {
+  { siteId, city, siteName, healthAuthority = 'Vancouver Island', ...rest }: TestSiteData = {
```
vanics, line 82: What's happening here?

server/services/participant-status/set-status.ts
```diff
@@ -105,7 +105,7 @@ export const setParticipantStatus = async (
     }
 
     // Load Participant
-    const participant = await tx[collections.PARTICIPANTS].findDoc({
+    const [participant] = await tx[collections.PARTICIPANTS].findDoc({
       id: participantId,
```
mmumby, line 109: nice fix!

server/tests/util/integrationTestData.ts
fw-noel: This is just for when the backend tests make test sites. I wanted to set a default for which health authority a site is in rather than defining it in all tests, and then tests on their own can set it themselves if needed (as seen on the test I added)

server/services/participant-status/set-status.ts
rhattie-ey: Agreed, nice to see this cleanup.

server/services/participant-status/handle-transition.ts
```diff
@@ -54,16 +54,39 @@ const handleArchivedTransition = async ({
...
   if (!isRosComplete) {
-    await withdrawParticipant(participant[0]);
+    await withdrawParticipant(participant);
   }
 
   return {};
 };
 
-const handleHiredTransition = async ({ tx, existingCurrentStatus, site, participantId }) => {
+// Handles the logic when a participant gets hired into a site
```
rhattie-ey, line 63: Make this a JSDoc comment so it is visible in IDEs. Also add argument typing if possible.

mmumby (approved):
Good job!

server/test-data/participants.csv
rhattie-ey, line 1: Is this to do with an actual schema change, or just fixing the CSV to be in line with how things always were?

fw-noel: Fixing the CSV to how they always were! I looked at the Test database and they're not represented as arrays

rhattie-ey (approved):

PR #939, opened 2023-03-27
HCAP-1443 MoH/HA See site that hired participant (#926)

mmumby:
## Work completed 
- This re-ads work in commit cc9e5679d277bc75033278b49ee780748ca7cf87.
- Added siteName to participantsDetails service


client/src/services/participant.js
```diff
@@ -44,12 +44,18 @@ export const fetchParticipant = async ({ id }) => {
...
+      : 'available';
+
     return {
       ...participant,
       cohort,
       cohortName: getCohortPsiName(cohort),
       postHireStatus,
       postHireStatusLabel: getPostHireStatusLabel(postHireStatus),
+      status,
+      siteName: participant.latestStatuses?.[0]?.siteName || null,
```
fw-noel, line 58: Instead of latestStatuses[0], I might suggest a .find here (where status = 'hired'). I don't think we can always guarantee that hired is the 0th status.

mmumby: The status is being populated from `latestStatuses[0]` line 47
```const status = participant.latestStatuses.length
      ? participant.latestStatuses[0].status
      : 'available';```
      
I wanted to display the site, that is associated with the latest status.

fw-noel: I guess I'm guarding against the possibility that, for example, there's two current statuses:
offer_made and hired (a valid scenario), and latestStatuses[0] is the offer made one
or
hired and reject_ack and archived (another valid scenario), and latestStatuses[0] is the reject one

mmumby: Should the UI display all valid statuses? (and associated sites next to the status)?

fw-noel: No, we're just doing hired site. My worry is if [0] doesn't capture the hired status, it might be missed!

mmumby: Updated as we discussed on the call!

PR #945, opened 2023-03-29
HCAP-1407 add a script to migrate keycloak users

sunghwan-park-fw:
- Supports migration to your local Keycloak using `password` grant_type and to MoH Keycloak using `client_credentials` grant_type
- Requires the following env variables
    - KEYCLOAK_API_CLIENTID
    - KEYCLOAK_FE_CLIENTID
    - KEYCLOAK_API_SECRET
    - KEYCLOAK_AUTH_URL
    - KEYCLOAK_REALM
    - KEYCLOAK_SA_PASSWORD
    - KEYCLOAK_SA_USERNAME
    - TARGET_KEYCLOAK_API_CLIENTID
    - TARGET_KEYCLOAK_FE_CLIENTID
    - TARGET_KEYCLOAK_API_SECRET
    - TARGET_KEYCLOAK_AUTH_URL
    - TARGET_KEYCLOAK_REALM
    - TARGET_KEYCLOAK_SA_PASSWORD
    - TARGET_KEYCLOAK_SA_USERNAME
    - POSTGRES_HOST
    - POSTGRES_PORT
    - POSTGRES_DB
    - POSTGRES_USER
    - POSTGRES_PASSWORD
- Create users if they are not found in the target
- Drop users if they have neither `bceid_userid` nor `idir_userid` in their `attributes`
- Use a batch pool to fetch and update users
- Found four users with the same username for OCIO and MoH Keycloaks, all from EY
- Command arguments
  - `--print-users` - print number of bceid and idir users
  - `--keycloak` - migrate keycloak users
  - `--database` - migrate database


fw-noel (commented):
Couple of things I think would be a good idea:
- Keeping the old keycloak ID somewhere, for potential cross referencing on logs. `users.body.oldKeycloakId` or something, 
- We may want to take this opportunity to make the database change for participants_status to be related to the actual foreign key of a user. I am thinking a migration for a new foreign key on the participants_status table to relate to users, and then we can decide what to do with ps.employer_id (we could potentially just leave it alone but never use it for anything; it would just be for audit purposes).

sunghwan-park-fw:
> Couple of things I think would be a good idea:
> 
> * Keeping the old keycloak ID somewhere, for potential cross referencing on logs. `users.body.oldKeycloakId` or something,
> * We may want to take this opportunity to make the database change for participants_status to be related to the actual foreign key of a user. I am thinking a migration for a new foreign key on the participants_status table to relate to users, and then we can decide what to do with ps.employer_id (we could potentially just leave it alone but never use it for anything; it would just be for audit purposes).

The old Keycloak ID is kept as `users.body.oldKeycloakId`.
For the relation of `users` and `parcipitants_status,` we need to define a foreign key, update `participants_status.employer_id`, and update some queries. I thought it would be better to be handled in a different ticket cause it is not logically related to this migration.

fw-noel (dismissed):
Seems good

PR #947, opened 2023-03-30
Cypress updates to inconsistent tests

mmumby:
## Work completed
- Possible fix to unreliable cypress tests. 
- Removed default timeout override
- Instead of using an arbitrary cy.wait, Updated the test to listen for the GET request to resolve before checking the DOM for an element.


rhattie-ey (approved):
Great work! Looks a whole lot more robust than the previous code.

fw-noel (approved):

PR #948, opened 2023-04-03
fix: cron job paths were no longer working

fw-noel:
This moves the logic to npm commands that the cron job can hit. This is due to the typescript change.


vanics (approved):
Straightforward changes. Approved!

PR #950, opened 2023-04-13
HCAP-1484 add foreign keys to participants_status

sunghwan-park-fw:


sunghwan-park-fw:
@fw-noel 
The branch is rebased on HCAP-1409 because Github Action's been updated with the new variables.

PR #953, opened 2023-04-18
add option for bcsc id and add user_id attr to keycloak users

coryatwater-FW:
- added user_id attributes to local keycloak users to prevent null userId's being returned

- ~~user id for bcsc login is now called `bcsc_guid`~~ **no longer needed**
- ~~added bcsc_guid attribute to test participant~~ **no longer needed**
- ~~added bcsc_guid client scope mapper for FE client so it is included in token~~ **no longer needed**

### DIFF
--- a/.docker/keycloak/realm-testing.json
+++ b/.docker/keycloak/realm-testing.json
@@ -817,6 +817,21 @@
       "nodeReRegistrationTimeout": -1,
       "protocolMappers": [
         {
+          "id": "99a318ea-7b43-45d4-af69-681d437d9c3e",
+          "name": "bcsc_guid",
+          "protocol": "openid-connect",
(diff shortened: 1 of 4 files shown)


.docker/keycloak/realm-testing.json
```diff
@@ -816,6 +816,21 @@
       "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": -1,
       "protocolMappers": [
+        {
+          "id": "99a318ea-7b43-45d4-af69-681d437d9c3e",
+          "name": "bcsc_guid",
```
fw-noel, line 821: We can just keep this all as user_id now! Sorry for the confusion there

fw-noel (approved):

PR #954, opened 2023-04-19
HCAP-1407 migrate keycloak users by middleware

sunghwan-park-fw:


server/keycloak.ts
```diff
@@ -118,13 +120,24 @@ class Keycloak {
     return async (req, res, next) => {
       try {
         const { content } = req.kauth.grant.access_token;
-        const roles = content?.resource_access[this.clientNameFrontend]?.roles || [];
-        const user = await getUser(content.sub);
+        let roles = content?.resource_access[this.clientNameFrontend]?.roles || [];
+
+        const keycloakId = content.sub;
+        const username = content.preferred_username;
```
fw-noel, line 126: May as well add `const email = ...` for consistency

server/keycloak.ts
```diff
@@ -373,6 +402,60 @@ class Keycloak {
...
+   * Migrate user's old keycloak id to new keycloak id and update roles
+   *
+   * @param keycloakId
+   * @param email
+   * @param username
+   */
+  async migrateUser(keycloakId: string, email: string, username: string): Promise<string[] | null> {
+    const migrationStatus = await dbClient.db[collections.USER_MIGRATION].findOne({
+      'email ilike': email,
+      'username ilike': username,
```
fw-noel, line 416: A gotcha here; the username in the new realm may be `username@bceid_business` while on our database, just `username@bceid`.

Maybe specifically for BCeIDs, we should probably ignore the latter part of the username
We can't ignore it in all cases in case someone's IDIR `username2@idir` is the same as someone else's bceid `username2@bceid_business`

server/keycloak.ts
```diff
@@ -301,33 +314,47 @@ class Keycloak {
     return `${this.apiUrl}/users/${userId}`;
   }
 
-  async setUserRoles(userId: string, role: string, regions: string[]) {
+  async deleteUserRoles(userId: string) {
```
fw-noel, line 317: Does this function need an authenticateIfNeeded if it's called outside of one of the `setUserRoles`?

server/migrations/1681833645009_create-user-migration-table.js
```diff
@@ -0,0 +1,44 @@
+import { PromisePool } from '@supercharge/promise-pool';
+import keycloak from '../keycloak';
+import { collections } from '../db';
+
+exports.shorthands = undefined;
+
+exports.up = async (pgm) => {
+  const users = [];
+
+  const keycloakUsers = await keycloak.getUsers();
```
fw-noel, line 10: Isn't an issue here that this will grab users from the target env?
I think let's keep the migration to just the table creation and populate it separately so as to avoid adding extra ENVs for the source keycloak environment.

server/keycloak.ts
```diff
@@ -373,6 +402,65 @@ class Keycloak {
...
+    }
+
+    try {
+      await this.setUserRoles(keycloakId, migrationStatus.roles);
+    } catch (e) {
+      logger.error(`failed to update user(${keycloakId})'s role`, { context: 'kc-user-migration' });
+      throw e;
+    }
+
+    const [user] = await dbClient.db[collections.USERS].findDoc({
```
fw-noel, line 438: Can I suggest instead storing the original keycloak ID in the table and using that instead of username match on the users table?

server/keycloak.ts
```diff
@@ -373,6 +402,65 @@ class Keycloak {
...
+      'userInfo.username ilike': username.includes('@bceid')
+        ? `${username.split('@')[0]}@bceid%`
+        : username,
+      'userInfo.email ilike': email,
+    });
+    if (user) {
+      user.userInfo.username = username;
+      await dbClient.db[collections.USERS].updateDoc(user.id, {
```
fw-noel, line 446: Will we want to reuse some of the work from your migration script to update id and participant status employer_ids?

server/migrations/1681833645009_create-user-migration-table.js
sunghwan-park-fw: This migration should be run with the current keycloak and we don't need additional envs.

server/keycloak.ts
sunghwan-park-fw: Right, I'll add it. However, we'd better refactor using axios interceptor to handle token expiration instead of an unnecessary blind call.

server/keycloak.ts
sunghwan-park-fw: I hope HCAP-1484 to be merged before this so that we don't need to update the `participant_status` table.

server/migrations/1681833645009_create-user-migration-table.js
fw-noel: I don't think we can separate that.

server/keycloak.ts
sunghwan-park-fw: I don't catch what you mean. I want to make sure that this logic doesn't require access to both Keycloaks.

1. Run database migration to create `user_migration` table with OCIO KEYCLOAK.
2. Switch to MoH KEYCLOAK
3. User log in with the MoH KEYCLOAK
4. Keycloak creates an account for the user
5. HCAP updates the user's roles on MoH KEYCLOAK.

server/migrations/1681833645009_create-user-migration-table.js
sunghwan-park-fw: Now, I get what you mean.

server/keycloak.ts
fw-noel: Agreed

server/keycloak.ts
fw-noel: No need for this, thank you!

fw-noel (approved):

PR #958, opened 2023-04-21
fix: do not request all users, only specific roles

fw-noel:
With no pagination, we want to ensure we avoid requesting all users. It'll be pretty big in the MoH Environment

### DIFF
--- a/server/constants/user-roles.ts
+++ b/server/constants/user-roles.ts
@@ -1,0 +1,12 @@
+export enum Role {
+  Superuser = 'superuser',
+  HealthAuthority = 'health_authority',
+  Employer = 'employer',
+  MinistryOfHealth = 'ministry_of_health',
+  Participant = 'participant',
+  Pending = 'pending',
+}
+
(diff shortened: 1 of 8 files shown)


server/keycloak.ts
```diff
@@ -286,12 +286,13 @@ class Keycloak {
         return response.data.filter((user) => user.username !== 'service-account');
       };
 
+      const rolesToRetrieve = ['ministry_of_health', 'employer', 'health_authority'];
       if (!ignorePending) {
-        return getData(`${this.apiUrl}/users?briefRepresentation=true&max=1000000`);
+        rolesToRetrieve.push('pending');
```
sunghwan-park-fw, line 291: We can remove getData cause it's not resued any more.

sunghwan-park-fw (dismissed):

server/constants/user-roles.ts
```diff
@@ -0,0 +1,12 @@
+export enum Role {
```
sunghwan-park-fw, line 1: I wanted to update all string roles with this enum, but I am going to do that in another ticket cause it'll cause too many files changes

PR #964, opened 2023-04-25
HCAP-1487 add user migration status table

sunghwan-park-fw:


server/services/user.ts
```diff
@@ -86,3 +86,24 @@ export const makeUser = async ({ keycloakId, sites }) => {
...
 };
+
+export const getUserMigrations = async () => {
+  const users = await dbClient.db[collections.USER_MIGRATION]
+    .join({
+      userJoin: {
+        type: 'LEFT',
+        relation: collections.USERS,
+        on: {
+          'body.keycloakId': 'id::TEXT',
```
fw-noel, line 97: Aren't we updating their keycloak id when we 'migrate' them right? So this won't join users that have already been updated, correct?

I'm okay with making this a "Users to be migrated" table though, and saying that its only for Pending users. We can remove the status column if we do end up doing that.

fw-noel (approved):

PR #966, opened 2023-04-25
HCAP-1486 log out of siteminder

sunghwan-park-fw:
Log out redirect chain: Keycoak -> SiteMinder -> HCAP

### DIFF
--- a/client/src/components/generic/Header.js
+++ b/client/src/components/generic/Header.js
@@ -85,7 +85,15 @@ export const Header = ({ hideEmployers = false }) => {
 
   const handleLogoutClick = async () => {
     store.remove('TOKEN');
-    await keycloak.logout({ redirectUri: window.location.origin });
+
+    let redirectUri = window.location.origin;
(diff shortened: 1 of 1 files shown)


client/src/components/generic/Header.js
```diff
@@ -85,7 +85,10 @@ export const Header = ({ hideEmployers = false }) => {
 
   const handleLogoutClick = async () => {
     store.remove('TOKEN');
-    await keycloak.logout({ redirectUri: window.location.origin });
+
+    const siteMinderHost = process.env.NODE_ENV === 'production' ? 'logon7' : 'logontest7';
```
fw-noel, line 89: Can we use `APP_ENV === 'prod'`?

sunghwan-park-fw: It is on FE, thus APP_ENV must be empty. I know what you mean, so I'll try to find a workaround.

fw-noel (approved):

PR #972, opened 2023-04-27
use axios on frontend

sunghwan-park-fw:


sunghwan-park-fw:
I let the report view keep using `fetch.` It requires much change out of the scope.

PR #983, opened 2023-09-26
HCAP-1491 add fields to participant

sunghwan-park-fw:
- Add new fields to participants
  - program (HCA or MHAW)
  - educationalRequirements (Yes/No)
  - driverLicense (Yes/No)
  - indigenous (Yes/No/Prefer not to answer/Unknown)
  - experienceWithMentalHealthOrSubstanceUse (Yes/No/Prefer not to answer/Unknown)
  - currentOrMostRecentIndustry
  - roleInvolvesMentalHealthOrSubstanceUse (Yes/No/Unknown)
- add a migration to initialize new fields for the current participants


server/validators/participant.ts
```diff
@@ -117,6 +125,9 @@ export const ParticipantSchema = yup
       .required(errorMessage)
       .matches(/^[A-Z]\d[A-Z]\s?\d[A-Z]\d$/, 'Postal code must be in format A1A 1A1'),
 
+    indigenous: yup.string().oneOf(yesOrNo, 'Indigenous should be Yes or No'),
```
coryatwater-FW, line 128: the docs say this question isn't required, should confirm

server/constants/participant-fields.ts
```diff
@@ -0,0 +1,21 @@
...
+  'interested',
+  'crcClear',
+  'callbackStatus',
+  'statusInfo',
+  'userUpdatedAt',
+  'distance',
+  'progressStats',
+  'postHireStatuses',
+  'rosStatuses',
+];
```
coryatwater-FW, line 21: @fw-noel  theres a few other fields/ conditional questions from the docs, should these be here? 

, I also have: `experienceWithMentalHealthOrSubstanceUse`, `currentOrMostRecentIndustry`, `roleInvolvesMentalHealthOrSubstanceUse`, and `otherIndustry`, I don't believe they're 'required' though

server/validators/participant.ts
```diff
@@ -97,13 +99,19 @@ export const ParticipantSchema = yup
...
     // Eligibility
     eligibility: yup
       .boolean()
       .typeError(errorMessage)
       .required(errorMessage)
       .test('is-true', errorMessage, (v) => v === true),
 
+    educationalRequirements: yup
+      .string()
+      .oneOf(yesOrNo, 'Educational requirements should be yes or no'),
```
coryatwater-FW, line 113: capitalize yes and no to match other messages?

server/tests/participant.test.ts
```diff
@@ -51,6 +51,10 @@ describe('Participants Endpoint', () => {
     const expectedRes = {
       ...newParticipant,
       preferredLocation: 'Fraser',
+      program: 'HCA',
+      educationalRequirements: 'unknown',
+      driverLicense: 'unknown',
```
coryatwater-FW, line 56: I've been thinking about how driver license can be unknown, I get this is for current EOI's, but the only values are yes/ no, so true false, would setting this to null be better?

server/validators/participant.ts
sunghwan-park-fw: It is optional. It could be debatable to treat 'Prefer not to answer' to be equal to 'unknown'.

server/constants/participant-fields.ts
sunghwan-park-fw: I'll add them.

server/validators/participant.ts
coryatwater-FW: changed value to `preferNotToAnswer`

server/validators/participant.ts
fw-noel: I do think we have to distinct between "I don't know" (FE) and "unknown" (what we backfill existing participants with). So I do think we change this validation

server/validators/participant.ts
fw-noel: Same with this one. We've determined that this question is optional so the options are
Yes
No
Prefer not to answer (Backend -> Prefer not to answer)
Blank (Backend -> Unknown)

sunghwan-park-fw: Then, optional yes or no can have four options, 'Yes, No, Prefer not to answer, Unknown' whereas required one has just 'Yes' or 'No'.

Capitalization for the value is a little messy. Some fields use capital and some don't. Even there is a migration to change `interested` value to lowercase.

server/validators/participant.ts
```diff
@@ -97,13 +100,19 @@ export const ParticipantSchema = yup
...
     // Eligibility
     eligibility: yup
       .boolean()
       .typeError(errorMessage)
       .required(errorMessage)
       .test('is-true', errorMessage, (v) => v === true),
 
+    educationalRequirements: yup
+      .string()
+      .oneOf(yesOrNo, 'Educational requirements should be Yes, No'),
```
coryatwater-FW, line 114: education question has a 3rd option `I don't know`, it's set to `do not know` on my end

server/validators/participant.ts
```diff
@@ -117,6 +126,24 @@ export const ParticipantSchema = yup
       .required(errorMessage)
       .matches(/^[A-Z]\d[A-Z]\s?\d[A-Z]\d$/, 'Postal code must be in format A1A 1A1'),
 
+    indigenous: yup
+      .string()
+      .oneOf(yesOrNoOptional, 'Indigenous should be Yes, No, Prefer not to answer, or Unknown'),
```
coryatwater-FW, line 131: since this is optional, what happens if this comes back as `' '` ?

server/validators/participant.ts
```diff
@@ -117,6 +126,24 @@ export const ParticipantSchema = yup
...
+      .oneOf(
+        yesOrNoOptional,
+        'Experience of MHSU should be Yes, No, Prefer not to answer, or Unknown'
+      ),
+
+    currentOrMostRecentIndustry: yup.string(),
+
+    roleInvolvesMentalHealthOrSubstanceUse: yup
+      .string()
+      .oneOf(yesOrNo, 'Involved in delivering MHSU service should be Yes, No, or Unknown'),
```
coryatwater-FW, line 145: same question here

PR #984, opened 2023-09-27
HCAP-1492 - add MHAW track to EOI form

coryatwater-FW:
### Pending
- need link for `MENTAL_HEALTH_AND_ADDICTIONS_WORKER_LINK`, it's currently empty, left `TODO` in links.js

### Summary
- added Mental Health and Addictions Worker track to EOI form
- add cypress test for 2 new `conditional` fields
- extract each section to own component to reduce duplicity errors

### New fields
- program
- educationalRequirements
- indigenous
- driverLicense
- experienceWithMentalHealthOrSubstanceUse - (conditional)
- currentOrMostRecentIndustry
- roleInvolvesMentalHealthOrSubstanceUse - (conditional)
- otherIndustry

### JIRA Ticket
[HCAP-1492](https://eydscanada.atlassian.net/browse/HCAP-1492)


client/src/components/participant-form/index.js
```diff
@@ -35,15 +35,23 @@ export const Form = ({
...
         postalCode: '',
+        indigenous: '',
         preferredLocation: [],
+        driverLicense: '',
+        experienceWithMentalHealthOrSubstanceUse: '',
         consent: '',
         reasonForFindingOut: [],
+        currentOrMostRecentIndustry: '',
+        roleInvolvesMentalHealthOrSubstanceUse: '',
+        otherIndustry: '',
```
sunghwan-park-fw, line 54: Isn't it possible to merge this into currentOrMostRecentIndustry?

coryatwater-FW: I do have code to do this saved locally, but I wanted to get some opinions! (also this other option may even be removed)

my thought  was having both would make it more apparent of the users `Other` industry as opposed to just having the specified value in `currentOrMostRecentIndustry`.  That way we know the user had selected Other + there industry.

Probably a stretch scenario, but if we ever wanted to sort by industry for example

sunghwan-park-fw: There might be pros and cons, but the mismatch of questions and fields might cause more trouble than we can expect now.

server/validators/helpers.ts
```diff
@@ -93,15 +95,19 @@ export const errorMessage = ({ path }) => {
 
     // Participant info
     eligibility:
-      "We're sorry, but current eligibility to work in Canada is a requirement to submit this form.",
-    preferredLocation: "Please select at least one location you'd like to work in.",
-    consent: "We're sorry, but we cannot process your request without permission.",
+      "We're sorry, but current eligibility to work in Canada is a requirement to submit this form",
+    educationalRequirements: `We're sorry, but this question is a requirement to submit this form`,
+    preferredLocation: "Please select at least one location you'd like to work in",
+    driverLicense: `We're sorry, but this question is a requirement to submit this form`,
```
fw-noel, line 101: I'd remove We're sorry out of this one, unless you got that from the PEOI? Same with the education requirements one

client/src/components/participant-form/index.js
fw-noel: I agree with Sunghwan that I don't see a huge value add of putting this in a different field

sunghwan-park-fw (approved):

sunghwan-park-fw (dismissed):

sunghwan-park-fw (approved):

PR #985, opened 2023-10-03
try to address duplicity sonar report, extract sections into components

coryatwater-FW:
- meant to address duplicity sonar report in base branch

### DIFF
--- a/client/src/constants/form-options.js
+++ b/client/src/constants/form-options.js
@@ -9,3 +9,141 @@ export const YesNoPreferNot = [
 ];
 
 export const YesNoDontKnow = [...YesNo, { value: `I don't know`, label: `I don't know` }];
+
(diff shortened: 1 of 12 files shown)


client/src/constants/form-options.js
```diff
@@ -9,3 +9,141 @@ export const YesNoPreferNot = [
 ];
 
 export const YesNoDontKnow = [...YesNo, { value: `I don't know`, label: `I don't know` }];
+
+export const currentOrMostRecentIndustryOptions = [
```
sunghwan-park-fw, line 13: I found that in some cases, we define values as an array and map options dynamically if the value and label are the same. See ArchiveHiredParticipantForm.formatOptions.

sunghwan-park-fw (approved):
Looks good, just left a minor comment.

PR #987, opened 2023-10-04
HCAP-1493 add MHSU employer role

sunghwan-park-fw:
Add MHSU employer user role to user management and access management dialogs


coryatwater-FW:
could we get the keycloak realm export for it too?

coryatwater-FW (approved):

PR #989, opened 2023-10-04
HCAP 1494/1495 update tests

coryatwater-FW:


server/scripts/randomize-participants.ts
```diff
@@ -31,6 +31,7 @@ const generateParticipants = async (amount: number) => {
         preferredLocation: DEFAULT_PREF_LOCATION,
         interested: 'yes',
         crcClear: 'yes',
+        program: Math.floor(Math.random() * 2) === 0 && 'MHAW', //NOSONAR
```
sunghwan-park-fw, line 34: Should we include 'HCA'? We can use `_.sample`.

coryatwater-FW: oh right yes, I thought it was HCA by default, but I was wrong in how this worked, will add

sunghwan-park-fw (approved):

fw-noel (approved):

PR #990, opened 2023-10-04
HCAP-1497 add new fields to participant view

sunghwan-park-fw:
screenshot-hcapemployers localhost_4000-2023 10 05-11_02_50


sunghwan-park-fw:
The ticket is only for the view, but I wonder if some new fields should be included in the edit dialogue.

client/src/constants/participantTableConstants.js
```diff
@@ -157,7 +157,24 @@ const columns = {
   phoneNumber: { id: 'phoneNumber', name: 'Phone Number', sortOrder: 8 },
   emailAddress: { id: 'emailAddress', name: 'Email Address', sortOrder: 9 },
   preferredLocation: { id: 'preferredLocation', name: 'Preferred Region(s)', sortOrder: 10 },
+  driverLicense: { id: 'driverLicense', name: `Driver's License`, sortOrder: 10.1 },
```
coryatwater-FW, line 160: how come these ones are decimals?

sunghwan-park-fw: `sortOrder` is a comparator parameter to sort columns, so it doesn't need to be an integer.
I just didn't want to reset all the higher ones.

coryatwater-FW (approved):

coryatwater-FW (approved):

PR #1000, opened 2023-10-16
HCAP-1510 allow only vancouver island and interior to access MHAW

sunghwan-park-fw:
- Hide disabled location and program filters
- Allow only vancouver island and interior HA and employer users to access MHAW

### DIFF
--- a/client/src/pages/private/ParticipantTable.js
+++ b/client/src/pages/private/ParticipantTable.js
@@ -380,14 +380,13 @@ const ParticipantTable = () => {
       isMoH || isSuperUser ? regions : roles.map((loc) => regionLabelsMap[loc]).filter(Boolean)
     );
 
-    if (isMoH || isHA || isSuperUser) {
-      if (isHA && !MHAW_ENABLED_REGIONS.some((region) => roles.includes(region))) {
-        // MHAW program isn't open yet to other regions
-        setPrograms(programsHCA);
-      } else {
-        // should be able to see all users in both MHAW/ HCA programs
-        setPrograms(allPrograms);
-      }
+    if (isMoH || isSuperUser) {
+      // should be able to see all users in both MHAW/ HCA programs
+      setPrograms(allPrograms);
+    } else if (isHA) {
+      // MHAW program isn't open yet to other regions
+      const allowedForMHAW = MHAW_ENABLED_REGIONS.some((region) => roles.includes(region));
+      setPrograms(allowedForMHAW ? allPrograms : programsHCA);
     } else if (isEmployer) {
       // private employers should only see HCA program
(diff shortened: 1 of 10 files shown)


client/src/pages/private/ParticipantTable.js
```diff
@@ -379,14 +379,22 @@ const ParticipantTable = () => {
     );
 
     if (isMoH || isHA || isSuperUser) {
-      // should be able to see all users in both MHAW/ HCA programs
-      setPrograms(allPrograms);
+      if (
+        isHA &&
+        !['region_vancouver_island', 'region_interior'].some((region) => roles.includes(region))
```
fw-noel, line 384: Can you pop this array in a const at the top?

sunghwan-park-fw: Done for BE and FE

server/services/participants/participant-entries.ts
```diff
@@ -188,13 +186,17 @@ export const getParticipants = async (
   isIndigenousFilter?: IsIndigenousFilter,
   programFilter?: ProgramFilter
 ) => {
+  // MHAW program isn't open yet to other regions
+  const hcaOnly =
+    user?.isHA && !MHAW_ENABLED_REGIONS.some((region) => user?.roles.includes(region));
```
coryatwater-FW, line 191: we should be checking for `isMHSUEmployer` too no?

server/services/participants/participant-entries.ts
```diff
@@ -188,13 +186,17 @@ export const getParticipants = async (
...
+
   // Get user ids
   const participantsFinder = new ParticipantsFinder(dbClient, user);
   const interestFilter =
     isPrivateEmployerOrMHSUEmployerOrHA(user) && statusFilters?.includes('open');
   let participants = await participantsFinder
     .filterRegion(regionFilter)
-    .filterProgram(programFilter)
+    .filterProgram(hcaOnly ? 'HCA' : programFilter)
```
coryatwater-FW, line 199: need to update the `isMHSUEmployer` check here to use `programFilter` if it exists, currently just hardcoded to `MHAW`

coryatwater-FW, line 199: nevermind!

server/services/participants/participant-entries.ts
sunghwan-park-fw: Why? I think the region has a higher priority than the employer type.

coryatwater-FW: I don't see anywhere limiting MHSU employer's by interior and costal.  If a MHSU employer from Fraser logs in, these changes won't show participants to them?

fw-noel (approved):

PR #1007, opened 2023-10-24
HCAP-1506 split allocations and hires into HCA/MHAW

sunghwan-park-fw:
- Add migration
  - to add `mhaw_allocation` column to `site_phase_allocation` table
  - to set `program` to `Non-HCAP` for `participant_status` entries whose `nonHcapOpportunity` is true
  - to set `program` to `HCA` for `participant_status` entries whose `nonHcapOpportunity` is null or false
- Add 'MHAW allocation' field to `Allocation Form`
- Change `Non-HCAP checkbox` to `Progrem select radio` in the `Hire Form`
- Change `Non-HCAP` column of the participant table to `Program` 
- Add `Program` field to the `Site Info` section 
- Add allocation and hires for each HCA and MHAW to `Site Positions Overview`
- Add allocation and hires for each HCA and MHAW to `Allocation table`
- Add program column to hire, ros, psi reports
- Upgrade Cypress to v13


client/src/constants/participantTableConstants.js
```diff
@@ -108,11 +108,15 @@ export const FILTERABLE_FIELDS = {
...
-export const programsMHAW = [MHAW];
+export const Program = {
+  HCA: 'HCA',
+  MHAW: 'MHAW',
+  NonHCAP: 'Non-HCAP',
+};
+
+export const allPrograms = [Program.HCA, Program.MHAW];
+export const programsHCA = [Program.HCA];
+export const programsMHAW = [Program.MHAW];
```
coryatwater-FW, line 119: I made a `constants/programs.js` file a little bit ago with a `Program` const, would adding `NonHCAP` to that work the same (It has `Unknown` as well)? and could we move this and `Program` into there?  just to match the purpose of the file

client/src/utils/site-util.js
```diff
@@ -0,0 +1,11 @@
+import { Program } from '../constants';
+
+export const getSiteProgramFromAllocation = (phases) => {
+  const hca = phases.some(({ allocation }) => allocation > 0);
+  const mhaw = phases.some(({ mhawAllocation }) => mhawAllocation > 0);
+  if (hca && mhaw) {
+    return `${Program.HCA} + ${Program.MHAW}`;
+  }
+  if (mhaw) return Program.MHAW;
```
coryatwater-FW, line 9: could we add curly braces, even if it's just a one liner? just for the styling

server/scripts/randomize-participants.ts
```diff
@@ -31,7 +32,7 @@ const generateParticipants = async (amount: number) => {
         preferredLocation: DEFAULT_PREF_LOCATION,
         interested: 'yes',
         crcClear: 'yes',
-        program: Math.floor(Math.random() * 2) === 0 && 'MHAW', //NOSONAR
+        program: Math.floor(Math.random() * 2) === 0 && Program.MHAW, // NOSONAR
```
coryatwater-FW, line 35: shoot, guess I didn't update this from one of my PR's, I think you're idea was to use `_.` something so it picks MHAW or HCA, if you don't mind?

server/services/phase.ts
```diff
@@ -21,11 +22,15 @@ type sitePhase = {
...
-  remainingHires: number;
-  /** Number of hires from HCAP */
-  hcapHires: number;
-  /** Number of hires from outside HCAP */
+  /** Number of remaining HCA hires */
+  remainingHcaHires: number;
+  /** Number of remaining MHAW hires */
+  remainingMhawHires: number;
+  /** Number of HCA employees hired (as a string) */
+  hcaHires: number;
```
coryatwater-FW, line 30: comments a little confusing, as a string but it's a number?

coryatwater-FW (approved):
pulled it down, looks and works good to me

PR #1013, opened 2023-11-02
fix participant table

sunghwan-park-fw:
- fix missing site names
- fix irrelevant sites lit
- fix missing status infos

### DIFF
--- a/server/services/employers.ts
+++ b/server/services/employers.ts
@@ -112,9 +112,12 @@ export const getSitesForUser = async (user: HcapUserInfo): Promise<EmployerSite[
     additionalCriteriaParams.userSites = user.sites;
   }
 
-  if (user.isHA && user.regions.length > 0) {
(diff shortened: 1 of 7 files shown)


server/services/employers.ts
```diff
@@ -112,10 +112,13 @@ export const getSitesForUser = async (user: HcapUserInfo): Promise<EmployerSite[
     additionalCriteriaParams.userSites = user.sites;
   }
 
-  if (user.isHA && user.regions.length > 0) {
+  if (user.regions.length > 0) {
```
fw-noel, line 115: An MoH will not enter this condition as they have no regions. Is that intentional?

sunghwan-park-fw: Not entering this condition means querying for all regions for MoH.

fw-noel (approved):

PR #1016, opened 2023-11-03
HCAP requested changes

coryatwater-FW:
- Changing ROS Date - allow future dates
- Remove Peer Roles column on participants table
- Remove Lived Checkbox and column if HCA selected from the dropdown
- "Is the employee remaining in the MHSU sector" text -> "Is the participant continuing to work in the MHSU sector, within this role or another"
- "Is the employee remaining in the HCA sector" text -> "Is the participant continuing to work as a HCA, within this role or another"


client/src/components/generic/Table.js
```diff
@@ -213,7 +214,21 @@ export const Table = ({
...
   multiSelectAction,
+  filter,
 }) => {
+  const [columnState, setColumnState] = useState(columns);
+  // living checkbox should be hidden if use filters program by HCA
+  // remove checked box if user selects living experience filter and switches to HCA
+  useEffect(() => {
+    if (filter?.programFilter?.value === Program.HCA) {
+      setColumnState((prevColumns) =>
+        prevColumns.filter((i) => i.id !== 'experienceWithMentalHealthOrSubstanceUse')
```
sunghwan-park-fw, line 225: Can't we use columns instead of prevColumns?

sunghwan-park-fw (approved):

client/src/components/generic/Table.js
coryatwater-FW: Yeah you're right, some testing I forgot to remove! Will fix

sunghwan-park-fw: I'll update this in my PR.

PR #1027, opened 2024-01-08
form changes

fw-noel:


client/src/components/modal-forms/NewParticipantForm.js
```diff
@@ -119,6 +135,62 @@ export const NewParticipantForm = ({ submissionCallback, onClose, sites }) => {
...
+              options={[
+                { value: 'yes', label: 'Yes' },
+                { value: 'no', label: 'No' },
+                { value: 'dontknow', label: 'Does not know' },
+              ]}
+            />
+            <Field
+              name='currentOrMostRecentIndustry'
+              component={RenderTextField}
+              label='What industry did they most recently work in?'
```
sunghwan-park-fw, line 171: don't the users respond about themselves? I wonder if it should be 'did you'.

sunghwan-park-fw (approved):
Left a comment about the point of view.

client/src/components/modal-forms/NewParticipantForm.js
fw-noel: This is when employers / HAs are adding non portal hires, so they would be appropriate!

PR #1028, opened 2024-02-27
HCAP-1520 - non portal hire form updates

coryatwater-FW:
- added additional fields from the PEOI form on the non-portal hire form, so that they have data and reporting consistency across all hires.
-  spreadsheet can be found in ticket

- removed `'Should be able to download Program Monitoring report'` test as the button was removed in [PR-1026](https://github.com/bcgov/hcap/pull/1026/files)
### Ticket
[HCAP-1520](https://eydscanada.atlassian.net/browse/HCAP-1520)


cypress/integration/reportingView.spec.js
```diff
@@ -57,11 +57,6 @@ describe('Reporting Download', () => {
       cy.get('.MuiAlert-message').contains('Report generated successfully!').should('be.visible');
     });
 
-    it('Should be able to download Program Monitoring report', () => {
-      clickReportButton('Download program monitoring report');
-      cy.get('.MuiAlert-message').contains('Report generated successfully!').should('be.visible');
-    });
```
dbayly-freshworks, line 63: What is the purpose of removing this test?

dbayly-freshworks (approved):
This all looks good to me. One non-blocking question

cypress/integration/reportingView.spec.js
coryatwater-FW: that button was removed in [PR-1026](https://github.com/bcgov/hcap/pull/1026/files) and the test was never removed

PR #1031, opened 2024-03-07
HCAP - allow all HAs access to MHAW pathway

coryatwater-FW:
- remove conditions of only certain HAs having access to MHAW pathway
- allow access to all HAs to MHAW pathway
- add `Are you a Canadian citizen or permanent resident?` question to non portal hire form and do not allow form submit if answer is 'No'


dbayly-freshworks (approved):
Looks ok to me

PR #1124, opened 2025-08-13
update node version on workflow files

chloe-yuu:
update node.js version from v20 -> v24 on workflow files


npham49 (approved):
Looks good to me!

PR #1126, opened 2025-08-13
Dockerfile update to include .npmrc to client container

chloe-yuu:
Dockerfile update to include .npmrc to client container


npham49 (approved):
looks good!

PR #1127, opened 2025-08-14
fixing npm build error on server triggered by nodejs upgrade

chloe-yuu:
fixing npm build error on server triggered by nodejs upgrade


npham49 (approved):
npm run build works on both directories on node 24 on my machine.

PR #1129, opened 2025-08-15
Fix SPA catch-all route for Express 5 / path-to-regexp v6

chloe-yuu:
Fix SPA catch-all route for Express 5 / path-to-regexp v6


npham49 (approved):
good to go!

PR #1133, opened 2025-08-18
fixing the CSP violations issue related to nonce

chloe-yuu:


npham49 (approved):
approved! My fix was going in the same direction lol!

PR #1137, opened 2025-09-25
backend library upgrades and mongodb version update

chloe-yuu:


npham49 (approved):
Works for me! Just a note, right now, when I run this after removing the images, it encounters an issue with the old 3.6 Mongo data. I had to remove the `.mongo-data` folder for it to run.

npham49:
Will the Mongo on OpenShift be upgraded Chloe? https://github.com/bcgov/hcap/blob/96def88db8e63173314b829dc76ce6da417f9942/openshift/mongo.yml#L55 This might cause some incompatible data

npham49 (approved):
Looks good to me!

PR #1138, opened 2025-09-30
fix the email validation on participant info edit pop-up

chloe-yuu:
This PR fixes the email validation issue on the "Edit Participant" pop-up and "Add New Non-Portal Hire" pop-up, ensuring that inputs like "user@example" are correctly flagged as invalid


npham49 (approved):
Looks good, tested locally and everything works, since the validations for email formatting is correct I'm approving.

but there was an interesting issue, when the email fields are empty, this shows up "Failed validation on..." 

Image

PR #1139, opened 2025-10-01
fix: resolve 'a few seconds ago' display for participant Last Updated column

chloe-yuu:
Fix some legacy participant "Last Updated" column showing incorrect "a few seconds ago"


npham49 (approved):
Works for me!

PR #1155, opened 2026-01-15
fix keycloak session timeout errors

chloe-yuu:
- Enable checkLoginIframe to extend session from 5 to 30 minutes
- Add proper error handling for expired sessions instead of 'Failed to fetch' errors  
- Increase OpenShift route timeout from 90s to 300s

Fixes screen freezing after 6 minutes of inactivity.


npham49 (approved):
Good to go!

PR #1159, opened 2026-01-19
fix keycloak session timeout errors

chloe-yuu:
1. Replace generic “Failed to fetch” errors with clear session-expiration messages to prevent the screen from appearing frozen.
2. Implement periodic KeyCloak token refresh (every 240 seconds) to avoid token expiry and extend the effective session duration from 5 minutes to 30 minutes.


npham49 (approved):
Good to go!

